PR #202, opened 2015-01-21
Generate token-replaced text when indexing

yhahn:
New feature: if you provide a `geocoder_tokens` key when indexing, it will be used to generate a fully condensed version of document when indexing. Saves us work on the source data side -- no more need to manually do this.

Current:

    geocoder_tokens: { 'Street': 'St' }
    document text: "Main Street"
    indexed as: "Main Street"

    # we go through the extra effort outside of carmen to do this...
    document text: "Main Street, Main St"
    indexed as: "Main Street", "Main St"

Example:

    geocoder_tokens: { 'Street': 'St' }
    document text: "Main Street"
    indexed as: "Main Street", "Main St"

Will do some testing before mergetown + release.

cc @karenzshea @ingalls @sbma44

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -91,7 +91,7 @@ function update(source, docs, zoom, callback) {
         }
         if (TIMER) console.timeEnd('update:getall');
         if (TIMER) console.time('update:indexdocs');
-        indexdocs(docs, freq, zoom, updateCache);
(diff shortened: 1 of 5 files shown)


sbma44:
Can you explain where this config needs to be supplied? AFAIK carmen indexing is hidden away in `./make.sh build` right now, I'm unsure where it's going to pull this from. If this will be apparent in the tests, nm.

yhahn:
Works great.

PR #208, opened 2015-01-26
stdFeat Fix

ingalls:


ingalls:
85.50% pass in reverse geocoding tests.

PR #209, opened 2015-01-27
Increase reverse tolerance

ingalls:


ingalls:
Going to close this for the time being in the wake of #219.

PR #213, opened 2015-02-02
[2.x] remove UTFGrid support, switch all tests to use vector tiles

ingalls:
- Adds vTile tests for address clusters
- Adds tests for Point/Poly reverse geocodes
- Fixes incorrect Features call to `putGeocoderData`
- Adds vTile support to `./lib/api-mem`
- Changes address source detection to `loaded` instead of `feat` as feat is only guaranteed to have the `_id` field.


ingalls:
All tests passing except the `geocoder-vtile` tests. These are failing due to the light feature load here: https://github.com/mapbox/carmen/blob/vtileUnit/lib/context.js#L138 only containing the `id` and not the `_text` object. This must be caused by some difference between what tilelive stores in the light feature and how the tests create a feature.

yhahn:
@ingalls nice work -- let's hold on a merge until you've got the migration path nailed down and we have 100% confidence in flying on just vtiles alone.

Issue #215, opened 2015-02-05
Failing test


springmeyer:
I'm noticed that carmen master is passing on travis. However when I install carmen as a dependency and run its tests I'm hitting this failure:

```
/home/travis/build/mapbox/mapnik-swoop/node_modules/carmen/index.js:41
        results.forEach(function(info, i) {
TypeError: Cannot call method 'forEach' of undefined
    at Geocoder.<anonymous> (/home/travis/build/mapbox/mapnik-swoop/node_modules/carmen/index.js:41:17)
    at notify (/home/travis/build/mapbox/mapnik-swoop/node_modules/carmen/node_modules/queue-async/queue.js:45:26)
    at /home/travis/build/mapbox/mapnik-swoop/node_modules/carmen/node_modules/queue-async/queue.js:35:11
    at /home/travis/build/mapbox/mapnik-swoop/node_modules/carmen/node_modules/mbtiles/lib/mbtiles.js:231:44
    at Statement.errBack (/home/travis/build/mapbox/mapnik-swoop/node_modules/carmen/node_modules/mbtiles/node_modules/sqlite3/lib/sqlite3.js:16:21)
```

@ingalls @yhahn - any ideas on where to start looking? I'm seeing this at https://travis-ci.org/mapbox/mapnik-swoop/jobs/49543819#L1270. This is https://github.com/mapbox/mapnik-swoop which installs a bunch of modules - from their master version - and ensures tests pass with an upcoming node-mapnik version. In this case it does not look like the node-mapnik version is causing the test failures but rather something that changed in carmen that I need to adapt to? Like something changed with some new permission or S3 credential maybe?

springmeyer:
@yhahn @ingalls So the above error does not happen when Carmen tests are run normally, but does happen when Carmen is run by mapnik-swoop. What mapnik-swoop does is install carmen into `node_modules` and then uses Node.js child process to call out to `npm test`: https://github.com/mapbox/mapnik-swoop/blob/master/test.js#L81-L85. 

I presume something obvious is failing given that custom setup, but I'm blind to what it is. One clue is that Carmen is the only dependency that fails. So, what about the Carmen tests are unique?

yhahn:
Cool, I'll dig around. OTOH sounds like it may just be a path or something that needs to be smarter.

springmeyer:
Just ran to completion for me locally with no errors. I saw:

```
carmen --> 
Missing version for mapnik. Need to run npm install ?
waiting...
waiting...
[...]

PR #218, opened 2015-02-18
Add geojson input plugin

yhahn:
^^ did the latest mapnik release switch `vtile.addGeoJSON` to use the geojson input plugin? Feels a bit not like a patch semver kind of change?

cc @springmeyer @flippmoke

### DIFF
--- a/test/context.test.js
+++ b/test/context.test.js
@@ -10,6 +10,7 @@ var path = require('path');
 var mapnik = require('mapnik');
 
 mapnik.register_datasource(path.join(mapnik.settings.paths.input_plugins,'ogr.input'));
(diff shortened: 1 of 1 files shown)


springmeyer:
I released 3.1.6 yesterday. Sorry, realize now it should have been more than a patch increment due to the need to register a new plugin. The reason I missed this is that I'm depending on mapnik-swoop to test all deps and it was totally green after this change (https://travis-ci.org/mapbox/mapnik-swoop/builds/51181041).
 
So, the error is that I removed carmen from mapnik-swoop in haste and forgot to ping you again for help in figuring out what is wrong: https://github.com/mapbox/carmen/issues/215.

springmeyer:
This change looks good, going to merge it.

Issue #220, opened 2015-02-25
Carmen Debug Options


ingalls:
At the moment Carmen is very difficult to debug. Carmen should use the `DEBUG` flag to tell it to log the current valid results at the end of each operation.

Logging at the end of
- [x] Grid lookup
- [x] PhraseMatch
- [x] Spatial Match

cc/ @yhahn @karenzshea

yhahn:
My recommendation over just dumping results at various stages (which can be so large they they are not very useful) would be to have an option for tracing a specific feature ID through the process.

------

When I debug manually the situation I'm usually in is:

> Some feature N that I'm expecting to see is not in the result set.

So then I

- First figure out the feature ID that I'm expecting to see.
- Is it in the grid results of phraseMatch? What is the relev + reason + count of the feature at this point?
- Is it in the results of spatialMatch? What is the relev + reason + count of the feature at this point? Does it stack with other features?
- Does it make it past verifyMatch?

**In theory this could all be added to https://github.com/mapbox/carmen/blob/master/lib/geocode.js without touching any of the underlying functions since this lib has access to each one's input/output.**

PR #221, opened 2015-02-25
Debug options [WIP]

ingalls:
See #220 for more information

### DIFF
--- a/lib/util/ops.js
+++ b/lib/util/ops.js
@@ -32,9 +32,9 @@ module.exports.zxy = function(id, zxy) {
 
 module.exports.grid = function(grid) {
     return {
-        x: Math.floor(grid|0 / mp39),
-        y: Math.floor(grid|0 % mp39 / mp25),
-        id: Math.floor(grid|0 % mp25)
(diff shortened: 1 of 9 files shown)


yhahn:
@ingalls awesome, some thoughts:

- how about not console.log()'ing inline but instead having geocode() return some structured data?
- how about not using an env var in the lib itself but having it as a flag in `options` ?
- then carmen.js the script can use env var or commandline flag or whatever to call the API with the option and format the results for reading however it wants

Should make it easier to test and use it in an API context if we ever need to.

lib/util/ops.js
```diff
@@ -30,6 +30,14 @@ module.exports.zxy = function(id, zxy) {
...
 };
 
+module.exports.grid = function(grid) {
+    return {
+        x: Math.floor(grid / mp39 / 10),
+        y: Math.floor(grid % mp39 / mp25 / 10),
+        id: Math.floor(grid % mp25)
+    };
+};
+
```
yhahn: is this used anymore?

PR #223, opened 2015-02-25
Support for non-us style street addresses [WIP]

ingalls:
TODO
- [x] parse all tokens in termops for address
- [x] check if potential housenumber is in cluser/itp
- [x] make sure there is no conflict with postcode


yhahn:
@ingalls the way I'm thinking about this is that currently this code is very absolute:

- https://github.com/mapbox/carmen/blob/master/lib/verifymatch.js#L24
- https://github.com/mapbox/carmen/blob/master/lib/verifymatch.js#L59

It's basically declaring: **Yes there is a housenum!** or **No there is no housenum.** Instead I think we want the code to capture the concept of:

"This *may* be a housenum -- if features you have appear to match it well let's proceed cautiously, leaving the door open to the possibility that it's not and other features should win out."

ingalls:
So put the loop in the verifyMatch, gotcha

yhahn:
Or an else case here https://github.com/mapbox/carmen/blob/master/lib/verifymatch.js#L59-L70

ingalls:
TODO
- [ ] Update Relev / Bitmask

ingalls:
- `queryAddress` now returns address and position
- Position is then used to create an address bitmask that is `&` against the `relev.reason` and if `0` incremented by the address bitmask.

ingalls:
RIP Mr. Spock

PR #235, opened 2015-03-16
More aggressive deduping of addresses.

yhahn:
Address features may be dupes but not have identical place names. Abstract example:

    100 main st springfield 00001 (exact address point)
    100 main st springfield 00002 (interpolated point)

The place names differ because a result falls into another postal code area. These could be legitimately different addresses but in many cases they're the results of multiple results from datasets (interpolated vs exact, etc.)

In addition to deduping by `place_name` this branch adds deduping for features with `address` by `address + text` and a distance threshold of 10km as well.

cc @ingalls @karenzshea @sbma44

### DIFF
--- a/lib/geocode.js
+++ b/lib/geocode.js
@@ -8,6 +8,7 @@ var sm = new (require('sphericalmercator'))(),
     queue = require('queue-async'),
     feature = require('./util/feature'),
     Relev = require('./util/relev');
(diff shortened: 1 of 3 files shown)


yhahn:
Note: this will *never* remove the first geocoding result. This only affects additional suggestions from carmen.

PR #237, opened 2015-03-20
Reverse course capt

ingalls:
Ensure that when `addressitp` layers are above `address` layers the address point is returned.


yhahn:
:ok_hand:

PR #239, opened 2015-03-25
Shady character (bitmask address lookup)

ingalls:
At the moment carmen determines an address query based on the first addres-like token it encounters in a query.

This means that streets like `1 street 100` (where `100` is an address) fail as `1` is assumed to be the address. Using bitmasks prevents this as it exudes any of the currently matched text features from being used as an address.

### DIFF
--- a/lib/verifymatch.js
+++ b/lib/verifymatch.js
@@ -20,8 +20,6 @@ function verifymatch(query, stats, geocoder, matched, options, callback) {
     var results = matched.results;
     var coalesced = matched.coalesced;
 
-    var address = termops.queryAddress(query);
-
     var contexts = [];
     var start = +new Date();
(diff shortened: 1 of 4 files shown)


lib/verifymatch.js
```diff
@@ -42,6 +40,19 @@ function verifymatch(query, stats, geocoder, matched, options, callback) {
...
+            for (var i = 0; i < query.length; i++ ) {
+                if ((term.reason & Math.pow(2, i)) === 0) {
+                    address = termops.address(query[i]);
+                    if (address) {
+                        address = {addr: address, pos: i}
+                        break;
+                    }
+                }
+            }
+
```
yhahn: Want to move this into an exported function so it can be unit tested?

PR #242, opened 2015-04-01
Don't log weights [WIP]

ingalls:
Increase distance between index weights


yhahn:
Went with a different approach based on what we learned here!

PR #245, opened 2015-04-17
Proximity alert [WIP]

ingalls:
Implements two proximity sorts

The first sort occurs before `VerifyMatch`. VerifyMatch has a hard 40 feature cut-off. The features are binned by relev and then sorted within the bin.

The second sort occurs at the end of a forward geocode. Features are binned by similar relev and idx.


lib/geocode.js
```diff
@@ -217,6 +223,32 @@ function forwardGeocode(geocoder, query, options, callback) {
...
 
+        //If proximity is set, terms that tie in relev will be ordered by distance
+        //This is performed as verify match has a cutoff of the first 40 results
+        if (options.proximity) {
+            //Setup grid lookup - each layer is an object in the array with id->grid
+            var gridlookup = [];
+            for (var layer_it = 0; layer_it < grids.length; layer_it++) {
+                gridlookup[layer_it] = {};
+                for (var grid_it = 0; grid_it < grids[layer_it].length; grid_it++) {
+                    gridlookup[layer_it][ops.grid(grids[layer_it][grid_it]).id] = ops.grid(grids[layer_it][grid_it]);
```
yhahn: Run `ops.grid` once here instead of twice?

lib/geocode.js
```diff
@@ -217,6 +223,32 @@ function forwardGeocode(geocoder, query, options, callback) {
...
+            }
+            //The centre of each tile is calculated in merc and then sorted
+            var proxSM = sm.forward(options.proximity);
+            matched.results.sort(function(a, b) {
+                var res = [a,b].map(function(ele) { return new Relev(ele); });
+                if (res[0].relev !== res[1].relev) return 0; //Only sort with same relev
+                res = res.map(function(ele) {
+                    var center = toCenter(sm.bbox(gridlookup[ele.idx][ele.id].x, gridlookup[ele.idx][ele.id].y, zooms[ele.idx]));
+                    return Math.sqrt(Math.pow(center[0] - proxSM[0],2) + Math.pow(center[1] - proxSM[1],2));
+                });
```
yhahn: You'll have better speed here ditching arrays + maps and simply hardcoding two vars to do the work with.

test/geocode-unit.test.js
```diff
@@ -1111,6 +1111,105 @@ mapnik.register_datasource(path.join(mapnik.settings.paths.input_plugins,'geojso
...
+        addFeature(conf.province, province, t.end);
+    });
+    test('index province', function(t) {
+        var country = {
+            _id:3,
+            _text:'province',
+            _zxy:['1/1/0'],
+            _center:[145,70]
+        };
+        addFeature(conf.country, country, t.end);
```
yhahn: Is this meant to be indexed into the `country` index?

yhahn:
@ingalls overall looks good -- how would you feel about splitting the two main pieces of logic out into a separate module `proximity.js` so the functions can be exposed and unit tested?

test/geocode-unit.test.js
ingalls: Yes, this is to test across indexes.

test/proximity.test.js
```diff
@@ -0,0 +1,108 @@
+var proximity = require('../lib/util/proximity');
+var test = require('tape');
+
+test('proximity#toCenter', function(t) {
+    t.deepEquals(proximity.toCenter([ -180, 0, 0, 85 ]), [ -20037508.342789244, -19971868.88040857 ]);
+    t.deepEquals(proximity.toCenter([ -180, -85, 0, 0 ]), [ -20037508.342789244, -19971868.88040857 ]);
+    t.deepEquals(proximity.toCenter([ 0, 0, 180, 85 ]), [ -9462156.717428254, -19971868.88040857 ]);
+    t.deepEquals(proximity.toCenter([ -84, 40, -80, 40 ]), [ -13803616.858365923, -7.081154551613622e-10 ]);
```
yhahn: I don't think this is correct... reread your `toCenter` logic?

lib/util/proximity.js
```diff
@@ -0,0 +1,89 @@
...
+ * @param  {Object} options geocoder options
+ * @return {Array}         sorted grid results
+ */
+function coarse(results, grids, zooms, options) {
+    //Setup grid lookup - each layer is an object in the array with id->grid
+    var gridlookup = [];
+    for (var layer_it = 0; layer_it < grids.length; layer_it++) {
+        gridlookup[layer_it] = {};
+        for (var grid_it = 0; grid_it < grids[layer_it].length; grid_it++) {
+            tmpGrid = ops.grid(grids[layer_it][grid_it]);
```
yhahn: missing a `var` here?

lib/util/proximity.js
```diff
@@ -0,0 +1,89 @@
...
+            gridlookup[layer_it][tmpGrid.id] = tmpGrid;
+        }
+    }
```
yhahn: Check out the decrementing while loop optimization trick from https://github.com/mapbox/carmen/pull/247/files. Might give you a boost over these for loops.

lib/util/proximity.js
```diff
@@ -0,0 +1,89 @@
...
+        if (a.relev !== b.relev) return 0; //Only sort with same relev
+
+        var centerA = toCenter(sm.bbox(gridlookup[a.idx][a.id].x, gridlookup[a.idx][a.id].y, zooms[a.idx]));
+        var centerB = toCenter(sm.bbox(gridlookup[b.idx][b.id].x, gridlookup[b.idx][b.id].y, zooms[b.idx]));
+        a = Math.sqrt(Math.pow(centerA[0] - proxSM[0],2) + Math.pow(centerA[1] - proxSM[1],2));
+        b = Math.sqrt(Math.pow(centerB[0] - proxSM[0],2) + Math.pow(centerB[1] - proxSM[1],2));
+
+        if (a < b) return -1; //a is closer
+        else if (a > b) return 1; //b is closer
+        else return 0; //same
```
yhahn: Trick from @mourner, you can replace this with just

    return a - b;

lib/util/proximity.js
```diff
@@ -0,0 +1,87 @@
...
+ * @return {Geojson}         return sorted result
+ */
+function fine(results, options) {
+    results.features.sort(function(a, b) {
+        //Only sort features with same relev/idx level
+        if (a.relevance !== b.relevance) return 0;
+        if (a.id.split('.')[0] !== b.id.split('.')[0]) return 0;
+
+        a = distance(Point(options.proximity), Point(a.center), 'miles');
+        b = distance(Point(options.proximity), Point(b.center), 'miles');
```
mourner: Is this a hot code? If so, you should cache as much as possible outside of the sort function. E.g. precalculate `.id.split...` and `Point(...)`.

lib/util/proximity.js
```diff
@@ -0,0 +1,87 @@
...
+    var proxSM = sm.forward(options.proximity);
+    results.sort(function(a, b) {
+        a = new Relev(a);
+        b = new Relev(b);
+        if (a.relev !== b.relev) return 0; //Only sort with same relev
+
+        var centerA = toCenter(sm.bbox(gridlookup[a.idx][a.id].x, gridlookup[a.idx][a.id].y, zooms[a.idx]));
+        var centerB = toCenter(sm.bbox(gridlookup[b.idx][b.id].x, gridlookup[b.idx][b.id].y, zooms[b.idx]));
+        a = Math.sqrt(Math.pow(centerA[0] - proxSM[0],2) + Math.pow(centerA[1] - proxSM[1],2));
+        b = Math.sqrt(Math.pow(centerB[0] - proxSM[0],2) + Math.pow(centerB[1] - proxSM[1],2));
```
mourner: No need to `sqrt` here since you just need to know how they compare. Also, `a * a` is usually faster than `Math.pow(a, 2)`. Also, centers and `new Relev` objects should be precalculated for each result before sorting as well.

test/geocode.test.js
```diff
@@ -56,13 +56,13 @@ test('geocode', function(t) {
             });
     });
     t.test ('invalid proximity lat', function(q) {
-            geocoder.geocode('n korea', { proximity: [98.177876,-59.504401]}, function(err, res) {
+            geocoder.geocode('n korea', { proximity: [-59.504401,98.177876]}, function(err, res) {
```
ingalls: Follow geoJSON order

lib/util/proximity.js
```diff
@@ -0,0 +1,106 @@
...
+    while (layer_it--) { //Allow grid lookup by gridlookup[idx][id]
+        gridlookup[layer_it] = {};
+        var grid_it = grids[layer_it].length;
+        while (grid_it--) {
+            var tmpGrid = ops.grid(grids[layer_it][grid_it]);
+            gridlookup[layer_it][tmpGrid.id] = tmpGrid;
+        }
+    }
+
+    relevlookup = {};
```
yhahn: var

PR #262, opened 2015-05-12
orderly

yhahn:
Iterates through ranges in a stable order when doing address interpolation.

### TODOs

- [x] Another testcase that is unit-testey
- [x] Test implications of stable sort order on reverse interpolation
- [x] RealWorldTesting


lib/pure/applyaddress.js
```diff
@@ -26,13 +27,41 @@ function standardize(feat){
...
+        ranges.push(f);
+    }
+    ranges.sort(function(a, b) {
+        if (a.lf !== b.lf) return a.lf < b.lf ? -1 : 1;
+        if (a.rf !== b.rf) return a.rf < b.rf ? -1 : 1;
+        if (a.lt !== b.lt) return a.lt < b.lt ? -1 : 1;
+        if (a.rt !== b.rt) return a.rt < b.rt ? -1 : 1;
+        if (a.lp !== b.lp) return a.lp < b.lp ? -1 : 1;
+        if (a.rp !== b.rp) return a.rp < b.rp ? -1 : 1;
+        if (a.lines.length !== b.lines.length) return a.lines.length < b.lines.length ? -1 : 1;
```
mourner: protip: the same can be written more efficiently as:

```js
return (a.lf - b.lf) || (a.rf - b.rf) ||
    (a.lt - b.lt) || (a.rt - b.rt) ||
    (a.lp - b.lp) || (a.rp - b.rp) ||
    (a.lines.length - b.lines.length);
```

Also, since the function doesn't depend on any outer variables, it's better to declare it separately in the top scope so that it's reused.

yhahn: :+1:

PR #274, opened 2015-05-26
Approximity

yhahn:
Prioritizes layer type + score consistently across proximity/non-proximity mode.

Stopgap for `proximity` mode until a stronger combination of score + distance can be tested for real world use.

cc @ingalls @jfirebaugh


jfirebaugh:
@yhahn Can you summarize the upshot for https://github.com/mapbox/mapbox-directions.js/pull/74?

yhahn:
@jfirebaugh yep, you'll get `San Francisco, CA` as the first result for a query like `San Francisco` universally, no matter where `proximity` is set to. Example results:

```
query: San Francisco
proximity: -121,37

- 0.99 San Francisco, 94127, California, United States (place.94102)
- 0.99 San Francisco, Zulia, Venezuela (place.26805575)
- 0.99 San Francisco, Heredia, Costa Rica (place.26809172)
- 0.99 San Francisco, Córdoba, Argentina (place.332510)
- 0.99 San Francisco, Leyte, Philippines (place.1199736)
```

Proiritizing by layer type, then score, then distance is a stopgap/non-optimal approach because there is no way atm for `San Francisco Ct, Merced, 95348, California, United States` to be in these top 5 results (though it'll surface nicely if your query is `san francisco ct`) even if you are 100 meters from this street.

@ian29 and I have planned some work to experiment with taking both score and distance into account and see how things behave when both factors are taken as influences.

PR #278, opened 2015-06-04
Ghost in the machine

ingalls:
This change alters the way `lib/verifyMatch.js` handles ghost features.

### Previous work:

1. @aaronlidman introduces ['ghost' features](https://github.com/mapbox/carmen/commit/906b025c9dadba8a7decc10cca301c1e47fa5c18) for alt places - [relevant devlog](https://github.com/mapbox/hey/issues/1825)
2. @yhahn allows ghost features to be [returned in context array](https://github.com/mapbox/carmen/pull/253)

### Before

- A ghost feature will never be returned in a reverse geocode
- A ghost feature can be returned in a context array _if_ the input query matches the ghost feature text

    - scored feature
    ```
    query:  7449 E River Rd Fridley
    result: 7449 E River Rd Fridley
    ```
    - ghost feature
    ````
    query:  7449 E River Rd Minneapolis
    result: 7449 E River Rd Minneapolis
    ````
    (these are the same feature)

- A ghost feature can never be directly searched for

    - scored feature
    ```
    query:  Fridley
    result: Fridley
    ```
    - ghost feature
    ````
    query:  Minneapolis
    result: null
    ````
    (these are the same feature)

### After

- A reverse geocode will never return a ghost feature
- A ghost feature can be returned in a context array
- A ghost feature can be searched for directly

### Why this change

Ghost features are currently useless if they are the lowest index. Since a ghost feature can only be returned in a context array, a ghost feature will never be returned if the query does not have an element below it. [relevant code](https://github.com/mapbox/carmen/blob/0bb5a624a26089fd50ed497ec103312ba36ecd31/lib/verifymatch.js#L57-L58).

In our places layer we have the postal polygons which one would use if they wanted to send mail. We have also ghosted in alternate places (populated places) that are useful to locals but not necessarily part of the postal service. When reverse geocoding the ghost features will ensure that the desired place is returned, however when forward geocoding one should be able to search for these populated places.

### Next up

- [x] test to see how it affects results before/after

cc/ @sbma44 @yhahn @ian29

### DIFF
--- a/lib/context.js
+++ b/lib/context.js
@@ -45,7 +45,8 @@ module.exports = function(geocoder, lon, lat, options, callback) {
     });
 
     q.awaitAll(function(err, res) {
-        if (err) return callback(err);
+        //if (err) return callback(err);
+        console.log(err);
         var stack = [];
         var memo = {};
 
@@ -121,7 +122,6 @@ function contextVector(source, lon, lat, full, matched, callback) {
         }, function(err, results) {
             if (err) return callback(err);
             if (!results || !results.length) return callback(null, false);
-
             var feat;
             var dist = Infinity;
 
--- a/lib/context.js
+++ b/lib/context.js
@@ -45,8 +45,7 @@ module.exports = function(geocoder, lon, lat, options, callback) {
     });
 
     q.awaitAll(function(err, res) {
-        //if (err) return callback(err);
-        console.log(err);
+        if (err) return callback(err);
         var stack = [];
         var memo = {};
 
--- a/lib/verifymatch.js
+++ b/lib/verifymatch.js
@@ -54,8 +54,6 @@ function verifymatch(query, stats, geocoder, matched, options, callback) {
                 // @TODO for fully accurate results, iterate through
                 // coalesced[coord] for a matching feature id.
                 var feat = features[id];
-                // filter out altnames, scored as -1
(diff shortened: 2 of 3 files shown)


lib/context.js
```diff
@@ -45,7 +45,8 @@ module.exports = function(geocoder, lon, lat, options, callback) {
     });
 
     q.awaitAll(function(err, res) {
-        if (err) return callback(err);
+        //if (err) return callback(err);
```
sbma44: leftover?

lib/verifymatch.js
```diff
@@ -54,8 +54,6 @@ function verifymatch(query, stats, geocoder, matched, options, callback) {
                 // @TODO for fully accurate results, iterate through
                 // coalesced[coord] for a matching feature id.
                 var feat = features[id];
-                // filter out altnames, scored as -1
-                if (typeof feat._score !== 'undefined' && feat._score < 0) continue;
```
sbma44: thinking out loud (the answer is probably no): with this line removed, is there any scenario by which multiple names come back for a single feature?

ingalls: As per voice, no these are the base features that are being matched, once the base features are matched it then goes and grabs the context arrays. See: https://github.com/mapbox/carmen/blob/GhostInTheMachine/lib/verifymatch.js#L34-L36

PR #279, opened 2015-06-09
enable regex token replacement

sbma44:
This functionality is already somewhat in place due to the nature of current token processing -- tokens containing regex characters could already behave in unexpected ways. This PR alters that codepath to use the XRegExp module.

- implement named group capture, making existing boundary checking safer and allowing for their use in replacement patterns
- check for numbered groups in the input and throw errors when they are encountered (carmen introduces new groups, making the ordering dangerous--we would have to do some fancy rewrite stuff which is best avoided in favor of named groups)


sbma44:
@ingalls ok to merge?

ingalls:
^^ Mind merging in master first and making this green?

sbma44:
ok, green! give it a once-over whenever you have a sec, please

package.json
```diff
@@ -23,7 +24,8 @@
     "turf-linestring": "1.0.x",
     "turf-point": "2.0.x",
     "turf-point-on-surface": "1.0.x",
-    "unidecode": "0.1.3"
+    "unidecode": "0.1.3",
+    "xregexp": "^2.0.0"
```
ingalls: I'd lock this to a specific version or at least to the patch level, the caret will match any minor version

ingalls:
Looks great @sbma44

PR #284, opened 2015-06-29
enable numbered group token replacement [vs c4]

sbma44:
`node bench/replaceToken.js`

**c4:** `token replace x 18,246 ops/sec ±0.91% (96 runs sampled)`
**regex-redux:** `token replace x 18,133 ops/sec ±0.82% (95 runs sampled)`


ingalls:
@sbma44 is this stale? If so can you close/delete the branch

sbma44:
yep stale

PR #308, opened 2015-07-23
Smallworld

yhahn:
The proximity zxy needs to stay within the unsigned integer space as it's hopping over to unsigned ints here https://github.com/mapbox/carmen-cache/blob/master/src/binding.cpp#L792.

Enforce this at the carmen input level.

Refs https://github.com/mapbox/geocoding-example/issues/2


tschaub:
Very nice.  Thanks @yhahn.

PR #319, opened 2015-08-03
Synonymize localized text

camilleanne:
Additional `carmen:text` parameters (e.g. `:carmen:text_ru` or `carmen:text_es`) for local language support in place names are indexed as synonyms.


test/geocode-unit.localtext.test.js
```diff
@@ -0,0 +1,68 @@
...
+};
+var c = new Carmen(conf);
+
+tape('index country', function(t) {
+    var country = {
+        _id:2,
+        _text:'Russian Federation, Rossiyskaya Federatsiya',
+        _text_ru: 'Российская Федерация',
+        _zxy:['6/30/30'],
+        _center:[0,0]
```
yhahn: @camilleanne do you want to update this feature to be of canonical geojson doc format?

camilleanne:
@yhahn I'd like your thoughts on my modification to decollide. Also, I feel like perhaps the step that's happening in `getIndexableText` and `decollide` may only need to happen once.

camilleanne:
Benchmark:
This branch (with changes to decollide):
```
# decollide
decollide (collision) x 96,011 ops/sec ±0.72% (91 runs sampled)
decollide (clean) x 101,122 ops/sec ±1.05% (85 runs sampled)
decollide (with localization) x 70,597 ops/sec ±7.10% (55 runs sampled)
decollide (with localization + unidecode) x 35,708 ops/sec ±1.91% (62 runs sampled)
```


Master:
```
# decollide
decollide (collision) x 164,937 ops/sec ±8.41% (61 runs sampled)
decollide (clean) x 138,953 ops/sec ±0.73% (88 runs sampled)
```

camilleanne:
@apendleton so what do we need to do to get your fork of unidecode running in this?

apendleton:
@camilleanne should just be swapping out the dep in package.json and changing the require, but I want to do a bit more testing for correctness if we're actually thinking about using it. I've stolen the upstream unidecode tests for now, but they're kind of lacking and I'd like to be more confident that what I have is right before relying on it. I think it's a small-ish lift, but wanted to make sure it was something that would be useful to you before I spent workday time on it. Sounds like that's a yes? If so, I'll test a bit more and submit a PR against this branch with the swap.

camilleanne:
@apendleton :+1: lets see what happens!

apendleton:
Cool. I spent a bit of time at the end of the day today identifying behavior differences between my code and node-unidecode, and there are some upstream behaviors that I'm 99% sure are wrong. I'm just gonna go for parity for now, but at some point it might be worth figuring out what the actual correct thing is. I think I can get to being confident enough to test with carmen without needing to tackle any of that, though, probably by end of day tomorrow.

apendleton:
@camilleanne okay, [submitted a PR](https://github.com/mapbox/carmen/pull/345) to switch this branch to unidecode-cxx.

New benchmarks from Travis:

    # decollide
    
    decollide (collision) x 115,361 ops/sec ±0.37% (97 runs sampled)
    decollide (clean) x 123,077 ops/sec ±0.53% (99 runs sampled)
    decollide (with localization) x 84,043 ops/sec ±0.48% (102 runs sampled)
    decollide (with localization + unidecode) x 77,548 ops/sec ±0.09% (103 runs sampled)

apendleton:
One other thing I should mention: neither old nor new unidecode do anything to detect whether or not running on a string is necessary before running on it. I'm not sure where the unidecode call is going to happen, and specifically whether it's only going to be called on text that we think is non-latin (like, from Chinese data or whatever) or whether it'll be called on everything, but if the latter, we could add a check in c++ land that would return the input if there's no operation necessary. That check would be pretty cheap, but not free (no allocations or function calls, but some bit-twiddling), so probably only worth it if we think there's going to be a decent number of calls where this thing ends up being a no-op.

Or it may be fast enough as-is, and not worth worrying about.

yhahn:
:rocket:

sbma44:
@apendleton I think that check is worth consideration. our customers are still overwhelmingly anglophone.

apendleton:
@sbma44 okay, I can cut a ticket. Given that this would probably result in a (slight) speed hit on non-latin text in exchange for a speed win on latin text, I'd probably need a test corpus to benchmark that represents a realistic mix of latin vs. non-latin text to see if the overall speed gain is worth it. Thoughts on what to use?

sbma44:
@apendleton I think this is a question for #devops. I suspect we can replay some portion of incoming queries. OTOH, our traffic is so bursty that it might take a long time to sample something representative. Curious if @yhahn has ideas here.

yhahn:
@apendleton @sbma44 I think we can leave this no-op to next time/if it matters. Going from 77k => 84k ops/sec (the fastest it would get us) is not that important at this point : )

apendleton:
@yhahn sounds good

PR #320, opened 2015-08-05
Catch misuse

yhahn:
Unclear what conditions cause this. Catch the error and pass back to learn more.


yhahn:
Tests pass locally, not waiting for travis backup.

PR #325, opened 2015-09-01
stricter address num criteria

sbma44:
Tightens up regex criteria used to match tokens as address numbers on address range features. In particular `1A` will continue to match but `1AB` will not. This is necessary to properly support UK postcodes -- e.g. `B77 1AA` -- when loaded alongside range features like `B77 ###`.

@yhahn would to hear your sense on whether included tests are adequate.

### DIFF
--- a/lib/util/termops.js
+++ b/lib/util/termops.js
@@ -467,7 +467,7 @@ function numTokenize(text, version) {
     var numTokenized = [];
     for (var i = 0; i < text.length; i++) {
         var replaced = text.slice(0);
-        num = parseSemiNumber(address(text[i]));
+        var num = parseSemiNumber(address(text[i]));
         if (num !== null) {
             replaced[i] = version >= 3 ?
                 numTokenV3(num.toString()) :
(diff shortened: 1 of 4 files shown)


lib/util/termops.js
```diff
@@ -467,7 +467,7 @@ function numTokenize(text, version) {
     var numTokenized = [];
     for (var i = 0; i < text.length; i++) {
         var replaced = text.slice(0);
-        var num = parseSemiNumber(text[i]);
+        num = parseSemiNumber(address(text[i]));
```
yhahn: What happened to `var` ...

yhahn:
@sbma44 last thing I would add is a unit test simulating the real world concrete failure you're seeing currently:

Indexes features like:

- Postcode: `B77 1AA`
- Address: `_text: B77, _cluster: { 1: { ... } }`

And then queries with `B77 1AA` and proves that postcode comes back first, with higher relevance than address feature.

sbma44:
:+1:

sbma44:
ok, new test created. present (and failing) in a branch off master at https://github.com/mapbox/carmen/blob/301ce0ab31d7623b5cab33caed228dae42f69c8d/test/geocode-unit.address-alphanumeric.test.js

PR #326, opened 2015-09-02
Isolinear chip

ingalls:
Change the indexing & internal carmen format to GeoJSON


ingalls:
We're getting there! Tests that are failing are:
- `index.test.js`
- `indexdocs-worker.js`
- `geocode-unit.gappy.js`

@sbma44 @ian29 if you want to grab one of these and start working through it let me know

ingalls:
Looks like there is some sort of bug with how adjacent tiles are being processed. Both `gappy` and `debug` are using the same test data and suffering from the same problems.

lib/pure/addresscluster.js
```diff
@@ -28,23 +26,22 @@ module.exports.reverse = function(feat,query) {
         lat = query[1];
 
     var cluster;
-    if (typeof feat._cluster === "string" )
-        cluster = JSON.parse(feat._cluster);
+    if (typeof feat.properties['carmen:addressnumber'] === "string" )
+        cluster = JSON.parse(feat.properties['carmen:addressnumber']);
```
yhahn: Could we remove this extra piece of logic from this refactor?

- `feature.transform` can handle the legacy logic for `_cluster` as a string => object => array
- any new geojson features must provide `carmen:addressnumber` as an array

lib/util/feature.js
```diff
@@ -1,13 +1,67 @@
...
+
+        doc.geometry = {
+            type: "MultiPolygon",
+            coordinates: zxys.map(function(zxy) {
+                return tilebelt.tileToGeoJSON([zxy[1], zxy[2], zxy[0]]).coordinates;
+            })
+        }
+    }
+    return doc;
+}
```
yhahn: In all three places you use `.map()` here you may want to switch to a straight for loop or descending while loop for speed.

You will probably want a bench js script for transform() against a pretty well-populated `range` or `addressnumber` object as well given how hot this codepath will be.

PR #327, opened 2015-09-08
[WIP] Templating index categories in carmen

heyitsgarrett:
This will have to be tweaked when Geojson formatting lands in Carmen shortly, but this is the first pass at how to directly template the returned result.

It allows an index to define how the `place_name` should be presented. Ex. an address index in USA might define its format as:

`{address.number} {address.name} {place.name}, {region.name} {postcode.name}, {country.name}`

which would return

`123 Fake Street Springfield, Illinois 45101, USA`

or alternately, for Germany:

`{address.name} {address.number}, {postcode.name} {place.name} {region.name}, {country.name}`

`Hufelandstraße 43, 10407 Berlin, Germany`

---

Template variables that aren't present in the returned feature are removed as before.


sbma44:
I like this but might not fully understand how context works. Other question for you: how does this interact with ranges, where a bunch of places have been combined into `eur` or `nam`?

sbma44:
from chat: idea here is to have non-address indexes fall back to current behavior (ordered feature lists, comma separated); and to add more compelxity to the if..then block to handle internationalization.

I'm not :100: on the non-address layer stuff yet -- a rev geocode for a point in China should go in reverse hierarchy order for `_en` and `_zh` IIRC -- but for now I think that problem can be safely folded into the bigger intlization question.

camilleanne:
So @heyitsgarrett and I were talking about how to have this support language localization, especially re: China. Right now only the default language/format scheme (whatever is represented in `carmen:text`) is covered by this templating.  We don't yet have a language flag that can be set at the query level (e.g. `?lang=en`) but I anticipate that there will be a language flag for the response.

1. `geocoder_address: geocoder_address_{language of default text}`

     `geocoder_address` would be a reference to the default language/template of another parameter with a language code `geocoder_address_en: {template for english}`. If the language requested in the language code is unavailable, it would default to `geocoder_address`, otherwise, it would grab `geocoder_address_zh` if `lang=zh` for instance.

2. nest the templates 

    ```
    geocoder_address: {
             en: {template for en},
             zh: {template for zh},
             default: 'en'
    }
    ```

3. Keep in on the api side -- api-geocoder would look for the language flag and have a template just for display. I'm not super clear on the effect of the `geocoder_address` field on indexing, would this hurt Chinese language queries since they are in reverse order (e.g. country -> region -> place)?

4. Split different languages for the same country into different indexes that have their own templates. I don't particularly care for this, but if it works...

looking for other thoughts @sbma44 @yhahn @ingalls @apendleton

yhahn:
I prefer 1. It's the easiest to iterate on without breaking changes (we start with `geocoder_address`, then we can add `geocoder_address_{code}` later on purely as an enhancement.

heyitsgarrett:
I've finished preliminary benchmarking against api-geocoder and results look good:

- [`templatequest` branch of carmen](https://gist.githubusercontent.com/heyitsgarrett/48e86d725495b70a7710/raw/e2149744332fa7a71323dbd2aaf713899068791a/benchmark-templatequest.md)
- [`master`](https://gist.githubusercontent.com/heyitsgarrett/48e86d725495b70a7710/raw/fc8a483e6695a83f9fff7d7f66a2b7995d52fa4d/benchmark-master.md)

Let me know if anything looks odd here.

I'm going to sync with @ingalls next on two things that need to happen in sync:
- Carmen version bump strategy here and final sanity check on code
- api-geocoder deploy with updated indexes

PR #331, opened 2015-09-16
All you can eat

ingalls:
Adds GeoJSON functionality to our unit tests as well as beginning the conversion to the new indexing format with `geocode-unit-address-format`

cc/ @yhahn @camilleanne


yhahn:
yummo

PR #332, opened 2015-09-17
Update to latest node-mapnik@3.4.6

springmeyer:
```
@yhahn: @gretacb - in short, with latest node-mapnik pbf decoding happens at `vt.query` such that `vt.query` will be slighly slower. To keep good speed it will be important to remove the use of `vt.parse` which serves no benefit and will only slow things down unnecessarily.

springmeyer [10:07 AM]
branch I’ll push here in a sec will do that

yhahn [10:07 AM] 
@springmeyer: what about gunzippin

springmeyer [10:07 AM] 
gunzipping happens during `setData`

springmeyer [10:08 AM]
so that price is only paid once

yhahn [10:08 AM] 
ok fab

springmeyer [10:08 AM] 
but… if `vt.query` is called multiple times for a given cached `vt`

springmeyer [10:08 AM]
then there may be a slight slowdown
```


GretaCB:
[Protozero bench testing](https://github.com/mapbox/carmen/blob/e8028ebc9d9f1139d7ecc08cec5ed2a17bb65c9d/bench/contextVector.js) versus master:
**master**
```
context vector x 234 ops/sec ±1.18% (80 runs sampled)
context vector x 235 ops/sec ±0.91% (86 runs sampled)
context vector x 229 ops/sec ±1.25% (80 runs sampled)
context vector x 232 ops/sec ±1.43% (80 runs sampled)
context vector x 236 ops/sec ±0.87% (87 runs sampled)
context vector x 236 ops/sec ±0.86% (82 runs sampled)
context vector x 235 ops/sec ±0.80% (83 runs sampled)
```

**protozero**
```
context vector x 238 ops/sec ±0.90% (82 runs sampled)
context vector x 233 ops/sec ±1.04% (85 runs sampled)
context vector x 234 ops/sec ±1.20% (82 runs sampled)
context vector x 234 ops/sec ±0.89% (81 runs sampled)
context vector x 233 ops/sec ±0.87% (86 runs sampled)
```

Relatively similar results :+1:

springmeyer:
I presume the reason this branch did not show perf gains is that the main event loop is pretty blocked up due to https://github.com/mapbox/carmen/issues/373

springmeyer:
> I presume the reason this branch did not show perf gains is that the main event loop is pretty blocked up due to #373

Nope. Further profiling shows that `vt.query` is taking >= 80% of the time. So the reason this branch shows no perf difference is that we're trying to measure an operation which is not adding significant time in general.

The biggest potential gain comes from #373

ingalls:
PR #398 Updates to mapik 3.4.15

springmeyer:
@ingalls - re-opening as this still needs to be merged into master. This branch was intended to upgrade the API to higher performance usage available in >= 3.4.6. So it contains improvements we want to use in general independent of other mapnik related changes. These are viable currently since master is alread at `~3.4.10` (https://github.com/mapbox/carmen/blob/b015287f60efc6c852fb12c498f55e6995283f14/package.json#L14).

springmeyer:
Latest bench confirms no slowdown from this upgrade in API usage (which is desirable):

## protozero branch
```
context vector x 197 ops/sec ±0.96% (81 runs sampled)
Memory -> peak rss: 55988224 / peak heap: 23734008
```

## master at b015287f60efc6c852fb1
```
context vector x 187 ops/sec ±0.77% (80 runs sampled)
Memory -> peak rss: 51425280 / peak heap: 19816256
```

springmeyer:
All tests are passing, this is ready to merge. Will make #415 fixes a lot easier.

PR #333, opened 2015-09-21
get benchmarks working

camilleanne:
to do:
 * [x] get these running in travis
 * [ ] ~~set acceptable thresholds~~ punting
 * [ ] ~~review whether decollide needs to be a thing~~ next pr
 * [x] fix `verifymatch.verifyFeatures` bench

spits out something that looks like https://gist.github.com/camilleanne/782756f3e6ab6b910be8 in travis. 

benchmarks can be run with `npm run bench`


camilleanne:
Punting on setting acceptable thresholds for now -- It's been a challenge to determine what that would look like and how those might be stored.

PR #334, opened 2015-09-22
Enforce ESLint

ingalls:


mourner:
:+1: but I'd suggest a stricter config, and extending `eslint:recommended`.

ingalls:
@mourner yeah I'm hoping to talk this over with the gang and get a sense of what we want to enforce. This was the default from `mapbox-gl-js`

mourner:
@ingalls I didn't update the config yet after ESLint 1.0 (which ditched rules by default and introduced eslint:recommended instead), a more recent config can be found e.g. here https://github.com/mapbox/geojson-vt/blob/master/.eslintrc

PR #336, opened 2015-09-23
Better Types [WIP]

ingalls:


yhahn:
Add a test for this option in https://github.com/mapbox/carmen/blob/master/test/bin.test.js ?

PR #340, opened 2015-10-01
Get rid of that smarmy sphericalmercator in context

yhahn:
This is the easiest solution I think since -- previously we were relying on tile-cover and sphericalmercator to agree on these tile edge cases and it's better to just use the one of them.

cc @camilleanne


camilleanne:
:+1:

ingalls:
Moved to #341

PR #341, opened 2015-10-02
Versions Are Cheap geocoder_format

ingalls:
Captured from chat

```
ingalls [22:09]
unintended side affects
basically really really early on the geocoder_address flag is used to determine if phrasematch should generate address permulations for a given source
for most layers this doesn’t really matter
as places/regions etc
don’t have numbers in them

heyitsgarrett [22:09] 
right

ingalls [22:09] 
so the address permutations fail and are then ignored
for postcode however
the address permutations ​_can_​ be generated
as a number is found
that is a potential housenumbers
but when it goes back to actually look up this permutation in the index
it can’t find a housenumber that matches
because the postcode indexes don’t have the additional housenumber info
which results in the postcode index hard failing and returning no results
```

cc/ @heyitsgarrett @sbma44

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,7 +56,8 @@ attribute               | description
 ------------------------|------------
 maxzoom                 | The assumed zoom level of the zxy geocoder grid index.
 geocoder_layer          | Optional. A string in the form `layer.field`. `layer` is used to determine what layer to query for context operations. Defaults to the first layer found in a vector source.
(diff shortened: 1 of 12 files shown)


ian29:
ooh @incanus @heyitsgarrett this means we need to preserve the info in [addrType](https://github.com/mapbox/mapbox-places/blob/master/scripts/addrType), right?

i wonder if carmen can parse the whole `geocoder_format` template to get the right address name/num order so we dont have to store address format twice (once in `geocoder_address` and once in `geocoder_format`)

ingalls:
No this just means the build script in mapbox places will need to set
geocoder_asdress:1 when the extid is address. The formatting of addresses
will now be in geocoder_format

ian29:
oh cool. :+1:

PR #343, opened 2015-10-07
add language flag

camilleanne:
Adds the ability to pass a language flag (an iso language code verified by [model-un](http://github.com/mapbox/model-un)) to Carmen.

https://github.com/mapbox/carmen/pull/319 paves the way for queries in local languages, this opens up returning a response in a specified language when available in the index. Defaults to whatever is in `carmen:text`.

Also supports custom templates in indices by language, e.g. `"geocoder_format_zh": "{country._name}{region._name}"`

to do:
* [x] test for context values

### DIFF
--- a/lib/geocode.js
+++ b/lib/geocode.js
@@ -44,7 +44,7 @@ module.exports = function(geocoder, query, options, callback) {
 
     // check that language code is valid
     if (options.language) {
-        options.language = mu.hasLanguage(options.language) ? options.language : false;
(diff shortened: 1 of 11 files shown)


lib/geocode.js
```diff
@@ -41,6 +42,11 @@ module.exports = function(geocoder, query, options, callback) {
             return callback(errcode('Proximity lat value must be a number between -90 and 90', 'EINVALID'));
     }
 
+    // check that language code is valid
+    if (options.language) {
+        options.language = mu.hasLanguage(options.language) ? options.language : false;
```
yhahn: Could we return an `EINVALID` error if the language specified doesn't match one in `model-un`?

PR #345, opened 2015-10-07
Doff thy name unidecode cxx

apendleton:
@camilleanne here are the changes to use the cxx version of unidecode (though it should probably go onto npm at some point)


apendleton:
Okay, I think this is good to go

yhahn:
:ship: (minor version bump since there's a dep change?)

apendleton:
@yhahn we're already behind master version-wise (8.1.0 in this branch vs. 8.1.1 in master). Do I skip ahead to 8.1.2?

camilleanne:
@apendleton :shipit:

PR #348, opened 2015-10-14
Universal Geocoder_tokens

ingalls:
Moves `geocoder_tokens` from the token level to the query level.


ingalls:
Closed in favour of better approach in #356

PR #350, opened 2015-10-15
Don't POInt at me [WIP]

ingalls:
- Add new `geocoder_type` parameter which allows for different types of indexes to be grouped together in a `geocoder_name` field. 

This ensures that if there are two bottom level indexes (ie poi & address) they can retain their type but be mutually exclusive.

Prior to this the lowest level index would always win, regardless of proximity to the context geocode.

**TODO**
- [x] More tests


yhahn:
- Test (and functionality) for filtering based on type rather than name when defined

ingalls:
- [x] Fix current broken tests
- [ ] add polygon context tests for new geom sort
- [ ] add/update context bench
- [x] redeploy to api-geocoder branch for testing/staging See https://github.com/mapbox/api-geocoder/pull/632

PR #352, opened 2015-10-16
expose indexes

sbma44:
adds an `indexes` property to result objects for forward and reverse geocodes


yhahn:
Could this be optional, much like the debug option? (`options.indexes = true;`)

sbma44:
it could! I was gonna ask if this merited a major version change. if this becomes a flag I think the answer is no. will implement first thing tmw.

PR #354, opened 2015-10-21
Filter first

ingalls:


lib/context.js
```diff
@@ -20,7 +20,7 @@ var cover = require('tile-cover');
 // @param {Float} lat: input latitude
 // @param {Object} options: optional options object
 // @param {Function} callback
-module.exports = function(geocoder, lon, lat, options, callback) {
+module.exports = function(geocoder, queryOpts, lon, lat, options, callback) {
```
yhahn: Why not just use `options` for this? `geocode` could be selective about what query options to pass in?

PR #370, opened 2015-11-06
Setup coveralls

ingalls:


package.json
```diff
@@ -9,9 +9,11 @@
     "carmen-cache": "0.10.0",
     "err-code": "0.1.2",
     "eslint": "^1.5.0",
+    "istanbul": "^0.4.0",
```
yhahn: This should be a devDep, no?

ingalls: @yhahn yup, missed that, thanks

ingalls:
cc/ @mapbox/geocoding-gang 

Awesome job folks, we're already at 96% coverage without even actively tracking it! Woot!

PR #371, opened 2015-11-07
expose non-carmen properties in context array

sbma44:
Country codes are a common feature request for geocoding. These are easy enough to expose in an country document's custom properties. However, it will often be desirable to understand the country that a non-country-level result resides in. Doing so will require more information to be exposed in a result's `context` array.

This PR exposes the `iso3166` custom property of `country` objects in a result's `context` array (when it exists).

This is a straightforward implementation and follows the precedent of hardcoding which properties get exposed in `context`. But I'm not sure it's the best or most flexible approach. 

An alternative would be to define a new Carmen GeoJSON property that indicates which properties should be exposed in `context`. For example:

```
properties: {
    "carmen:center": [0,0],
    "carmen:text": "Canada",
    "carmen:extid": "country.1"
    "carmen:contextProperties": ["iso3661"],
    "iso3661": "ca"
}
```

This bulks up the documents, but for this layer it's unlikely to be a big issue.

Happy to go either way on this (or a third I'm not seeing yet). I can see arguments for both approaches.

### DIFF
--- a/lib/context.js
+++ b/lib/context.js
@@ -234,6 +234,15 @@ function loadFeature(source, feat, full, query, language, callback) {
         } else {
             loaded.properties['carmen:text'] = properties['carmen:text'];
         }
+
+        // copy non-carmen:* properties
+        var propertyKeys = Object.keys(properties);
+        for (var propi = 0; propi < propertyKeys.length; propi++) {
+            if (!/^carmen:/.test(propertyKeys[propi])) {
+                loaded.properties[propertyKeys[propi]] = properties[propertyKeys[propi]];
+            }
+        }
+
         return callback(null, loaded.properties['carmen:text'] ? loaded : false);
     }
     feature.getFeatureById(source, feat.id, function(err, loaded) {
--- a/lib/context.js
+++ b/lib/context.js
@@ -238,7 +238,7 @@ function loadFeature(source, feat, full, query, language, callback) {
         // copy non-carmen:* properties
         var propertyKeys = Object.keys(properties);
         for (var propi = 0; propi < propertyKeys.length; propi++) {
-            if (!/^carmen:/.test(propertyKeys[propi])) {
+            if (!/^(carmen:|id)/.test(propertyKeys[propi])) {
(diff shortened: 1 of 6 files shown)


ingalls:
@sbma44 Definitively all for getting ISO codes into the geocoder. I'm afraid I'm not sold on this approach however. Since Carmen is at the moment source/index agnostic in that it does not care about the index id/name/group of the index I think hardcoding non address type values is something we should avoid at all costs.

Both `country` and `ISO3166` are values we use in our internal implementation but these values are passed into it by the server, Carmen itself has no expectation that a `country` index will exist.

Definitely agree that the second option is a bit messier but I think it is closer to the level of index agnosticism that we need to maintain. I'm going to mull on this tonight and will hopefully have some more ideas tomorrow.

sbma44:
Fair enough! I appreciate the eyes. And yeah, I agree: it's gross. I should be able to bang out option 2 pretty quickly if it looks like a decent path forward tmw morning.

twbell:
@sbma44 we can expect similar requests to overload the response packet in the near future.  While I would hope that we want to be very critical of each and avoid a kitchen-sink approach (e.g. opencage), I would support an angle of attack that allows us to easily include additional attributes, on their merits.

sbma44:
new behavior: all feature properties not prefixed with `carmen:` make it into the `context` entries.

lib/context.js
```diff
@@ -234,6 +234,15 @@ function loadFeature(source, feat, full, query, language, callback) {
         } else {
             loaded.properties['carmen:text'] = properties['carmen:text'];
         }
+
+        // copy non-carmen:* properties
+        var propertyKeys = Object.keys(properties);
+        for (var propi = 0; propi < propertyKeys.length; propi++) {
+            if (!/^(carmen:|id)/.test(propertyKeys[propi])) {
```
yhahn: This will excluding anything that starts with `id`, no? Like `ideas` or `idaho_potatoes` ?

sbma44: great point. I wrote that in haste when the bug popped up at the end of the day. Will adjust behavior & write a test.

PR #378, opened 2015-11-12
Back to the future

ingalls:
Update unit tests from legacy index format to new GeoJSON compliant format.

cc/ @yhahn @sbma44


yhahn:
:+1:

PR #379, opened 2015-11-12
Add geocoder_stack test [WIP]

ingalls:
One of our most requested features is the ability to filter results by a given country. 

This PR exposes a `geocoder_stack` parameter which allows each index to be assigned to a given stack (in many cases an ISO 3166-1 alpha 2 code). 

Passing in the `stack` parameter at query time will restrict the results to those that are coming from the given stack.

cc/ @mapbox/geocoding-gang


sbma44:
if I'm reading this right, a query can ask for multiple stacks but an index can only belong to one. Is this desirable? I could imagine assigning an index to both `opendata` and `US`, as a trivial example.

Thinking about this more, could we deprecate the `type` param entirely if both `index.info` and the `query` params support arrays? doing so would admittedly complicate AND and OR behavior...

ingalls:
Rebased with master - DAWG

ingalls:
### TODO

- [ ] Discovered that having a `geocoder_stack` on the index level but not on the feature level results in 0 results. 
- [ ] Test no index level result but feature level

lib/verifymatch.js
```diff
@@ -36,7 +36,19 @@ function verifymatch(query, stats, geocoder, matched, options, callback) {
...
+            }
+            if (options.stacks && source.stack) {
+                for (var j = 0; j < source.stack.length; j++) {
+                    if (options.stacks.indexOf(source.stack[j]) !== -1) dropRes = false;
+                }
+            } else if (options.stacks) {
+                dropRes = false;
+            }
+
+            if (!dropRes) filtered.push(results[i]);
```
yhahn: It would be awesome to be able to unit test this flow -- could it be moved to a function?

ingalls: agreed, this was the source of a couple of initial bugs, will do

ingalls: Unit testing revealed that when using the `stacks` and `types` filtering options they don't work in unison. Working on a fix.

yhahn:
@ingalls :+1: 11.2.x?

PR #383, opened 2015-11-20
[yo-dawg] Unobject encodePhrase

yhahn:
This rolls back the phrase-as-object concept bringing `yo-dawg` a bit closer to `master`. (https://github.com/mapbox/carmen/compare/yo-dawg-patchtext)

I'm realizing the reason I knee-jerked in this direction is because at runtime we already have the `subquery` object that serves a very similar concept to the phrase object currently in the `yo-dawg` branch.

Should pass tests locally -- not passing on circle for build-ey reasons.

cc @apendleton for your review

### DIFF
--- a/test/dawgcache.test.js
+++ b/test/dawgcache.test.js
@@ -23,12 +23,12 @@ tape('dump/load', function(assert) {
             assert.ifError(err);
             var loaded = new DawgCache(data);
             for (var i = 1; i <= 4; i++) {
-                assert.equal(loaded.hasPhrase({text: "a" + i, degen: false}), true, 'has a' + i);
+                assert.equal(loaded.hasPhrase({text: "a" + i, ender: false}), true, 'has a' + i);
             }
-            assert.equal(loaded.hasPhrase({text: "a5", degen: false}), false, 'not a5');
+            assert.equal(loaded.hasPhrase({text: "a5", ender: false}), false, 'not a5');
 
-            assert.equal(loaded.hasPhrase({text: "a", degen: false}), false, 'not a');
-            assert.equal(loaded.hasPhrase({text: "a", degen: true}), true, 'has a as degen');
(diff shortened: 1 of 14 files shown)


test/dawgcache.test.js
```diff
@@ -23,13 +23,13 @@ tape('dump/load', function(assert) {
...
             var loaded = new DawgCache(data);
             for (var i = 1; i <= 4; i++) {
-                assert.equal(loaded.hasPhrase({text: "a" + i, degen: false}), true, 'has a' + i);
+                assert.equal(loaded.hasPhrase({text: "a" + i, ender: false}), true, 'has a' + i);
             }
-            assert.equal(loaded.hasPhrase({text: "a5", degen: false}), false, 'not a5');
+            assert.equal(loaded.hasPhrase({text: "a5", ender: false}), false, 'not a5');
 
-            assert.equal(loaded.hasPhrase({text: "a", degen: false}), false, 'not a');
-            assert.equal(loaded.hasPhrase({text: "a", degen: true}), true, 'has a as degen');
```
apendleton: Should the booleans be flipped? I'd expect things that are degen: true to be ender: false, no?

apendleton: (this isn't commenting on quite the right thing so only part of the diff is showing up, but I meant for this to be about the addition of ender to replace degen here)

yhahn: No it's weirdly the same now that I think about it. `ender` means that the subquery is at the end of the query and is elligible for degen lookup. If it's not an `ender` it's been typed past and can't be continued/modified by further typing.

:confounded: confusing I know!

apendleton:
Yeah, I didn't use subquery before because I (wrongly, as it turns out) assumed that there was a reason to keep both the pre- and post-unidecode text around, so made a new thing to encapsulate the ID and the post-unidecode text.

PR #386, opened 2015-11-22
[yo-dawg] Back to dictsize

yhahn:
- Rolls back to `geocoder_dictsize` as option key for bitcache,
- Goes with `geocoder_dicttype` as option key for determining type,
- Moves env var from main carmen index loading to `api-mem` source so it can only really affect tests.

cc @apendleton


apendleton:
Closing in favor of https://github.com/mapbox/carmen/pull/390

PR #388, opened 2015-11-23
[notready] [yo-dawg] dawg as only dictcache type

yhahn:
In theory we may be able to use the dawg cache as our primary + only cache type if we're ok with some waste i/o on early cjk unicode + autocomplete scenarios.

Next actions before making the call:

- [ ] More IRL testing with real indexes

cc @apendleton


apendleton:
This makes me super nervous, mostly because of the unknowns about CJK. Your Slack summary matched my understanding -- that extra DAWG matches on autocomplete won't break anything exactly, but will lead to extra IO just like other collisions. So in that sense I think it's fine, but it seems to me like there could be lots of extra IO autocompleting on, say, just the letter "a." Of course that's testable...

But even if dawg ends up being better than bitcache in all instances, I think there's also value in the modularity, in case there's something with better behavior for CJK than dawg.

apendleton:
Closing in favor of https://github.com/mapbox/carmen/pull/390

PR #395, opened 2015-12-21
Replicant

ingalls:
- Allow up to 10 forward geocodes (default 5) specified by the limit param
- Allow up to 5 reverse geocodes (default 1) when only querying for a single type


ingalls:
The proximityVector pass (used when a limit > 1) for reverse geocoding, uses the `carmen:center` for subsequently building a context stack for each result. This result works well for polygons & POIs but does not currently support address interpolation/clustering. 

**TODO**
- [ ] Add clustering support to proximityVector
- [ ] Add interpolation support to proximityVector

ingalls:
Had a great chat with @springmeyer. To implement this feature optimally its going to take some changes in node-mapnik to return the vertex that the query hits along with the currently returned `layer` & `distance`. This will allow the `context.js#proximityVector` to aggregate the lon/lats to build a context array of without having to decode the geometry.

springmeyer:
@ingalls - node-mapnik work started at https://github.com/mapnik/node-mapnik/pull/575. You can test this - via installing from a binary - by pointing to that branch:

```js
"mapnik": "https://github.com/mapnik/node-mapnik/tarball/hits",
```

If you fix any bugs in the C++ code while I'm not around you can regenerate binaries by adding `[republish binary]` to the commit message when pushing to that branch. But otherwise feel free to start adding tests to see if things behave as you hoped.

ingalls:
Closing in favour of the approach outlined in #398

PR #398, opened 2016-01-04
Caprica six

ingalls:
Refs: https://github.com/mapbox/carmen/pull/395

- Allow up to 10 forward geocodes (default 5) specified by the limit param
- Allow up to 5 reverse geocodes (default 1) when only querying for a single type

Where #395 decodes LineString & MultiPoint geometries and then creates individual geometries at `context` time, this branch decodes linestrings & multipoint geomtries at index time. Indexing a single MultiPoint geometry but splitting out into unique geomtries in the vector tiles.


test/address.test.js
```diff
@@ -21,7 +21,7 @@ test('address.reverse', function(assert) {
...
...24', '2500', '1900', '2700', '2200', '', '1720', '2600' ], 'carmen:ltohn': [ '1620', '2798', '3020', '3098', '2998', '1718', '1698', '', '1616', '2398', '2898', '2522', '2448', '2098', '1998', '1818', '1898', '2498', '2198', '2598', '2510', '1922', '2748', '2298', '', '1798', '2698'], 'carmen:parityl': [ 'E', 'E', 'E', 'E', 'E', 'E', 'E', '', 'E', 'E', 'E', 'E', 'E', 'E', 'E', 'E', 'E', 'E', 'E', 'E', 'E', 'E', 'E', 'E', '', 'E', 'E' ], 'carmen:parityr': [ 'O', 'O', 'O', 'O', 'O', 'O', '', 'O', 'O', 'O', 'O', 'O', 'O', 'O', 'O', 'O', 'O', 'O', 'O', 'O', 'O', 'O', 'O', 'O', 'O', 'O', 'O' ], 'carmen:rangetype': 'tiger', 'carmen:rfromhn': [ '1623', '2701', '3001', '3033', '2851', '1701', '', '2751', '1601', '2301', '2801', '2501', '2351', '2001', '1921', '1801', '1825', '2401', '2101', '2511', '2427', '1901', '2651', '2201', '2565', '1721', '2601' ], 'carmen:rtohn': [ '1699', '2749', '3031','3099', '2999', '1719', '', '2799', '1621', '2349', '2849', '2509', '2399', '2099', '1999', '1823', '1899', '2425', '2199', '2563', '2499', '1919', '2699', '2299', '2599', '1799', '2649' ], 'carmen:text':'14th St NW', 'carmen:tmpid': 323909959, 'carmen:zxy': [ '14/4686/6266', '14/4686/6265' ] } });
```
ingalls: This looks like a scary change, it's not - simply changing `_id` to the correct `id`.

lib/geocode.js
```diff
@@ -131,13 +136,72 @@ function reverseGeocode(geocoder, tokenized, options, callback) {
...
-        full:true,
-        maxidx: maxidx,
-        types: options.types,
-        stacks: options.stacks
-    }, function(err, context) {
-        if (err) return callback(err);
+
+    if (options.limit > 1) {
+        context(geocoder, queryData.query[0], queryData.query[1], {
+            full: false,
```
yhahn: Hrm, did you leave the `maxidx` option here on purpose?

ingalls: Since the `types` field must be true for a limit>1 in reverse, `maxid` is optional

package.json
```diff
@@ -11,7 +11,7 @@
     "err-code": "1.1.1",
     "eslint": "^1.5.0",
     "locking": "2.0.2",
-    "mapnik": "~3.4.10",
+    "mapnik": "3.4.16",
```
yhahn: Can you use the `~` here so that when mapnik 3.4.17+ comes out we don't need to bump?

ingalls: :+1: Done in 9356f92

PR #400, opened 2016-01-06
Quantum slipstream drive

ingalls:
Implements the beginnings of a streaming indexer interface for carmen.

**TODO**
- [x] Ensure SlipStream drive produces identical index as legacy propulsion 
- [x] Ensure shuttle is backwards compatible with legacy propulsion
- [x] Capture SlipStream output in readable stream (Blocking #398) Basically allow the indexer GeoJSON to be different than the GeoJSON in the vector tiles


scripts/carmen-index.js
```diff
@@ -4,24 +4,82 @@ var fs = require('fs');
...
+    console.log('carmen-copy.js --config=<path> --index=<path> [options]');
+    console.log('[options]:');
+    console.log('  --help                  Prints this message');
+    console.log('  --version               Print the carmen version');
+    console.log('  --config="<path>"       Path to JSON document with index settings');
+	console.log('  --index="<path>"        Tilelive path to output index to');
+    console.log('  --ouput="<path>"        Path to output Normalized GeoJSON to');
+    console.log('                            If unset outputs to STDOUT');
+    console.log('  --input="<path>"        File containing GeoJSON FeatureCollection to index');
+    console.log('                            If unset waits for FeatureCollection from STDIN');
```
yhahn: Hrm do we need these input/output options? Seems like cat + stdout redirection should be fine?

ingalls: hmm I've always been an anti `cat |` user after reading http://mywiki.wooledge.org/BashGuide/Practices

yhahn: > If the manual of the program does not specify any way to do this, you should use redirection

Fine you could do this too : P

ingalls: Haha fair fair

lib/indexer/indexdocs.js
```diff
@@ -39,6 +41,19 @@ function indexdocs(docs, freq, zoom, geocoder_tokens, callback) {
...
+
+                    if (!options.firstWrite) {
+                        options.output.write(',');
+                    } else {
+                        options.firstWrite = false;
+                    }
+                    options.output.write(JSON.stringify(patch.vectors[docs_it]));
+                }
+            }
+
```
yhahn: Does this have to happen here? Couldn't it happen back in the caller who has access to the stream so this could remain unchanged?

PR #404, opened 2016-01-12
Line Delimited Streaming JSON

ingalls:


lib/index.js
```diff
@@ -65,13 +113,20 @@ function index(geocoder, from, to, options, callback) {
...
     //   final in-memory frequency count of all terms involved with this set of
     //   documents to be indexed.
     // - Stores new frequencies.
+
+    if (!options) return callback(new Error('options argument requied'));
+
+    //Legacy API - deprecate on next major version
+    //             Previously options was zoom - an integer 0-14
+    if (typeof options === 'number') options = { zoom: options };
+
```
yhahn: Why options couldnt this just be zoom?

yhahn: nm I see : )

ingalls: :+1: :)

PR #406, opened 2016-01-15
NCC-1701-J (Dep Update)

ingalls:
Update carmen to Node 4.x.x


ingalls:
Since tilelive-s3 will have to be updated before nodev4 can go up this branch is a general dep update branch.

PR #412, opened 2016-01-25
avoid massive callback stacks

ingalls:
We've seen several errors around `push.apply` generating massive callback stacks. This has been resolved twice before in carmen by migrating to the faster and non-callback based approach of `concat`

cc/ @yhahn @karenzshea


ingalls:
Going to limit query length + token length instead.

PR #413, opened 2016-01-25
Ensure query lengths don't cause stack errors

ingalls:
- Enforce max query length of 256 chars
- Enforce max token length of 20 tokens
- Unit tests

cc/ @karenzshea @yhahn @ianshward


ingalls:
screenshot 2016-01-25 11 24 45

Rebuilt and coveralls is now passing but didn't update?

PR #422, opened 2016-02-08
Only run tests once

apendleton:
Don't run carmen tests twice with different dictcache env vars; those don't do anything anymore

@yhahn can you confirm that this is kosher?

cc @ingalls


yhahn:
:ship:

Issue #424, opened 2016-02-09
Incorrect interactions between unidecode and token replacement


apendleton:
I just dug into an issue where a query for the string "Rua Oratório" returned a correct result and "Rua Oratorio" did not, and several issues came up. In general, we handle squashing accented characters in inbound queries correctly, but failed this time. It turns out the token replacement operation is being performed on text pre-unidecode, but the regular expression for identifying words is "\W+" which maps to [^a-zA-Z_0-9] and thus excludes accented characters. Consequently, the terminal "rio" in the accented version matched a predefined Portuguese substitution "rio" => "r", which transformed "Oratório" into "Oratór" and then into "orator" (after lowercase and unidecode). This didn't happen in the unaccented version, which was left free of transformations. As the index-time string also had the accented character, the same behavior occurred at index time as at search time with the accented character, and that version matched, whereas the unaccented version didn't.

So, a few problems:
- The word-boundary regex in the token replacer fails to account for unicode characters, but unicode strings are being passed in
- The token replacement code generally doesn't assume unicode characters in other ways; the to- and from- replacer strings are unidecoded when the token replacer is built, for example. This means all replacement strings with accented characters currently don't work as intended, because the index-time strings still have their accents whereas the to/from strings have had their accents stripped
- Conversely, if end-user queries omit accents, they _do_ have replacements applied even though they weren't applied at index time

A lot of this probably stems from [moving the unidecode operation later in the process](https://github.com/mapbox/carmen/commit/7b0083f) to fix degen generation for Chinese characters, I think? So moving unidecode earlier in the process is probably not straightforward.

@ingalls and I talked it through and each have ideas, but would be curious to hear others' thoughts.

cc @yhahn @sbma44 @camilleanne

apendleton:
Concretely as to a fix, three directions we could go:
* **Option 1:** Keep unidecode late as it is, and fix the token replacement's handling of unicode text. Specifically, first I think this would involve changing how word boundaries are represented in the token replacement regex, probably to be a positive set rather than a negative one, so rather than no letters or numbers, it should be all whitespace and punctuation. Second, for the from- and to- replacement strings, for all the ones that contain accented characters, add them twice, once with accents and once without, so they work the same way regardless of whether user queries have accents or not.  This is probably my current inclination.
* **Option 2:** move the unidecode step back to earlier in the process and implement the degen logic around Chinese characters some other way.  I don't really know how this would work.
[...]

PR #425, opened 2016-02-10
Fix replacer

apendleton:
Addresses the word boundary issue that came up in #424 .

### DIFF
--- a/lib/util/token.js
+++ b/lib/util/token.js
@@ -1,6 +1,9 @@
 var unidecode = require('unidecode-cxx');
 var XRegExp = require('xregexp');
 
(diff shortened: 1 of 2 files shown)


ingalls:
@aarthykc and I fixed the str=>strasse issue. Going to close this as stale - feel free to reopen if additional changes are needed here.

apendleton:
@ingalls this wasn't the straße issue. It was this issue https://github.com/mapbox/carmen/issues/424 which was low priority but, I don't think, ever got fixed. I'll dig up the original test case that prompted it this afternoon, though, and confirm that it's still a problem, and reopen if so.

PR #430, opened 2016-02-12
Expose _commit

ingalls:


ingalls:
Please don't delete this branch yet! Doing a rebuild which still points at it. Will update this coming Monday to point to master

Issue #433, opened 2016-02-20
tighten behavior around cross-charset unidecoded false positives


sbma44:
Everything in Carmen indexes is transliterated by the `unidecode` module into a string of ASCII characters. As additional language data is loaded, more collisions are becoming apparent. One recent example:
- `Alberta` is a province in Canada
- it's called `アルバータ州` in Japanese
- unidecode transforms this to `arubataZhou`
- ... which matches an autocomplete (default) query for "Aruba"

Unidecode performs several functions for us:
- it's a lossy hash to collapse unicode into more compact ASCII
- it cleans up accents on latin character sets
- it provides _some_ transliteration capability, allowing queries across character sets

Preserving this third behavior looks increasingly tenuous--there are probably better long-term approaches to that problem. But that's a larger conversation.

In the short term, we need to reduce our incidence of false positives, particularly for CJK terms. @apendleton suggested an elegant approach: detect the string character set, then prefix it with a letter (say, `C` for `CJK`, `M` for mixed/everything else).

We could separate out other character sets like Cyrillic but a) haven't yet observed problematic behavior with it b) think it might work better for transliterations than CJK. 

This will require a global reindex. Also, character set detection may not be trivial.

cc @yhahn @apendleton @ingalls @camilleanne

apendleton:
Copying what I added in chat:

> Actually, thinking more, I think the super-simple solution that combines several already-proposed ideas: we pick a single ascii character that will never occur in unidecode output, and if the source string contains any characters in complex alphabets, we prefix with that before storing in or looking up from the dawg, so then we have two trees, basically.
> 
> I think this could actually be pretty fast to hack together, and English/otherwise-Latin text stays the same in the dawg as it is now.

The last property would help shield us from any unforeseen behavior changes this might introduce (though I don't think there would be any).

apendleton:
Also, for alphabet detection, haven't found anything for node, but this Python library does what we want: https://github.com/EliFinkelshteyn/alphabet-detector so might be worth figuring out how it works.

tmcw:
Alphabet detection... someone shine the @ericfischer bat signal

ericfischer:
@tmcw The one I've used before is https://github.com/mikemccand/chromium-compact-language-detector. (The old version that I used was https://github.com/ericfischer/chromium-compact-language-detector; I'm not sure what has changed since then.)

Or are you just wanting to know script, not language? Just the Unicode block of the alphabetic characters ought to be enough for that.

[...]

PR #434, opened 2016-02-21
[ready] unidecollide

sbma44:
refs https://github.com/mapbox/carmen/issues/433


sbma44:
WIP -- I see problems with `termops.getIndexablePhrases.test.js`. Also prob going to need to consult about what the `waste` property does so I can be sure my excisions from `geocode-unit.dict-collision.test.js` are defensible.

apendleton:
@sbma44 did you see https://github.com/mapbox/carmen/issues/433#issuecomment-186464488 ? I think we should use prefix bytes that won't ever be in text that comes from unidecode, rather than x and z. That way, there's no way for collisions to happen with typed text, because the only way for those bytes to make it into the strings will be because we put them there.

sbma44:
Yeah, I saw it but after I'd worked on the implementation. I considered using something like that but felt enough uncertainty around carmen string-cleaning and DAWG internals not to mess with them. I'm not averse to redoing this (though I do have a reindex going already), but:

> there's no way for collisions to happen with typed text

Is there a way for collisions to happen with the current implementation? I didn't think so but am now wondering if the way input queries are tokenized could introduce collisions. For example a feature `House of XXXs` transformed to `XHouse of XXXs` and a tokenized query for `house of xxs` turning into `xhouse` `xof` `xxxs` -- not sure if that's a correct description, though.

Relatedly, do you know offhand which ASCII symbols would be safe or should I go hunting?

mtmail:
You can try lower non-printable ASCII character. See caret notation on https://en.wikipedia.org/wiki/ASCII

In Javascript
`node -e 'console.log("abc\001def")'`
pipe into `more` and you'll see an `^A` (single character).

In UNIX terminals you can type them in as CTRL-V, then CTRL-A, useful if you want to run `grep` or `tr`. Sadly survive copy&paste.

I used to work at Yahoo Search and we dealt with every query string imaginable. Those characters were great delimiters in CSV files.

sbma44:
I think that's a promising suggestion but I'm specifically curious about which chars are safe from a carmen perspective. All else being equal I'd prefer to use underscores or something similar (or possibly just capital letters, which they currently are) to facilitate debugging.

apendleton:
At this point in the chain most carmen processing has already occurred. We could also trivially move the prefixing to even later in the process (basically, at the point of dawg insertion).

> Is there a way for collisions to happen with the current implementation?

I didn't catch that we were using uppercase characters and counting on carmen to lowercase everything... unidecode definitely sometimes outputs uppercase characters, though, and the order of operations there isn't clear to me, but I think unidecode happens after case normalization, so uppercase characters could sneak into the final string, maybe? I'm not sure.

> Relatedly, do you know offhand which ASCII symbols would be safe or should I go hunting?

Hm, so I actually might have been wrong in my initial thoughts about this being workable. It looks like unidecode actually preserves all the 7-bit ASCII characters exactly as-is. There'd maybe be some possibilities of using "high ascii" characters (>= \x80), but node doesn't seem to like having a string with just "\x80" in it (it pads it), probably because it's not valid UTF8. The dawg would still be okay with it, but we'd have to insert it with a buffer, so it might not be worth the bother.

(Or we can just later the unidecode table, I guess)

sbma44:
@apendleton I've poked around a bit more and expanded tests further, but I can't produce the kind of accidental collision behavior hinted at above. I'm back to thinking the addition of the extra `X` or `Z` is safe even though it's not a reserved character. Can you suggest a test case that might prove otherwise? I can't say I feel as confident about this as I'd like...

apendleton:
The key would be to find a character from a non-CJK character set that unidecode translates into a string that begins with a capital Z, and also that doesn't get mutated itself by carmen in the lowercase operation (so I think "Z̆" should be safe even though it unidecodes to "Z" because it also lowercases to "z̆"). It's plausible to me that such a character exists, but it would take spelunking through the unidecode data files and testing against your charset regex to find something that interacts badly with the combination. I guess it's probably not worth worrying about.

sbma44:
Hm. As things stand everything is lowercased post-unidecode: https://github.com/mapbox/carmen/pull/434/files#diff-3c11f53e589cc6fc90ecc902a6ec4af4R69

This includes the leading `X` or `Z`. So we're not relying on capitalization to make these reserved chars My current belief is that collisions are impossible even though the prefixing character will be present in other parts of the term (including the beginning) because the prefixing is consistently applied to both input and indexed text (and my efforts to poke holes in this behavior have so far failed).

apendleton:
Ah, alright. I misremembered the order. I withdraw my objection/concern/whatever.

lib/util/termops.js
```diff
@@ -64,7 +65,8 @@ function terms(tokens) {
  * @param {Array} tokens Tokenized array of user's input query
  */
 function encodableText(tokens) {
-    return unidecode(typeof tokens === 'string' ?  tokens : tokens.join(' ')).toLowerCase().trim().replace(/\s+/g, ' ');
+    var text = (typeof tokens === 'string') ?  tokens : tokens.join(' ');
+    return unidecode((nonCJKchar.test(text) ? 'X' : 'Z') + text).toLowerCase().trim().replace(/\s+/g, ' ');
```
yhahn: @sbma44 do you want to make these lowercase explicitly to remove any confusion about whether we're relying on the uppercase behavior of these prefixes (we're not).

sbma44: :+1:

sbma44:
per chat I'm going to finish building new indexes using this code and evaluate their impact on result quality before we consider a merge.

sbma44:
rebuilding indexes was stymied by a now-hopefully-squished (and unrelated) bug in carmen. rebuilding, hopefully they'll finish promptly.

sbma44:
It wasn't prompt, but the indexes have rebuilt and passrates look good (though indexing still appears to be iffy to me, for unrelated reasons). I am going to merge and bump major versions once Circle passes.

PR #435, opened 2016-02-23
make sure timer for getIndexableDocs gets started

mtmail:
Inside `indexDocs()` there's a `console.timeEnd` which fails if this timer was never started.

Of course you may chose to delete the timer or the if-clause completely since it applies less to the streaming interface that's now in use.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -58,13 +58,13 @@ function index(geocoder, from, to, options, callback) {
     getDocs(options);
 
     function getDocs(options) {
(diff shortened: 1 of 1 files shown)


ingalls:
@mtmail we just deprecated the old `tilelive` indexer which this patch relies on. Do you mind rebasing with current master and I'll gladly merge this!

mtmail:
You're right. Typo in my git upstream URLs. New pull request coming in 1 min.

ingalls:
:+1: Awesome!

PR #436, opened 2016-02-23
[ready] Dialect coach

apendleton:
Add language-guessing behavior to carmen for cases where a user requests a language and we don't have exactly the language requested. New logic:
- Return exact match if we have it
- If not, look for matches that are of the same language, but different subtags:
  - first prefer matches in the same script
  - otherwise go with ones of a different script
- If not, give up and return default

This logic is applied to both forward and reverse geocoding, and to both selection of strings and selection of formats.

Possible variations we could consider:
- Less strict: return script matches even if the languages differ, e.g.: if a user requests Ukrainian and we have Russian (both Cyrillic) return Russian rather than English
- More strict: only return "simpler" languages than requested if there's no exact match (e.g., return "pt" if "pt-BR" is requested, but don't return "pt-PT" if "pt-BR" is requested and we don't have "pt")

I think overall, though, this is a good balance.


apendleton:
Okay, bringing this back from the dead. The strategy actually evolved somewhat, such that now for non-Latin scripts, it actually does allow matches only on script and not on language, and now has a favorite language for each script if the language match is not exact (so if it wants Bulgarian and has only Ukrainian and Russian [all Cyrillic] it prefers Russian). This is all accomplished by calculating a score for how well each available language matches the requested language, sorting, thresholding, and picking the best one left.

This is code is not very well-tested, and is also pretty complicated, so if we end up going this route it will probably need more eyes.

Also: for the Chinese case specifically, model-un doesn't know enough about scripts to make good decisions, so we'll need to hack in some extra stuff for `zh`.

So, to-do as I see it:
* [ ] make sure the strategies employed here match our current expectations for behavior
* [ ] test it
* [ ] make `zh` better (and maybe other macrolanguages without scripts, if any)

cc @sbma44

apendleton:
@sbma44 looks like a busted merge ^

sbma44:
y, I resolved this yesterday but ran into test problems before I pushed. Working through that now.

apendleton:
In general this looks great to me. A couple thoughts (mostly about Chinese, because that's what's in my head at the moment):

I know this is going back on what I said re: case in the model-un PR (there, I had wanted it to be consistent with the rest of the library, which is spec-confirming with respect to case), but I wonder if matches in the fallback code should actually be case-insensitive? Returned text should probably be case-specific (because it's case-specific in the indexes), but maybe keys shouldn't be? So like, `{zh_hans: {'zh', 'zh_Hant'}` or whatever, and then looked up with `.toLowerCase()`? I dunno, just spitballing, and maybe that's not worth worrying about.

More Chinese things... it would be good to have `zh-CN` (mapped to `zh-Hans`) in addition to `zh-TW`. `model-un` doesn't know about per-country scripts for Chinese, so the auto-guessing will fail us there and we'll end up returning plain `zh` even if `zh-Hans` is present. Longer-term I'd like to flesh it out on the model-un side, which will also benefit `zh-HK`, etc., but we should cover our bases on the carmen side at least for CN and TW for now.

Also, gut sense: I might add plain `zh` to some of the fallback lists, maybe? So like: `zh-TW -> zh-Hant,zh,zh-Hans`, etc., just because plain `zh` will at least sometimes be traditional, whereas `zh-Hans` never will be (ditto for fallback from `zh-Hans`, `zh-Hant`, and `zh-CN`: may as well try plain `zh` before trying the "wrong" script). But I haven't actually tried to characterize how often that gives a better result, so maybe not?

sbma44:
:+1: good idea. Added more fallback info for `zh*` and made the fallback matches case-insensitive (while preserving the formal codes in the JSON for maintainability).

apendleton:
I mentioned earlier in chat that this branch should take care of `zh-Hant` vs `zh_Hant`, but I circled back to confirm and it turns out it didn't. The language-guessing part of the function was good with both case and punctuation, but the exact-match (the very first step) worked only for "_", and the fallback code (the second step) worked only for "-", and the exact-match part also broke on case mismatch. That should all be fixed now, but the case-mangling stuff has gotten extra gross at this point. @sbma44 I would definitely be on board now for just forcing everything to be lowercase (and maybe reformatting again at the last second for appearance's sake in the output JSON), but I bandaided it for now.

PR #437, opened 2016-02-23
make sure timer for getIndexableDocs gets started

mtmail:
Inside `indexDocs()` there's a `console.timeEnd` which fails if this timer was never started.


ingalls:
:+1: thanks for the fix!

PR #442, opened 2016-02-25
during index update, make dictcache writable first

mtmail:
`update()` receives a CarmenSource instance.

That `source` might already contain geocoder data and thus during index opening/loading `source._dictcache` is set to `ReadCache` instance. See https://github.com/mapbox/carmen/blob/master/lib/util/dawg.js#L43

To be able to update the `._dictcache` we'd need a `WriteCache` though.

I opted to clear the `._dictcache` since the update operation seems to be a full overwrite without rollback on error. I couldn't find a good place to add test to an existing test file.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -5,6 +5,7 @@ var termops = require('./util/termops'),
     uniq = require('./util/uniq'),
     ops = require('./util/ops'),
     queue = require('queue-async'),
+    dawgcache = require('./util/dawg'),
     indexdocs = require('./indexer/indexdocs'),
     split = require('split'),
     TIMER = process.env.TIMER
@@ -171,6 +172,10 @@ function update(source, docs, options, callback) {
                 return source._commit ? source._commit(callback) : callback();
             });
         }, features);
+
+        //create a new empty writeable cache
+        source._dictcache = new dawgcache;
+
         setParts(patch.grid, 'grid');
         setText(patch.text);
         q.awaitAll(callback);


apendleton:
Hm... I don't think this does what you intend. Calling the "dictcache" a cache is sort of a misnomer; it's a lightweight in-memory set membership data structure, but it's not ephemeral the way, say, memcached might be. It compactly stores all of the strings in a given carmen index so carmen can, at forward geocode time, check to see if a string is a member of a given index before doing the more-expensive disk IO to actually read the full index record (geometry, etc.). If strings are not in the dictcache, the full record will never be retrieved for a given string. What your operation here would do, when trying to add data to an already-built index, would be to nuke the existing contents of the dictcache. It would add new ones, but then those new strings would be the only things that could be looked up in the updated index.

The fundamental problem is that the current data structure we're using to back the dictcache isn't mutable; building it is a one-time deal. That's the reason for the ReadCache/WriteCache split; you can't do both on the same structure. That works fine for us because we tend to rebuild indexes when we make changes, rather than adding data in-place.

That said, the thing you're trying to do is possible, but it's considerably more involved than this. Basically what you'd probably want to do is read all of the strings out of the current structure and write them into a new WriteCache, then proceed with your update, and dump the new WriteCache when you're done. Unfortunately, while the structure we're using (a directed acyclic word graph) is capable of having all of its strings read in theory, we (and be we I mean I; sorry!) haven't written the necessary API to do it, so some C++ spelunking would likely be necessary to make it happen. If you're interested in that, let me know and I can send you some notes as to how the DAWG implementation works.

I'm going to close this for now, but feel free to reopen if you end up wanting to dig in.

PR #444, opened 2016-03-03
Tests for fallback

arunasank:
cc/ @yhahn


yhahn:
@arunasank ok, time to adjust failing tests. I'm seeing tests fall into roughly two categories:

1. :smile: Failures because relevance values are different,
2. :fearful: Failures where the actual expected result is different

For no.1 I think you should be fine just updating the test assertions.

For no.2 take a good :eyes: and first get a sense of what's being tested for -- then let's talk about whether the failure is ok or means this adjustment needs more work.

yhahn:
Continued in https://github.com/mapbox/carmen/pull/448

PR #447, opened 2016-03-08
first cut index merge test

aarthykc:
Setting up tests for multipart indexing

cc @yhahn

### DIFF
--- a/test/index.merge.test.js
+++ b/test/index.merge.test.js
@@ -7,7 +7,6 @@ var Carmen = require('..');
 var index = require('../lib/index');
 var MBTiles = require('mbtiles');
 var mem = require('../lib/api-mem');
-var byline = require('byline');
 
 var UPDATE = process.env.UPDATE;
 var test = require('tape');
--- a/test/index.merge.test.js
+++ b/test/index.merge.test.js
@@ -14,9 +14,7 @@ var termops = require('../lib/util/termops');
 var token = require('../lib/util/token');
 var merge = require('../lib/merge');
 
-
 test('index - streaming interface', function(assert) {
-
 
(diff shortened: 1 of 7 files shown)


test/index.merge.test.js
```diff
@@ -0,0 +1,130 @@
...
+        });
+    });
+    
+    var memObjectC = new mem([], null, function() {});
+    var confC = {};
+    var carmenC = new Carmen(confC);
+
+    merge(carmenC, memObjectA, memObjectB, memObjectC, {}, function(err, done) {
+        if (err) throw err;
+    });
```
yhahn: Because this `merge` call is asynchronous you'll want to run it in an `assert.test('merge indexes', function(q) { ... });` block as well.

Tape runs only one test at a time so it'll run the merge first before moving on to the tests below.

test/index.merge.test.js
```diff
@@ -0,0 +1,130 @@
...
+            q.end();
+        });
+    });
+    assert.test('ensure index was successful for index B after merging', function(q) {
+        carmenB.geocode("Paraguay", {}, function(err, result) {
+            assert.ifError(err, "error");
+            assert.equal(result.features[0].text, "Paraguay", "found Paraguay");
+            q.end();
+        });
+    });
```
yhahn: Hrm, these queries should be run against `carmenC` since that's the merged index right?

aarthykc:
@apendleton I see that you've pulled these changes and it's in master now. Good to close?

apendleton:
@aarthykc yep, I'll close. I ended up having to cherry pick them rather than merge the PR, but they were super helpful; thanks!

PR #450, opened 2016-03-14
restore status output to stderr

sbma44:
minor functionality change to restore better logging during indexes (both to the console and via surrounding tools).

cc @ingalls


sbma44:
successfully built a test index with this. once tests are green I'll merge and bump the patch version.

PR #452, opened 2016-03-28
Speed up getPhraseDegens 30%

tmcw:
micro-optimization hacking for old times sake.

gist of the strategy:

* avoid creating a new string with .substr until we know we want it
* try to avoid indexOf in a hot loop because it's o(n) and we're looping
  through the chars ourselves so we should know what's up

This is reliably 20-30% faster on my machine with the provided
benchmark.


sbma44:
@yhahn @apendleton any reason not to merge this?

apendleton:
@sbma44 looks good to me

PR #453, opened 2016-03-28
Reduce address cluster memory usage

tmcw:
This method was keeping an array in memory but only using a minimum
value. Minimum can be calculated online, so let's do that and avoid
requiring any more memory.


ingalls:
This looks awesome! :+1:

sbma44:
any reason not to merge @ingalls?

ingalls:
Yeah I'm good on this crossing the finish line as long as @tmcw is finished with the branch

tmcw:
Yep, I'm done w/ this branch

PR #455, opened 2016-03-29
Fix bad country code error message. Fixes #454

### LINKED ISSUE #454: carmen calls countries stacks, causing a mismatch between input and output
Input URL:

```
https://api.mapbox.com/geocoding/v5/mapbox.places/vancouver.json?country=US&access_token=...
```

Output error:

``` json
{"message":"Stack \"US\" is not a known stack. Must be one of: cn, au, nz, sg, br, ca, at, de, dk, es, fr, nl, gb, ie, us, ao, bw, dz, eg, ke, ma, za, ae, bh, bn, id, in, jo, kw, my, ph, th, tw, vn, ad, al, ba, be, bg, by, ch, cy, cz, ee, fi, gr, hr, hu, is, it, lb, lt, lu, lv, md, me, mk, mt, no, pl, pt, ro, rs, ru, se, si, sk, sm, tr, ua, jp, ar, cl, co, cu, gf, gp, mx, pe, uy, ve"}
```

What is a stack? Carmen knows, but this isn't mentioned anywhere in our documentation and this would be a very confusing error message for a user who is just tinkering around with the API for the first time. Instead, we should say

``` json
{"message":"Invalid country code provided. Currently supported country codes: cn, au, nz, sg, br, ca, at, de, dk, es, fr, nl, gb, ie, us, ao, bw, dz, eg, ke, ma, za, ae, bh, bn, id, in, jo, kw, my, ph, th, tw, vn, ad, al, ba, be, bg, by, ch, cy, cz, ee, fi, gr, hr, hu, is, it, lb, lt, lu, lv, md, me, mk, mt, no, pl, pt, ro, rs, ru, se, si, sk, sm, tr, ua, jp, ar, cl, co, cu, gf, gp, mx, pe, uy, ve"}
```

The revised error also intentionally removes the string interpolation to avoid the reflection that could be used in an XSS attack with a bad client.

tmcw:
I implemented the more conservative route of the carmen error message and then realized... if we use  EINVALIDSTACK and provide a totally different error message, how will we list valid country values?
- If we use .replace, then we'll make ~~ dependent on the specific error message text, which is a no-no

---

Maybe there's another solution here?

cc @ingalls @apendleton


tmcw:
Closing, stale.

PR #456, opened 2016-04-08
Bench fix

mattficke:
Several required parameters in the geocode and phrasematch benchmark tests were incorrect or missing and raised errors. This changes "Type", "zoom", and "autocomplete".


apendleton:
Good catch. You probably want autocomplete to be true in the phrasematch call, as that's the default behavior.

mattficke:
Got it. Updated now.

mattficke:
stale

PR #457, opened 2016-04-12
Add BBox Filter

mattficke:


lib/geocode.js
```diff
@@ -62,6 +63,14 @@ module.exports = function(geocoder, query, options, callback) {
         if (!mu.hasLanguage(options.language)) return callback(errcode('\'' + options.language + '\' is not a valid language code', 'EINVALID'));
     }
 
+    // bbox option
+    if (options.bbox) {
+        // check if valid bbox
+        if(!options.bbox instanceof Array || options.bbox.length < 4) {
```
sbma44: might as well make this a strict check for length===4

lib/verifymatch.js
```diff
@@ -103,6 +104,12 @@ function dropFeature(geocoder, options, results, callback) {
             }
         }
 
+        function bboxMatch() {
+            if (options.bbox) {
```
sbma44: what's going on here?

sbma44:
this looks good -- left two line notes. @yhahn indicated to me that there might be more to do on carmen-cache from what you two undertook yesterday, but I'm afraid I might not be the best person to explicate that further. let's bug him when he has a moment.

lib/geocode.js
ingalls: can we also add `Must be an array of format [...]` to mirror the other error codes above which provide a hint to what the format must be?

lib/verifymatch.js
mattficke: Whoops, leftover I forgot to delete. Thanks.

lib/phrasematch.js
```diff
@@ -1,17 +1,24 @@
...
 var queue = require('queue-async');
+var bb = require('./util/bbox');
 
 // # phrasematch
 //
 // @param {Object} source a Geocoder datasource
 // @param {Array} query a list of terms composing the query to Carmen
 // @param {Function} callback called with `(err, features, result, stats)`
-module.exports = function phrasematch(source, query, autocomplete, callback) {
+module.exports = function phrasematch(source, query, autocomplete, bbox, callback) {
```
ingalls: Since we're breaking the function parameters here what would you think about a generic `options` map instead of another argument. This will help in the future if we need to pass more data to phrasematch

lib/geocode.js
```diff
@@ -67,7 +67,7 @@ module.exports = function(geocoder, query, options, callback) {
     if (options.bbox) {
         // check if valid bbox
         if(!options.bbox instanceof Array || options.bbox.length === 4) {
-            return callback(errcode('BBox is not valid', 'EINVALID'))
+            return callback(errcode('Must be an array of format [minX, minY, maxX, maxY]', 'EINVALID'))
```
ingalls: Hey sorry for the confusion here, you'll want this in addition to the other mesage or else the user will be left without context.

ie: `BBOX is not valid - Must be an ...`

lib/util/bbox.js
```diff
@@ -0,0 +1,52 @@
...
+* inside - Return whether a coordinate is inside a bounding box.
+* @param {Array} coordinates A lon/lat array
+* @param {Array} bbox A bounding box array in the format [minX, minY, maxX, maxY]
+* @return {boolean} Is the point inside the bbox
+*/
+function inside(coordinates, bbox) {
+    var point = Point(coordinates);
+    if (!bbox) return true;
+    var poly = BBox(bbox);
+    return Inside(point, poly);
```
danpat: If this is always going to be a box, then `turf-inside` is probably overkill, as does a full point-in-polygon test.  Could this be replaced with with a simple bbox (a-la http://gamedev.stackexchange.com/a/913) ?

sbma44: good catch. the code as it exists outside of carmen-cache is very close to supporting arbitrary polys -- but carmen-cache is a fairly big caveat. agree that simplifying this & ditching the dependency is probably worthwhile.

mattficke:
The validation check currently rejects bboxes that cross the dateline, in which minX > maxX (i.e., `[170, 40, -175, 50]`. This is a less-than-ideal punt, but it at least gives a clear error message instead of returning incorrect results.

A fix for this issue should probably be tackled on its own, possible options include:
* splitting the bbox into two separate bboxes along the dateline
* transforming longitude to a 0-360 scale

Starting to see what @tmcw meant by "[The dateline](http://www.macwright.org/2015/03/23/geojson-second-bite.html#the-dateline) is one of the shames of geospatial technology".

PR #458, opened 2016-04-15
Add infrastructure for merging indexes

apendleton:
Okay, here finally is the index merging PR. I could use particular :eyes: on the functions responsible for merging particular pairs of index shards for each merge type, here: https://github.com/mapbox/carmen/blob/merge-redux/lib/merge.js#L111-L235

There's one each for freq, grid, features, and dawg.

Note also that it was necessary to avoid blowing the callstack to update from queue-async to d3-queue, and then fix a few spots unrelated to merge where subtle d3-queue behavior changes resulted in breakage. Sorry about the PR noise.

cc @mapbox/geocoding-gang


yhahn:
:palm_tree: :ship:

PR #459, opened 2016-04-28
Disable autocomplete indexing for translations

yhahn:
Disables generation of autocomplete degens in the grid cache at indexing time for translated text for now.

### Next actions

- [x] Test IRL
- [x] Merge + release


yhahn:
This is pretty breakingish so if we do roll a release we should major bump :grimacing:

PR #466, opened 2016-05-23
remove check for carmen:center at load time

mattficke:
Carmen will fail if you attempt to index features that do not have a center set.

From chat with @ingalls:

>ingalls [4:13 PM] 
@sbma44: @mattficke Carmen won’t actually generate centres
it will only correct them if there are invalid centres
so the centre column has to be populated with a valid lon/lat coord
or else it will hard fail
We should consider changing this as carmen’s centres are perfectly fine

This check seems to be redundant and causes unnecessary failures since we also check for and fix the center with this code:
```js
if (!doc.properties["carmen:center"] || !verifyCenter(doc.properties["carmen:center"], tiles)) {
        console.warn('carmen:center did not fall within the provided geometry for %s (%s). Calculating new point on surface.',
            doc.id, doc.properties["carmen:text"]);
        doc.properties["carmen:center"] = centroid(doc.geometry).geometry.coordinates;
        if (!verifyCenter(doc.properties["carmen:center"], tiles)) {
            return 'Invalid carmen:center provided, and unable to calculate corrected centroid. Verify validity of doc.geometry for doc id:' + doc.id;
        } else {
            console.warn('new: carmen:center: ', doc.properties["carmen:center"]);
            console.warn('new: zxy:    ', doc.properties['carmen:zxy']);
        }
    }
```
This removes the first check for `carmen:center` that runs before carmen attempts to fix the issue.

### DIFF
--- a/lib/indexer/indexdocs.js
+++ b/lib/indexer/indexdocs.js
@@ -100,8 +100,6 @@ function runChecks(doc, zoom) {
         return 'doc has no properties on id:' + doc.id;
     } else if (!doc.properties["carmen:text"]) {
         return 'doc has no carmen:text on id:' + doc.id;
-    } else if (!doc.properties["carmen:center"]) {
-        return 'doc has no carmen:center on id:' + doc.id;
     } else if (doc.properties["carmen:geocoder_stack"] &&
         typeof doc.properties["carmen:geocoder_stack"] !== 'string') {
         return 'geocoder_stack must be a string value';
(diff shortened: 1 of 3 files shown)


ingalls:
Yeah this looks good to me, didn't realize you were going after this one! Let's get a thumbs up from @sbma44 and @yhahn to see if they can think of any edge cases where this could be problematic.

PR #468, opened 2016-05-25
Address order

mattficke:
Token replacement tests for Japan


mattficke:
@yhahn still poking at things, but here's a test for one of the behaviors I'm seeing: https://github.com/mapbox/carmen/blob/japan/test/geocode-unit.jp.test.js#L260-L270

If the query is ordered small-to-big, it works, but not if it's big-to-small.

mattficke:
Currently trying to isolate the relevance scoring behavior, it looks like a lot of addresses get lost in a sea of similarly low-scored features. Here's a test for a similar problem (the relevance hit that happens when an apartment number is included) using a western-style address: https://github.com/mapbox/carmen/blob/f916ad956337c78ee7089e48228d46131a1af338/test/geocode-unit.relevance.test.js#L90-L100

mattficke:
@mapbox/geocoding-gang ok, help! After adding a bunch of tests I'm coming to the conclusion that Carmen doesn't like addresses that are ordered from big-to-small.

This is a problem for Japan, where we're trying to order the addresses like this:
```
"{country._name}, {region._name}{district._name}{place._name}{address._name}{address._number}"
```

As best I can tell, the correct features are found in `phraseMatch`, but get lost in `spatialMatch` (specifically [here] (https://github.com/mapbox/carmen/blob/master/lib/spatialmatch.js#L162)). However, many of the operations there are done using bitmasks, and I'm getting a little lost trying to isolate exactly where Carmen decides these features don't stack.

Any thoughts?

mattficke:
I noticed some weird caching behavior on Circle, features added and (supposedly) removed in a test will still be present in the next test.

First, in geocode-unit.jp.test.js, queries for the addresses `4433 勝田度会郡` and `3591 勝田度会郡` fails the test because it returns back `岩出市` as the `place`, rather than `度会郡`, from the query, which lowers the relevance below the expected value.  This is weird because that incorrect place isn't present in the test at all; it is part of geocode-unit.jp-order.test.js, two files before this.

image
image

The other error is in geocode-unit.language-flag-reverse.test.js, where `Japan` is being returned back as the country instead of `China`, even though Japan isn't added in the test. It only appears in the test immediately prior.

image


I added a zero-second timeout to the teardown function just in the Japan tests:
```js
tape('teardown', function(assert) {
    context.getTile.cache.reset();
    setTimeout(function() {
        assert.end();
    }, 0);
});
```
Which prevented the cross-pollination within the other Japan tests, but not the language-flag-reverse test.

I then added a timeout to the language-flag-reverse test, but this still did not prevent `Japan` from showing up in results. Changing the timeout to 5 seconds didn't change anything.

This seems to be happening because the test features all have the same z/x/y coordinates (6/32/32) and are lingering somehow between tests.

I don't know why the timeout would work for one part of the problem but not the other, however.

* First test run where it showed up here: https://circleci.com/gh/mapbox/carmen/965
* setTimeout fixed part of the problem here: https://circleci.com/gh/mapbox/carmen/970

Search for `not ok` to find the failing tests.

cc @yhahn @sbma44

test/geocode-unit.jp-numeric.test.js
```diff
@@ -0,0 +1,109 @@
...
+        t.equal(res.features.length, 1, "1 feature");
+        t.equal(res.features[0].address, '5', "right address");
+        
+    });
+    c.geocode('神明台三丁目 5', null, function(err, res) {
+        t.ifError(err);
+        t.equal(res.features.length, 1, "1 feature");
+        t.equal(res.features[0].address, '5', "right address");
+        t.end()
+    });
```
yhahn: @mattficke I'm not sure if this is the cause of your teardown woes but making multiple calls to `c.geocode()` within a single tape test is a little suspect.

The `t.end()` here may be called when the query above finishes which will trigger the teardown to be called next. But this _may_ be called before the first 3 queries are done running and teardown may attempt to happen while they're still in progress.

Easiest fix for this is to move each `c.geocode()` call into its own `tape()` test since tape runs only one test at a time. Alternatively you could set up a queue of geocode calls to run and then only call `t.end()` when all the queries have completed.

mattficke: @yhahn thanks for catching, looks like that was it. Updated now.

mattficke:
🍏 🎉 

Addresses can now be ordered big-to-small or small-to-big.

I removed the order check from spatialmatch entirely, and am relying on the relevance penalty in verifymatch to take care of mismatched ordering. This does increase the number of stackable combinations that are passed on to verifymatch; it's unclear at the moment if this will impact performance or accuracy.

lib/verifymatch.js
```diff
@@ -250,17 +250,28 @@ function verifyContext(context, strict, loose, groups) {
...
+        if (c === 1) {
+            direction = lastmask > matched.mask ? "descending" : "ascending"
+        }
+
+        // penalize stacking bonus if the order of address components is inconsistent
+        if (direction === "descending") {
+            backy = lastmask < matched.mask;
+        } else if (direction === "ascending") {
+            backy = lastmask > matched.mask;
+        }
```
yhahn: Should this check be within the `lastgroup > -1` block?

lib/spatialmatch.js
```diff
@@ -183,7 +183,7 @@ function stackable(phrasematches, idx, mask, nmask, stack, relev) {
 
         if (targetMask & next.mask) continue;
         if (targetNmask & next.nmask) continue;
-        if (targetMask && targetMask < next.mask) continue;
+        // if (targetMask && targetMask < next.mask) continue;
```
yhahn: Per your notes if this check was helping reduce the inputs to `verifymatch()` then it's very likely both perf + accuracy will suffer at scale without this.

Do you think it might be possible to set a similar directionality flag when the first two masks are compared and then use that directionality for all subsequent checks here?

mattficke: @yhahn yes, that'll probably be a good idea. Should be able to use the same directionality check here.

lib/verifymatch.js
mattficke: @yhahn technically doesn't need to be since direction isn't set until the second time through the loop, but you're right that it would make it easier to understand.

yhahn: Ah, makes sense.

test/spatialmatch.stackable.test.js
```diff
@@ -68,17 +68,17 @@ test('stackable complex', function(assert) {
...
-        // '0.66 - a1, b2', => mask goes backwards
-        // '0.66 - b1, c2', => mask goes backwards
+        '0.66 - a1, b1',
         '0.66 - b2, c1',
+        '0.66 - a1, b2',
+        '0.66 - a1, c2',
         '0.66 - a1, c1',
-        '0.66 - a1, b1',
+        '0.66 - b1, c2'
     ]);
```
yhahn: :+1: now that both mask flow directions are supported.

It'd be great if you could annotate an example stacking permutation that gets excluded by the backwards/forwards flow-direction check so it's clear we're not letting everything through.

test/geocode-unit.score.test.js
```diff
@@ -151,7 +151,7 @@ var addFeature = require('../lib/util/addfeature');
         addFeature(conf.city, city, t.end);
     });
     tape('china', function(t) {
-        c.geocode('china', { limit_verify:1 }, function(err, res) {
+        c.geocode('china', { allow_dupes: true}, function(err, res) {
```
yhahn: What was this for?

lib/util/termops.js
```diff
@@ -112,9 +112,10 @@ function tokenize(query, lonlat) {
         .split(/[\s+]+/gi);
 
     var pretokens = [];
-
     for (var i=0;i<normalized.length;i++) {
-        if (/(\d+)-(\d+)[a-z]?/.test(normalized[i])) {
+        if (normalized[i].length && normalized[i][0].charCodeAt() >= 19968 && normalized[i][0].charCodeAt() <= 40959) {
+            pretokens = pretokens.concat(normalized[i].split(/([0-9０-９]+)/));
+        } else if (/(\d+)-(\d+)[a-z]?/.test(normalized[i])) {
```
yhahn: Hm, what is this additional logic for? Could you add a unit test to `termops.tokenize` that demonstrates this functionality?

test/geocode-unit.score.test.js
mattficke: whoops, left over from some earlier debugging, meant to change back. thanks.

lib/util/termops.js
mattficke: @yhahn this is to keep multi-digit numerals in Japanese addresses from being split into individual digits during tokenization. I'll add a test.

mattficke:
I think this is good to go! Looks ok to me in [api-geocoder](https://github.com/mapbox/api-geocoder/pull/1066). There's a little jiggle in the results, which is expected, but it's mostly positive changes. When we're ready to deploy, I'm going to roll #491 in as well. Already merged in #497.

@yhahn, can I get your 👀 on this when you get a chance?

lib/spatialmatch.js
```diff
@@ -175,11 +182,20 @@ function stackable(phrasematches, idx, mask, nmask, stack, relev) {
...
-        if (targetMask && targetMask < next.mask) continue;
+
+        // sort by order of input query
+        targetStack.sort(sortByMask);
+        // compare index order to input order to determine direction
+        if (targetStack.length) {
+            direction = targetStack[0].idx < next.idx ? "descending" : "ascending";
+        }
+
+        if (direction === "ascending" && targetMask && targetMask < next.mask) continue;
```
yhahn: Small possible perf optimization: could this direction-setting happen once outside the for loop over the phrasematches? It seems like this direction will always be the same for every element in the loop since `targetStack` is a copy of `stack`.

e.g. could stack be copied, sorted, and direction set here? https://github.com/mapbox/carmen/blob/4c9efc2e1b2167b9d12a61e7a3c4025ed385cb34/lib/spatialmatch.js#L178

yhahn:
Heh new github review system is weird : ) @mattficke added some notes above -- I think once you feel good about these things :ship:

PR #469, opened 2016-05-26
Address cluster CAPS

ingalls:
Currently address clusters are retained as they are submitted to carmen. So if `24A` is submitted in a cluster it remains `24A`, `24a` remains `24a` etc. The input however is always lowercase. Since the way these are matched up is simply a javascript array, if there is a case different, a match doesn't occur.

`["24a", "24A"]`

This PR ensures that both the query address and the address cluster array are always lowercased.

cc/ @yhahn @sbma44


sbma44:
makes sense to me

PR #471, opened 2016-06-14
Added reverse permutation

aarthykc:
Added reverse permutation, i.e: 
`st => saint`

cc @yhahn @arunasank


yhahn:
:+1: looks good, a couple notes:

- If the code is largely the same as `createReplacer(tokens)` I wonder if rather than a separate function as we originally spec'd it could just be a second true/false argument, like `createReplacer(tokens, reverse)` ?
- Do you want to add a testcase for using the reversed replacer generated with the `replaceToken()` function?

lib/util/token.js
```diff
@@ -1,13 +1,20 @@
...
 
-module.exports.createReplacer = function(tokens) {
+var createReplacer =  module.exports.createReplacer = function(tokens, reverse) {
     var replacers = [];
 
     for (var token in tokens) {
-        var from = token; // normalize expanded
-        var to = tokens[token];
 
+        if (reverse == false) {
```
yhahn: Do you want to use the looser check:

```
if (!reverse)
```

This will let existing callers of `createReplacer()` not have to add an additional arg (you can go back in the test to leaving out the `false` argument) since this check also passes for `null` and `undefined`.

aarthykc:
All commits here are in `carmen-token-replacement` branch, closing here.

PR #474, opened 2016-06-15
Phrasematch wire

aarthykc:
integrating replace token work

cc @arunasank @yhahn

### DIFF
--- a/lib/phrasematch.js
+++ b/lib/phrasematch.js
@@ -12,7 +12,14 @@ module.exports = function phrasematch(source, query, options, callback) {
     options = options || {};
     options.autocomplete = options.autocomplete || false;
     options.bbox = options.bbox || false;
-    var tokenized = termops.tokenize(token.replaceToken(source.token_replacer, query));
+    //console.log('query', query);
+    //console.log('source token replacer', source);
+
+    var tokenized = termops.replaceTokenPermutations(source.token_replacer, query);
+
+    //var tokenized = termops.tokenize(token.replaceToken(source.token_replacer, query));
+
+    //console.log('tokenized', tokenized);
     var getter = source.getGeocoderData.bind(source);
     var loadall = source._geocoder.loadall.bind(source._geocoder);
 
@@ -34,6 +41,7 @@ module.exports = function phrasematch(source, query, options, callback) {
     }
 
     subqueries = termops.uniqPermutations(subqueries);
+    console.log('subqueries', subqueries);
 
     for (var l = 0; l < subqueries.length; l++) {
         var phrase = termops.encodePhrase(subqueries[l], options.autocomplete ? subqueries[l].ender : false);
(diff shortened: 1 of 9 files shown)


lib/util/termops.js
```diff
@@ -232,50 +232,54 @@ function getIndexableText(replacer, doc) {
...
-
-    var keys = Object.keys(doc.properties);
-    var length = keys.length;
-    for (var i = 0; i < length; i ++) {
-        var code = keys[i].match(/text_(.+)/);
-        if (code) {
-            if (!mun.hasLanguage(code[1])) throw new Error(code[1] + ' is an invalid language code');
-            if (doc.properties[keys[i]]) getTokens(doc.properties[keys[i]].split(','), false);
-        }
-    }
```
yhahn: @aarthykc you'll want to keep this portion of the indexing code as it's gathering `text_{language}` fields.

lib/phrasematch.js
```diff
@@ -12,7 +12,14 @@ module.exports = function phrasematch(source, query, options, callback) {
     options = options || {};
     options.autocomplete = options.autocomplete || false;
     options.bbox = options.bbox || false;
-    var tokenized = termops.tokenize(token.replaceToken(source.token_replacer, query));
+    //console.log('query', query);
+    //console.log('source token replacer', source);
+
+    var tokenized = termops.replaceTokenPermutations(source.token_replacer, query);
```
yhahn: You may want to think about what `replaceTokenPermutations()` is doing as providing n different versions of the query that you now need to search with.

e.g. think of this step as

```js
var queries = termops.replaceTokenPermutations(source.token_replacer, query);
// Loop through each permutation as if it's a totally separate query
queries.forEach(function(query) {
    // everything that used to happen before... except without replaceToken
    var tokenized = termops.tokenize(query);

    ...
    subqueries = termops.uniqPermutations(subqueries);
});

// eventually these all need to come back together + merge all the individual subqueries arrays into one
```

Here's a crummy ascii diagram of how i'm imagining this flow:

```
                                               / queries[0] => termops.tokenize => termops.permutations => subqueries \
                                              /                                                                        \
query -- replaceTokenPermutations => queries --- queries[1] => termops.tokenize => termops.permutations => subqueries --- merge subqueries into a single array
                                              \                                                                        /
                                               \ queries[2] => termops.tokenize => termops.permutations => subqueries /
```

lib/phrasematch.js
```diff
@@ -21,20 +22,21 @@ module.exports = function phrasematch(source, query, options, callback) {
         var intersects = bb.intersect(options.bbox, source.bounds);
         if (!intersects) return callback(null, []);
     }
+    
+    var queries = termops.replaceTokenPermutations(source.token_replacer, query);
+    queries.forEach(function(query) {
+        tokenized = termops.tokenize(query);
+        // Get all subquery permutations from the query
+        subqueries = termops.permutations(tokenized);
```
yhahn: Each time the loop runs this part is actually replacing any previous value of `subqueries`.

Check out the way `.concat()` can be used to add onto the existing value of `subqueries` on each iteration of the loop as a way to build up the list rather than replace:

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/concat

aarthykc:
All commits here are in `carmen-token-replacement` branch, closing here.

PR #476, opened 2016-06-20
Carmen token replacement

arunasank:
Generate permutations for a query after replacing tokens.


arunasank:
Three tests fail:
1. [`geocode-unit.gappy.test.js`](https://github.com/mapbox/carmen/blob/master/test/geocode-unit.gappy.test.js#L102-L117) both those tests fail because spatialmatch doesn't return any features.
2. [`geocode-unit.stacky.test.js`](https://github.com/mapbox/carmen/blob/master/test/geocode-unit.stacky.test.js#L62-L70) - The order of the features change:
```
   [ { id: 'city.1',
       type: 'Feature',
       text: 'windsor',
       place_name: 'windsor, connecticut',
       relevance: 0.6656666666666666,
       properties: {},
       bbox: [Object],
       center: [Object],
       geometry: [Object],
       context: [Object] },
     { id: 'street.1',
       type: 'Feature',
       text: 'windsor ct',
       place_name: 'windsor ct, windsor',
       relevance: 0.49,
       properties: {},
       bbox: [Object],
       center: [Object],
       geometry: [Object],
       context: [Object] } ] }
```
3. [`index.test.js`](https://github.com/mapbox/carmen/blob/master/test/index.test.js#L78-L84)
```
expected:
      { 0: [ 4 ], 1: [ 2 ], 3363289958149993: [ 1 ], 441841902895320: [ 1 ], 609659059851264: [ 2 ] }
    actual:
      { 0: [ 4 ], 1: [ 2 ], 1247264641460936: [ 1 ], 1804046053253033: [ 1 ], 609659059851264: [ 2 ] }
```
^^^ I think this is because of us changing the token replacer to generate both forward and reverse tokens.
The index test fails because of different bench values from what it expects.

arunasank:
Also new bench results after we generate both forward and reverse tokens and check for both:
```
# replace tokens
Permutations all together of 1 token x 54,200 ops/sec ±1.85% (95 runs sampled)
Permutation all together of 3 tokens x 30,298 ops/sec ±1.94% (84 runs sampled)
##### WITHOUT XREGEXP:
 SPLIT(1)create token replacer x 1,660 ops/sec ±2.52% (93 runs sampled)
SPLIT (2): filtering tokens x 20,457 ops/sec ±0.83% (96 runs sampled)
SPLIT (3): Permutation of US index tokens without XRegexp x 14,191 ops/sec ±0.90% (95 runs sampled)
##### WITH XREGEXP
SPLIT(1)create token replacer x 900 ops/sec ±3.27% (83 runs sampled)
SPLIT (2): filtering tokens x 10,173 ops/sec ±1.69% (91 runs sampled)
SPLIT (3): Permutation of US index tokens with a query regex x 5,470 ops/sec ±8.86% (76 runs sampled)
```

yhahn:
### Step 1

Per chat, the issue atm lies in the subquery deduplication code, but it'll be hard to refactor this and test for the issue without unnesting all our arrays a bit.

Ideally you can eliminate the double-nesting of `subqueriesCollection` in phrasematch -- see if calculating the weight in the first loop lets you eliminate the need to track a `tokenizedCollection` and just `concat()` the subqueries into the collection.

### Step 2

Big picture, this deduplication needs to preserve unique subqueries by:

- Their text,
- Their `ender` property,
- Their `mask` property

This means that if you have two subqueries that have the following characteristics:

```
// subquery A
[ `new`, `york`, mask: 3, ender: false ]
// subquery B
[ `new`, `york`, mask: 12, ender: false ]
```

They should both be present after deduplication. If you have two subqueries with the same text and the same ender/mask flags, they need to be deduped.

cc @arunasank @aarthykc

arunasank:
@yhahn @aarthykc Tests pass 👯
* We are [calculating the weight earlier](https://github.com/mapbox/carmen/pull/476/commits/a641853919c5d7cc39193079ff8c89d3be3b38c7)
* I am using  `Array.push()` instead of `Array.concat()` because when I concat, the deduplication function gets an input like:
```
[ [ 'windsor', 'ct', ender: true, mask: 3, text: 'xwindsor ct', phrase: 452684258444624, weight: 1 ],
  [ 'windsor', ender: false, mask: 1, text: 'xwindsor', phrase: 4002333647442019, weight: 0.5 ],
  [ 'ct', ender: true, mask: 2, text: 'xct', phrase: 2028537849417076, weight: 0.5 ] ]

[ [ 'windsor', 'ct', ender: true, mask: 3, text: 'xwindsor ct', phrase: 452684258444624, weight: 1 ],
  [ 'windsor', ender: false, mask: 1, text: 'xwindsor', phrase: 4002333647442019, weight: 0.5 ],
  [ 'ct', ender: true, mask: 2, text: 'xct', phrase: 2028537849417076, weight: 0.5 ] ]
```
instead of
```
[ [ [ 'windsor', 'ct', ender: true, mask: 3, text: 'xwindsor ct', phrase: 452684258444624, weight: 1 ],
  [ 'windsor', ender: false, mask: 1, text: 'xwindsor', phrase: 4002333647442019, weight: 0.5 ],
  [ 'ct', ender: true, mask: 2, text: 'xct', phrase: 2028537849417076, weight: 0.5 ] ],
[ [ 'windsor', 'ct', ender: true, mask: 3, text: 'xwindsor ct', phrase: 452684258444624, weight: 1 ],
  [ 'windsor', ender: false, mask: 1, text: 'xwindsor', phrase: 4002333647442019, weight: 0.5 ],
  [ 'ct', ender: true, mask: 2, text: 'xct', phrase: 2028537849417076, weight: 0.5 ] ] ]
```
and I found parsing an array easier than parsing a non-array. Would love to hear 💭 on this.

* [Changed the frequency values in `index.test.js`](https://github.com/mapbox/carmen/pull/476/commits/c5d86a74c71ef3eee7b394b2d694c2700d5d2c0f) - I am not sure at this point what this does _exactly_ - seems like it returns the number of times a text is found in the docs. [The stats for *frequency with tokens* and *frequency without tokens* becomes the same](https://github.com/mapbox/carmen/blob/carmen-token-replacement/test/index.test.js#L70-L84). Is this because now there is no tokenising in the index, and therefore the [ids generated in generateFrequency](https://github.com/mapbox/carmen/blob/carmen-token-replacement/lib/index.js#L237) are the same?

  Can we remove one of these tests entirely? Also, does this involve changing [generateFrequency](https://github.com/mapbox/carmen/blob/carmen-token-replacement/lib/index.js#L214) so that it doesn't accept tokens?

yhahn:
@arunasank @aarthykc nice work! I'll take a look at this tonight and respond to your notes as well.

aarthykc:
These are the permutations which we get for `中国` -- https://gist.github.com/aarthykc/17ec4243ea6f309bee2cbe90cd797ab3

cc @yhahn, @arunasank

aarthykc:
Special characters such as `中国` when it is sent as a query to `replaceTokenPermutation()` would give back [these permutations](https://gist.github.com/aarthykc/17ec4243ea6f309bee2cbe90cd797ab3), causing tests to fail on our api, because of the number of permutations created. The replacer was replacing every '' with a regex [in our global tokens](https://github.com/mapbox/carmen/pull/476/files#diff-87da5a82ba81f7e3f490746bf72135a3R13). 

The fix for this was to prevent reverse token replacement, for e.g., for the token `Street: st` replacing `St => Street` 
from happening when it's the global tokens. Which meant that when there is '' it wouldn't be replaced by a regex. 

cc @arunasank, @yhahn

yhahn:
@aarthykc @arunasank :ok_hand: 

I think the last hurdle here is seeing if we can restore the `decollide()` functionality or something equivalent so we can ship this work without waiting on the new string-based gridcache.

I'll give this some thought today and write up notes!

PR #477, opened 2016-06-28
[carmen-token-replacement] Update + reenable decollide

yhahn:
Quick adjustment to reenable `decollide()` for the `carmen-token-replacement` branch.

I'm seeing that this isn't enough to address our test changes though so I may :mag: :eyes: for a deep dive...

cc @aarthykc @arunasank


ingalls:
Stale - Ok to close?

PR #478, opened 2016-06-30
Unicode test

aarthykc:
Unicode tests to see how carmen handles these now and  compare it with the query time `carmen-token-replacement` quest.

cc @arunasank, @yhahn


ingalls:
Stale - Ok to close?

PR #479, opened 2016-07-08
[ready]Validate if token replacer is global or per index

aarthykc:
To check if tokens which have a space between them are global or generated per index. 

cc @yhahn, @arunasank

### DIFF
--- a/index.js
+++ b/index.js
@@ -237,8 +237,10 @@ function boundsIntersect(a, b) {
 }
 
 function tokenValidator(token_replacer) {
-    if (token_replacer[0].from.toString().indexOf(' ') >= 0 || token_replacer[0].to.toString().indexOf(' ') >= 0) {
-        return true;
+    for (var i = 0; i < token_replacer.length; i++) {
+        if (token_replacer[i].from.toString().indexOf(' ') >= 0 || token_replacer[i].to.toString().indexOf(' ') >= 0) {
+            return true;
+        }
     }
 }
 
--- a/index.js
+++ b/index.js
@@ -101,6 +101,12 @@ function Geocoder(indexes, options) {
             source.geocoder_layer = (info.geocoder_layer||'').split('.').shift();
             source.geocoder_tokens = info.geocoder_tokens||{};
             source.token_replacer = token.createReplacer(info.geocoder_tokens||{});
+
+            if(tokenValidator(source.token_replacer)) {
+                err = new Error('Using global tokens');
+                return;
+            }
+
             source.maxzoom = info.maxzoom;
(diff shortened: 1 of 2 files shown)


aarthykc:
Added tests to the validator, throws an error if it uses global tokens. 

Would be great if you could take a look at this, @yhahn? 

cc @arunasank

index.js
```diff
@@ -231,6 +236,12 @@ function boundsIntersect(a, b) {
     return true;
 }
 
+function tokenValidator(token_replacer) {
+    if (token_replacer[0].from.toString().indexOf(' ') >= 0 || token_replacer[0].to.toString().indexOf(' ') >= 0) {
+        return true;
+    }
```
yhahn: Hm, this looks like it's checking the very first `token_replacer[0]` from/to pair. Do you want to see if we could check them all?

(You can add a test for this by having an index with multiple geocoder_token pairs and having one of the invalid ones be later in the list).

yhahn:
@aarthykc want to update the name of the test? And then I think this is ready for :ship:!

mattficke:
@aarthykc @yhahn Hm, is there a way to accomplish this by actually checking if there's a cardinality change, rather than just checking if there are spaces? This current version means that all CJK tokens must be global (becuase of unidecode), and makes it more difficult to test.

Example that doesn't work now: `geocoder_tokens: {"[３3]丁目": "三丁目"}`

Thinking something like this instead:
```js
if (token_replacer[i].from.toString().split(' ').length !== token_replacer[i].to.toString().split(' ').length
```

mattficke:
Turns out unidecode makes this somewhat more complicated since CJK queries are split per token. In my above example it turns into this, with lengths differing by 1:
```
from: /(\W|^)[33]Ding Mu (\W|$)/gi,
to: '$1San Ding Mu $2'
```
Need to mull it over more. One possibility is that this error shouldn't throw in Carmen, it should be in either api-geocoder or mapbox-places, since those are where the tokens actually live.

mattficke:
@aarthy To clarify what's happening here, `3丁目` (3-chome, in latin characters) is an address component that can be written with either western or Japanese numerals. `三丁目` is the equivalent with all Japanese characters.

Our data source uses Japanese numerals, so I want to add an index-level token replacer to match queries with western numerals.

CJK queries are split per character, so the query is `3,丁,目`, and then the `from` token is unidecoded during the replacer creation process.

Here's the unidecoded versions of both (note the space after `Mu`, which unidecode typically inserts):

* `3丁目` --> `3Ding Mu `
* `三丁目` --> `San Ding Mu `

PR #480, opened 2016-07-13
Dedup addresses w/address vs w/o [WIP]

ingalls:
screenshot 2016-07-13 18 18 51

This really bothers me. WIP method to fix it.

cc/ @mapbox/geocoding-gang


sbma44:
ugh yeah hate this. we return addressless street features way too eagerly right now. thanks for grabbing.

ingalls:
Fixed on the data side with PT2ITP.

PR #482, opened 2016-07-15
set relevance to 1 when querying by id

### LINKED ISSUE #481: id search
searching by id

ie: `address.1234` should return with a relev of 1.

cc/ @mapbox/geocoding-gang

mattficke:
fixes #481


ingalls:
Looks good, can you make sure to add some tests to ensure this behavior doesn't ever get changed back by accident.

ingalls:
@mattficke This looks good to go!

PR #492, opened 2016-08-23
[WIP] Additional Score Filter

arunasank:
Filter features by a score X that's a CLI parameter. This will also mean that we _may_ expose the scores of features in the result json.

**CLI parameter**
- `--scoreAbove` (can range between 1 -100)

**How this works**
- The value of `scoreAbove` is normalised for each index that carmen decides to search in:
  `Normalised value of scoreAbove in an index = scoreAbove * scorefactor`

if **Normalised value of scoreAbove in an index > feature score**, filter feature;
else, drop;

cc/ @mapbox/geocoding-gang

### DIFF
--- a/lib/geocode.js
+++ b/lib/geocode.js
@@ -33,8 +33,9 @@ module.exports = function(geocoder, query, options, callback) {
     }
 
     //scoreAbove option
-    if (options.scoreAbove && Number.isNaN(options.scoreAbove)) {
-        return callback(errcode('--scoreAbove can only be a number'));
+    if (options.scoreAbove) {
+        if (!Number.isNaN(options.scoreAbove) || (options.scoreAbove < 1 || options.scoreAbove > 100))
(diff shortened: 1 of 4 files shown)


mattficke:
>This will also mean that we may expose the scores of features in the result json

Let's make sure this is ok, my previous understanding was that we can't display this publicly.

sbma44:
I think we're fine to display scores from a legal perspective, the question is whether we want to. We already get a bunch of annoying questions from users about `relevance`. I am not looking forward to being asked "why does feature X have a higher score than feature Y, I think feature X is much more important". It will also make it harder to radically adjust our score system without damaging users' implementations.

OTOH scores are potentially useful and illuminating, so please don't take the above as me expressing a strong preference against. Worth considering the maintenance burden, however.

ingalls:
Stale - Ok to close?

PR #493, opened 2016-08-29
Unscaled scoring

arunasank:
Filter features by a score X that's a CLI parameter.

`--scoreAbove` - takes in an arbitrary number. Returns features that have a `carmen:score` higher than the the CLI parameter.

Different from https://github.com/mapbox/carmen/pull/492 (which is incomplete) in the sense that this is a _direct_ comparison between the CLI score and the `carmen:score` with the `scorefactor` not influencing feature selection in any way.
- Use `carmen:score` from the indexes to qualify a feature, instead of `score`.
- Wrote a new function to conditionally qualify features depending on options passed.


ingalls:
Stale - Ok to close?

PR #494, opened 2016-09-12
str -> strasse

aarthykc:
Right now, 
We get results for `xyz strasse`, `xyz straße`, but, don't for `xyzstr` although it is a an acceptable format to search for a place. This PR has tests written to support being able to search for `xyz strasse` using `xyzstr`

cc @yhahn, @sbma44


test/geocode-unit.tokens.test.js
```diff
@@ -176,6 +176,49 @@ var addFeature = require('../lib/util/addfeature');
...
 })();
 
+(function() {
+    var conf = {
+        address: new mem({
+            maxzoom: 6,
+        }, function() {})
+    };
+    var opts = {
+        tokens: {'(str) ': "strasse"}
```
yhahn: @aarthykc cool test! Do you want to do the above token replacement in the contraction direction `'(strasse) ': 'str'` and then show that it works for both versions of the query?

sbma44:
lmk when you want :eyes: on this -- seeing a couple of things I have questions about but I know it's still a work in progress

ingalls:
@aarthykc brought some unit tests to me that should have been passing with global tokens but unfortunately were failing. In our dance to update the indexer to streaming and with the changes that were made in the tokenizer to prevent changes in cardinality we accidentally broke tokens that don't equate to `""`. 

ie: current tokens that map `"something something" => ""` will work. But the additional tokens in this branch that map `"something something" => "Something"` will not.

This is due to the fact that originally the indexer didn't have any distinction between `geocoder_tokens` (tokens) and the `options.tokens` (global tokens). This broke when we changed termops to automatically add word boundaries to all the tokens. 

To fix this we need to change the indexer so that it applies `option.tokens` first (and separately) from `geocoder_tokens`

lib/util/token.js
```diff
@@ -56,3 +56,20 @@ module.exports.replaceToken = function(tokens, query) {
 
     return abbr;
 }
+
+module.exports.createGlobalReplacer = function(tokens) {
```
sbma44: hmm, how come this is its own function? just curious -- I thought we had the token replacement stuff we would need already in place

lib/indexer/indexdocs.js
```diff
@@ -21,18 +21,19 @@ module.exports.verifyCenter = verifyCenter;
...
 function indexdocs(docs, source, options, callback) {
-
-    if (typeof zoom !== 'number')
+    if (typeof options.zoom !== 'number')
         return callback(new Error('index has no zoom'));
-    if (zoom < 0)
-        return callback(new Error('zoom must be greater than 0 --- zoom was '+zoom));
-    if (zoom > 14)
-        return callback(new Error('zoom must be less than 15 --- zoom was '+zoom));
+    if (options.zoom < 0)
```
sbma44: is this related to the tokenquest or did you run across a zoom bug while working on it?

lib/util/token.js
ingalls: @sbma44 Tl;Dr: We can't add word boundaries to global tokens as they were initially written to be assumed complete in themselves. 

Since writing the global replacement code, there have been two subsequent token refactors that made my original approach of combining them with normal tokens to silently break. The original global tokens were all complete in themselves in that they didn't require word boundaries to be added and in most cases would break/behave unexpectedly if they were modified in an automated fashion.

Since the original rewrite we've added a bunch more global tokens that depend on these word boundaries. We intend to rewrite these to get back to the original, much more flexible regex structure.

lib/indexer/indexdocs.js
aarthykc: @sbma44, I think I removed that by mistake, will add it back !

ingalls: Hmm @aarthykc this may be a leftover from debugging

apendleton:
@ingalls per chat: I'm actually not sure about the current state of word boundary handling in this branch, but let's chat about it tomorrow; in current carmen it's actually subtly broken as it relates to non-ascii letters, which the global token code treats as word boundaries but the regular carmen tokenizer does not. I attempted to address this in [this branch](https://github.com/mapbox/carmen/pull/425) but never got it merged. Looks like it may now be moot, but worth confirming, especially as it relates to 'straße,' which I've confirmed has a JS regex word boundary in the middle of it ('ß').

ingalls:
@apendleton yes `ß` and just about any Unicode character is considered a word boundary. That said since this explicitly has `ß` it forces the word boundary match to be at the actual end of the word. It isn't ideal but this change retains the current behavior of regex. I'd definitely welcome a subsequent PR that improved this either with a custom flag as per your past PR or a better (but efficient) regex engine.

For the time being it's a good thing to keep note of and improve in the future but doesn't directly impact this work so far as we've found.

yhahn:
:ship: This is going to start carmen 17.x as this affects the format of global tokens!

ingalls:
👍 Woot!

PR #497, opened 2016-09-19
fix reverse address interpolation and use stricter eslint

ansis:
It will be easiest to review each commit separately. I can split the pr if that helps.

- fcdcf75 adds a more eslint rules related to variable definition and deletes a bunch of unused vars
- 1f6d3ee makes reverse address interpolation a tiny bit more accurate
- 5741c47 fixes a reverse address interpolation bug identified by the stricter eslint rules

@ingalls can you review at least the last two commits of this?


ingalls:
@ansis can you comment on the bug that this fixes? Reading through it looks good but not sure what to look for in relation to the bug.

ansis:
The commit message contains a bit of extra details. I should've note that or copied to the pr

It fixes reverse address interpolation for features with more than a single geometry in their geometry collections.

The `ranges` variable passed to `matchSide` used always be the last set of ranges for that feature. `ranges` is was set in the for loop and `matchSide` called it with whatever value it was last set to. It now passes the value that corresponds to `potential`.

ingalls:
Fantastic! I'm 👍

mourner:
Nice illustration of how useful strict linting can be! As a follow-up, I'd recommend upgrading to ESLint 3.5 (tons of changes and useful rules since 1.5) and extending some base config in the rules, like `eslint:recommended` (or maybe check out [eslint-config-mourner](https://github.com/mourner/eslint-config-mourner)).

PR #499, opened 2016-09-21
stricter eslint rules (fixed)

ansis:
re-applies fcdcf757c959a25452e890d191525324e4ce877a with a fix:

The current
`var combined = combined.sort(function(a, b) {`
is replaced with
`combined = combined.sort(function(a, b) {`
instead of fcdcf757c959a25452e890d191525324e4ce877a's broken version:
`combined.sort(function(a, b) {`

The `sort` modifies the array in-place but the chained `.filter()` does not:  https://github.com/mapbox/carmen/blob/5e3a306b6f57f8e805f0ce6c0f7df4b6d023ffcd/lib/context.js#L71-L76

@mattficke can you review?

### DIFF
--- a/lib/analyze.js
+++ b/lib/analyze.js
@@ -1,6 +1,5 @@
 var uniq = require('./util/uniq');
 var decode = require('./util/grid').decode;
-var termops = require('./util/termops');
 
 module.exports = function analyze(source, callback) {
     var s = source;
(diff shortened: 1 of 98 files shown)


mattficke:
@ansis looks good! Let me know when you want to go over deploying.

PR #500, opened 2016-09-22
Filter based on scaled scores

arunasank:
Add a new filter on carmen `scoreAbove`, such that if the user queries for `--scoreAbove 2`, return all features having `3 bit grid score * scorefactor` > `2 * scorefactor`.

cc/ @mapbox/geocoding-gang


ingalls:
Stale - Ok to close?

PR #501, opened 2016-09-27
Use the "maxscore" property for the scorefactor

arunasank:
Use the "maxscore" property instead of the maximum score from a sample of scores (freq) that we do currently to determine the scorefactor.

cc/ @mapbox/geocoding-gang


ingalls:
Closing: Tracking here: https://github.com/mapbox/carmen/pull/505

PR #503, opened 2016-09-27
Strip unused address properties from vector output

yhahn:
Removes interpolation properties from point address features and vice versa in vector output.

- [x] IRL integration testing
- [x] Size comparisons/tests with real datasets
- [x] Update proximityVector() to load features

cc @mapbox/geocoding-gang

### DIFF
--- a/lib/indexer/indexdocs.js
+++ b/lib/indexer/indexdocs.js
@@ -93,77 +93,69 @@ function parseDocs(docs, settings, full) {
             return res;
         } else docs[i] = res;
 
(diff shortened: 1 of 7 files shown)


yhahn:
Seeing about a 50% max tile size reduction in a mixed address point / interpolation index with a pretty good distribution of each type. Would love to squeeze more out of this if possible.

Per chat with @ingalls, I'm going to look next at loading top-level features in the `proximityVector()` codepath in `context()`. If we have this in place we can drop all address-related feature properties from our vector tiles.

PR #504, opened 2016-09-29
[ready] clarify callback scope; throw on carmen-cache error

sbma44:
could use some advice on what to do with these errors. pairs with https://github.com/mapbox/carmen-cache/pull/56


sbma44:
per chat with @yhahn, throwing is an appropriate response here.

bumps `carmen-cache` version to avoid a rare class of indexing error involving merge docs that have a key but no value, which previously threw a protozero error

PR #505, opened 2016-09-29
Extend `--types` to include subtypes.

arunasank:
Extend the `--types` filter to include certain subtypes.

Currently, `poi.landmark` is the only subtype available. 

Usage:

`--types=["poi.landmark", "poi", "place",... ]`

cc @mapbox/geocoding-gang


scripts/carmen.js
```diff
@@ -91,7 +91,8 @@ carmen.geocode(argv.query, {
     'debug': argv.debug,
     'stats': true,
     'language': argv.language,
-    'indexes': true
+    'indexes': true,
+    'scoreAbove': argv.scoreAbove
```
ingalls: We should probably drop this now that we are doing subtypes instead.

ingalls:
@arunasank Hey quick update here as to the new changes in the branch

- refactored a bit of code to use `carmen:idx` instead of mapping from geocoder_name/type
- merged in min/maxscore indexing code
- Added fatal errors for scorerange without maxscore
- Expanded landmarky tests

sbma44:
with apologies for `vim -d`'s horrifying default color scheme:

image

image

these seem like fine numbers! I think we can start thinking about the merge & deploy dance.

PR #508, opened 2016-09-30
Refactor types/limits code flow + requirements

yhahn:
Turns out the types/limits flow was already loading features per need in https://github.com/mapbox/carmen/pull/503. This PR does some cleanup/refactoring to make this clearer.

- `types + limit` reverse query mode is now only a concept handled by `reverseGeocode()`.
- `context()` always returns a single context.
- Adds `context.nearest()` for playing the role that the proximity context mode played before -- returns a flat array of `[ lon, lat ]` points that can then be `context()` queried for full features.
- Adds additional unit test to demonstrate that in types/limits mode reverse geocodes do indeed load full features/derive address points properly.

cc @ingalls for the review

### DIFF
--- a/lib/context.js
+++ b/lib/context.js
@@ -28,11 +28,8 @@ module.exports = function(geocoder, lon, lat, options, callback) {
     var full = options.full || false;
     var matched = options.matched || {};
     var language = options.language || false;
-    var limit = options.limit || false;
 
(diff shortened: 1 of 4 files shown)


ingalls:
This is a fantastic change. Makes reading the program flow a ton easier! I'm fully 👍 !

PR #511, opened 2016-10-04
Subtype factory reverse

ingalls:
Allow filtering by subtype for reverse geocodes.

cc/ @sbma44

### DIFF
--- a/README.md
+++ b/README.md
@@ -534,3 +534,31 @@ However, transliteration increases the potential for collisions between queries.
 For this reason, termops examines whether a given piece of text contains characters from the CJK (Chinese/Japanese/Korean) unicode blocks. If the text consists exclusively of such characters, a `z` is prepended to it. If there are any non-CJK characters, an `x` is prepended. This effectively isolates all-CJK tokens from everything else (including tokens that contain CJK characters alongside non-CJK characters).
 
 For clarity and simplicity, the above examples do not include these prepended chars. But in practice a query for `seattle washington` will be tokenized to `xseattle`, `xwashington` and `xseattle washington`.
+
+### geocoder_name, geocoder_type and combining indexes
+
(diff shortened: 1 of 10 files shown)


sbma44:
still tbd:
- [ ] unit test for nearestPoints' scoreFilter integration

sbma44:
ok @ingalls if this passes I think it's ready for some eyes. Especially wary of what's going on around `_exclusive` mode on contextVector -- I like your solution but I definitely got myself into a "this was passing but I hadn't written the code yet" situation 😬

sbma44:
hmm. I take it back, integration with api-geocoder is revealing problems:

- identical results on reverse queries with `limit > 1` -- I think we are going to need (address feature compatible) deduplication (ex: `./scripts/carmen.js --query="-43.210,-22.951" --types="poi.landmark" --limit=5 --config=../api-geocoder/dev/mapbox.places.js`)
- POIs making it out of reverse queries with `poi.landmark` type filtering. See above query for an example; I've confirmed the POI's `carmen:score` is 0

sbma44:
Issue one was easy -- scoreFilter can't be applied cleanly when `carmen:score` is undefined.  Detecting that condition and setting it to zero keeps the rabble out.

Issue two is trickier. Evaluating deduping strategies now.

sbma44:
Resolved the second issue by adding `carmen:address` to the deduping key (when present). unearthed a few bugs while I was at it.

sbma44:
This is looking pretty good on staging:
- types = poi returns pois
- types = poi.landmark returns landmarks

unfortunately:
- types = poi + poi.landmark returns only landmarks

This is contrary to the behavior we've defined for forward geocoding. Unfortunately I'm having some trouble getting unit tests to return the features I expect to set up a proper debug environment, which I think probably comes down to tile wackiness (coordinate precision at low zoomlevels, maybe?). I suspect this will be an easy one for @ingalls to spirit guide me on tomorrow morning.

PR #514, opened 2016-10-18
Switch to geobuf for internally storing original featuers [WIP]

ingalls:
Will flesh this PR description out if this experiment works.

cc/ @yhahn


lib/util/feature.js
```diff
@@ -241,7 +249,7 @@ function getFeatureById(source, id, callback) {
             return callback();
         }
 
-        var feature = loaded && loaded[id];
+        var feature = loaded && geobuf.decode(new Pbf(Buffer(loaded[id])));
```
yhahn: The check below makes it seem unsafe to assume you can decode `loaded[id]` here without checking for its existence first.

lib/util/feature.js
```diff
@@ -317,11 +329,21 @@ function putFeatures(source, docs, callback) {
...
+                    doc.geometry = {
+                        type: 'Point',
+                        coordinates: doc.properties['carmen:center']
+                    }
+                    delete doc.properties['carmen:center'];
+                }
                 delete doc.properties['carmen:hash'];
                 delete doc.properties['carmen:grid'];
+
+                current[doc.id] = Buffer(geobuf.encode(doc, new Pbf()));
```
yhahn: I think you'll explicitly want to base64 encode this buffer. At least in some versions of node JSON.stringify'ing a Buffer directly results in a huge array of numbers which is definitely not what we want.

ingalls:
Size changes were almost nil due to efficient compression of GeoJSON.

PR #516, opened 2016-10-19
Log scale

mattficke:
**Breaking Change**

Feature scores are indexed as a 3-bit value plus a per-index max, which approximates the raw score. To better represent the real-world distribution of scores, this uses a log scale for the 3-bit value, ~~and caps the scorefactor based on the max index value at 7x the median index maximum. Scorefactor values above the max cutoff are collapsed to fit within the range of 6x to 7x the median, while maintaining relative order.~~

**Updated**: No longer setting a cap on scorefactor.

Requires a global rebuild, since changes are made both at index time and at run time.

cc @yhahn


mattficke:
One more to-do: `maxScore` should be set at launch, the query-time setting I have now should just be used as a fallback. Started here since, in unit tests, no indexes exist when creating a new Carmen instance, but are added later with `addFeature`.

mattficke:
Testing reveals that capping the max score, at least in this manner, is not the right approach. Going to 🔪 that part.

test/termops.logScale3Bit.test.js
```diff
@@ -5,13 +5,37 @@ test('encode', function(t) {
     var term = 0;
     term = termops.encode3BitLogScale(3566,180000);
     t.equal(term, 5);
+
+    t.equal(termops.encode3BitLogScale(-1,10), 0);
```
yhahn: @mattficke the -1 value is failing (`NaN`) and based on reading codepaths it seems like there's nothing that'll guard against this in front of the log scaler. Would it make sense to handle/scale values < 0 to 0 ?

PR #523, opened 2016-11-01
make sorting more deterministic

### LINKED ISSUE #519: stack sort order is not completely deterministic
The stacks of subqueries are sorted before they are slice to a max of 30 stacks. This is the sort function: https://github.com/mapbox/carmen/blob/021e4df61d3f2897cb0b9a2296892e77f02157e7/lib/spatialmatch.js#L228-L233

It is possible for `a` and `b` to be different but for all four of the comparisons to return `0`. In this case they are sorted in an undefined order.

I hit this when a phrasematch change caused the input array to be in a different order, which caused the sorted stack array to be in a different order, which caused a stack to get excluded by the slice, which caused a test failure.
- [ ] fix this sort
- [ ] check all other sorts for undefined behavior

ansis:
fix #519 

This makes sorting of intermediate results more deterministic so that code changes have less of an impact on results. Without this, unrelated changes can trigger different sorts and cause test failures that are ok but different. This reduces the number of test failures that you need to debug.

The new sorting should only make things stricter. All sorts produced by the new versions could have been produced by the old functions.

@yhahn can you review or tag someone else?

I can merge this as part of #517 but I'd like to get it reviewed earlier so that I can update the tests with confidence.

### DIFF
--- a/lib/spatialmatch.js
+++ b/lib/spatialmatch.js
@@ -226,10 +226,14 @@ function stackable(phrasematches, idx, mask, nmask, stack, relev) {
 }
 
 function sortByRelevLengthIdx(a, b) {
-    return (b.relev - a.relev) ||
-        (a.length - b.length) ||
-        (b[b.length-1].scorefactor - a[a.length-1].scorefactor) ||
-        (a[a.length-1].idx - b[b.length-1].idx);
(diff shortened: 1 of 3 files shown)


ansis:
Adding this to #517, closing without merging.

PR #524, opened 2016-11-03
Let process.stdout finish before exiting

mattficke:
Writes to `process.stdout` aren't necessarily blocking, so calling process.exit() can sometimes end the process before indexing has finished. This change allows `process.stdout` to finish before exiting.

Node documentation reference: https://nodejs.org/api/process.html#process_process_exit_code

Many thanks to @ericfischer for assistance tracking this down.


yhahn:
:raised_hands: thx @ericfischer !!

PR #529, opened 2016-11-14
Multitype features

yhahn:
Big picture: allows a feature to move flexibly between multiple types if specially defined to do so. The final type of a multitype feature depends on the query, filtering options, and accompanying context.

- Adds support for `carmen:types` property on features -- this is an array of types supported by a single feature. This is optional, and the behavior of any features without this is as if it were defined as a single type based on its index `carmen:types: [ <type> ]`.
- Adds support for `geocoder_types` property at the source level. This should be defined for sources that contain multitype features so that carmen knows what additional types it might expect to find from that source.
- Moving/consoldating forward geocoding stack/type filtering to a separate lib.
- Unit tests for multitype scenarios.

### Remaining

- [x] Stronger unit test coverage of core context feature stacking logic
- [ ] Review from @mapbox/geocoding-gang


lib/context.js
```diff
@@ -82,74 +70,115 @@ module.exports = function(geocoder, lon, lat, options, callback) {
...
+            subtypeLookup[splitType[0]] = splitType[1];
+        else
+            subtypeLookup[splitType[0]] = true;
+    }
+    return subtypeLookup;
+}
 
-        var toFilter = options.full || false; // Only filter for reverse geocodes
-        for (var i = 0; i < res.length; i++) {
-            if (!res[i]) continue;
```
yhahn: This is the heart of this PR: for a given set of features retrieved/loaded from VTs at a point, how should they stack as a context object? Previously this was a pretty straightforward process with a little deduping across type -- if two features A and B were matched and had the same type there was some logic to negotiate only one of them winning.

Now with a feature potentially having multiple acceptable types, the code adds an additional inner loop where each acceptable type is tried in order of priority.

lib/util/feature.js
```diff
@@ -205,19 +205,13 @@ function getFeatureByCover(source, cover, callback) {
             console.warn('[warning] Feature not found: %s.%d (%s)', source.name, cover.id, source.id);
             return callback();
         }
+
```
yhahn: Refactoring here just reduces copypasta across 3 different code flows for loading features. Removes `carmen:geocoder_type` which I could find in use anywhere across our codebase.

lib/util/filter.js
```diff
@@ -0,0 +1,100 @@
+module.exports = {};
```
yhahn: This lib consolidates our logic to do source/feature filtering by `options.types` and `options.stacks`.

Still mulling on whether the filtering done in context queries can also be consolidated here -- for now, I've left that in place as it's got a pretty different code flow.

lib/context.js
```diff
@@ -408,6 +437,17 @@ function loadFeature(source, feat, full, query, language, callback) {
         loaded.properties['carmen:dbidx'] = dbidx;
         loaded.properties['carmen:vtquerydist'] = feat['carmen:vtquerydist'];
         loaded.properties['carmen:geomtype'] = feat['carmen:geomtype'];
+        loaded.properties['carmen:stack'] = source.stack;
+        loaded.properties['carmen:conflict'] = source.name !== source.type ? source.name : undefined;
```
yhahn: I'd love to eventually rename the `geocoder_name` concept to `conflict` which we're now essentially using exclusively as a way to declare a conflict between two types (e.g. address/poi => only one should be returned in a context stack)

sbma44:
:+1: this makes sense. I haven't traced through the type adjudication code in minute detail -- if you'd like that level of review, happy to do so. Makes sense big-picture though.

PR #532, opened 2016-11-17
use distance to geometry for proximity

ansis:
Previously `carmen:center` was used as the distance for proximity sorting. It now uses the distance to the nearest part of the geometry when the geometry is included.

This pr:
- adds a `distanceToGeometry(loc, geometry)` 
- uses the above function for calculating `carmen:distance` when the geometry is available.
- uses this `carmen:distance` value for proximity-based scoredist calculations

todo
- [ ] add a `geocode-unit` test that covers this
- [ ] test agains real data
- [ ] benchmark with real data
- [ ] review

@mattficke 

EDIT: I think I might have been mistaken about the availability of the geometry. Hmm


ansis:
It looks like geometries are only available for address features which get converted to a point already. They aren't included in the features in the index (because of huge size, I'm guessing). The geometries are in the vector tiles.

New plan:
Use the cover zxy to improve distance calculations in verifymatch and use 
- `closest(carmen:center, furthestTileCorner(zxy, proximity))`, OR
- `closest(carmen:center, tileCenter(zxy))`

These are both approximations. With the `furthest` approach, the distance would be off by 14km in a z12 source in the worst case.

Or we could always include all features that *may* be in the closest 10 results and do a more exact sort once the geometry is loaded.

Or we could load geometries before the sort and slice. I'm guessing this would be too slow, but with diskquest done it might be ok.

Either way we could load the geometries and do a more exact calculation for sorting the final results after their contexts are loaded.

mattficke:
Overtaken by #554

PR #536, opened 2016-11-21
[wip] remove unneeded vector props

mattficke:
To slim down vector tile size, this removes two types of unneeded properties:

* properties with null values
* the `carmen:center` property for features with point geom, since it was a duplicate of the `geometry.coordinates` value


mattficke:
stale

PR #537, opened 2016-11-21
add Phrasematch and PhrasematchResult types 

ansis:
this does two things:
- splits some source-level properties (`loadall`, `getter`, `nmask`, `bmask`) into a separate object to make it clearer that they are source-level and to avoid things like [picking a value off of the first phrasematch](https://github.com/mapbox/carmen/blob/7150b4adf4c81b41deb832ef6e38349120d84714/lib/spatialmatch.js#L188)
- adds classes for `Phrasematch` and `PhrasematchResult` so that they are more clearly and rigidly defined

The main changes are in `lib/phrasematch.js`. All the other changes are just adapting to that.

---

motivation:

One of the more confusing things about carmen right now is understanding what type an object is and what properties it has. Sometimes the members of an object are added in multiple places ([1](https://github.com/mapbox/carmen/blob/7150b4adf4c81b41deb832ef6e38349120d84714/lib/phrasematch.js#L31), [2](https://github.com/mapbox/carmen/blob/7150b4adf4c81b41deb832ef6e38349120d84714/lib/phrasematch.js#L39-L40), [3](https://github.com/mapbox/carmen/blob/7150b4adf4c81b41deb832ef6e38349120d84714/lib/phrasematch.js#L58-L66)) making it hard to keep track. This tries to improve things by using named js classes which has worked well in mapbox-gl-js.

@yhahn do you like this approach with classes or do you have other ideas?

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,5 +1,4 @@
-var path = require('path'),
-    EventEmitter = require('events').EventEmitter,
+var EventEmitter = require('events').EventEmitter,
     queue = require('d3-queue').queue;
 
 var dawgcache = require('./lib/util/dawg');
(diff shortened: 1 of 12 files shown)


yhahn:
@ansis capturing what we talked about in our voice convo here:

- :+1: to this change and to the move toward more types to define our data structures,
- Please stage/put this branch through the full integration testing gauntlet but once you feel confident there go for the :green_apple:

PR #538, opened 2016-11-21
promote smallest when identically named features are nested (aka squishy)

sbma44:
This PR adjusts the calculation of `relevance` in `verifyMatch` to give a small score boost to features nested inside of larger features with which they share a name. Consider the example:

```
place: New York
region: New York
country: USA
```

Currently, queries for `New York USA` will return the region feature because of the gappy penalty applied to the place feature. Even in the absence of the gappy penalty, region features will tend to win because of their greater score (or, perhaps more accurately, because of their layer's larger `scorefactor`).

This PR detects nested features with identical names (default text field only, synonyms are ignored). When present, the gappy penalty is negated and a small bonus is applied to promote the most deeply-nested feature. In some circumstances this bonus leads to a relevance greater than 1.0. That value is preserved internally until ops.toFeature where it is capped at 1.0.

In some cases it might be undesirable to promote identically-named features. For instance, consider a coffeeshop named `New York` that is located in New York (tacky, I know). A new boolean index property `prevent_promotion` may be used to prevent the application of this bonus to such features.

It's possible that I am not yet fully anticipating the side-effects that this change could cause, in which case I will be mostly-glad to sheepishly close this PR. Feedback very welcome.

cc @yhahn


test/ops.test.js
```diff
@@ -2,52 +2,52 @@ var ops = require('../lib/util/ops');
 var test = require('tape');
```
yhahn: What was the motivation for removing the relevance value in many of the tests here?

index.js
```diff
@@ -115,6 +115,8 @@ function Geocoder(indexes, options) {
             source.idx = i;
             source.ndx = names.indexOf(name);
             source.bounds = info.bounds || [ -180, -85, 180, 85 ];
+            if (info.prevent_promotion)
+                source.prevent_promotion = true;
```
yhahn: Super small style thing -- would it be more consistent with defaulting of other properties (and making sure this property is always set to _something_) to write this as

```
source.prevent_promotion = info.prevent_promotion || false;
```

test/ops.test.js
sbma44: the relevance value test looked busted to me, a likely byproduct of the transition to `carmen:` prefixed keys. Specifically:
- the relevance value was being placed on the context object members rather than the context object itself
- the tests were asserting that the property comes back `undefined` which, yup, it was, because the property was never accessed
- `carmen:relevance` doesn't appear in the codebase outside of that test
- some git blaming showed components of these tests dating back to garrett

test/geocode-unit.types.test.js
```diff
@@ -33,7 +33,7 @@ tape('index region', function(t) {
         id:1,
```
sbma44: yeah good point, didn't pay enough attention to the original intent here

lib/util/ops.js
```diff
@@ -106,7 +106,7 @@ function toFeature(context, format, language, debug) {
         type: 'Feature',
         text: featureText.text,
         place_name: place_name,
-        relevance: context._relevance,
+        relevance: Math.min(context._relevance, 1.0),
```
yhahn: This can be rolled back now with the score approach, ya?

lib/verifymatch.js
```diff
@@ -239,13 +238,22 @@ function verifyContext(context, strict, loose, indexes) {
     var lastgroup = -1;
     var relevance = 0;
     var direction;
```
yhahn: @sbma44 want to dump a solid few sentences in here explaining what `squishy` is attempting to do (and provide a New York, New York style example?)

sbma44: roger that, done

lib/util/ops.js
sbma44: yup, removed

PR #539, opened 2016-11-21
use exceptions instead of string return values for error handling

ansis:
- switches from returning strings for errors to throwing exceptions
- cleans up a couple minor things like string concatenation and long lines

---

- [x] fix error / tests
- [x] review @yhahn (or assign to someone else)
- [x] stage indexing and test


ansis:
Tested indexing against a real index.

PR #541, opened 2016-11-23
a bit of cleanup

ansis:
A bunch of small changes that (I think) make the code clearer.

- [x] @yhahn review or assign to someone else


ansis:
> yhahn	@ansis ok, the second two both look good

PR #544, opened 2016-11-24
context: Defer language resolution

yhahn:
**The problem:**

- The goal of `squishy` was to be able to promote a feature like `New York (place), New York (region)` past `New York (region)` on the basis of its text matching its parents.
- When localization is introduced, there's some ambiguity about what ^^ means. Is the text supposed to match between localized versions? Non-localized versions? What if there's some mixture of scripts or language because of language fallback handling?
- Currently in `master` you can lose the effect of the `squishy` adjustment when using the `language:` option because `context()` is resolving the localized version of `carmen:text` early, but the top feature in the stack is being resolved late.

**The fix:**

- `context()` is going to pass all `carmen:text*` properties onto the caller. It's the caller's job to localize text and clean up the properties for final output.
- The `squishy` logic is going to always operate only on `carmen:text` before it's been localized, for now. So there's no confusion about whether we're squishing based on localized text, fallback logic, etc.

This PR adds unit tests for the behavior above and makes the adjustment to context.

------

### TODO

- [ ] Integration/IRL tests + staging

cc @sbma44 @mapbox/geocoding-gang

### DIFF
--- a/lib/context.js
+++ b/lib/context.js
@@ -5,8 +5,7 @@ var mapnik = require('mapnik'),
     queue = require('d3-queue').queue,
     Locking = require('locking'),
     addressCluster = require('./pure/addresscluster'),
-    addressItp = require('./pure/addressitp'),
-    closestLang = require('./util/closest-lang');
+    addressItp = require('./pure/addressitp');
 var cover = require('tile-cover');
 
 // Returns a hierarchy of features ("context") for a given lon, lat pair.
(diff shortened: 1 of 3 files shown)


sbma44:
woof great catch. I stupidly assumed that `carmen:text` always corresponded to the default text field. In retrospect it should have been obvious that that was not the case given the brevity of the context object entries.

PR #546, opened 2016-11-28
Score inheritance (v3)

yhahn:
- Switch score inheritance from opt-out to opt-in on a per index basis to make this feature 100% optional: `prevent_promotion => geocoder_inherit_score`
- Compares `carmen:text` and all `carmen:text_{language}` strings (between the same language codes) for the target and candidate features.
- Considers text matching if target text is fully present as a substring of the candidate text.

Functionally this helps cover cases like:

- Bern (city) within, Canton of Bern (administrative region)
  - Requires substring text match between candidate + target
- Inconsistent use of `carmen:text` (known issue where the language is often indeterminate/inconsistent) making one or more of the `carmen:text_{language}` properties often more suitable for consistent/robust comparison
  - Requires considering all `carmen:text*` fields

------

Running IRL tests to make sure this is a safe change. While this alters an API (`prevent_promotion`), as this API was internal/undocumented I feel ok about this being a minor version bump.

cc @sbma44 @mapbox/geocoding-gang

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,6 +69,7 @@ geocoder_types          | Optional + advanced. An array of type strings. Only ne
 geocoder_version        | Required. Should be set to **6** for carmen@v11.x. Index versions <= 1 can be used for reverse geocoding but not forward.
 geocoder_cachesize      | Optional + advanced. Maximum number of shards to allow in the `carmen-cache` message cache. Defaults uptream to 65536 (maximum number of possible shards).
 geocoder_address_order  | Optional + advanced. A string that can be set to `ascending` or `descending` to indicate the expected ordering of address components for an index. Defaults to `ascending`.
+geocoder_inherit_score  | Optional + advanced. Set to `true` if features from this index should appear above other identically (ish) named parent features that are part of its context (e.g. promote New York (city) promoted above New York (state)). Defaults to `false`.
 
 *Note: The sum of maxzoom + geocoder_resolution must be no greater than 14.*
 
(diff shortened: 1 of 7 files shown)


sbma44:
> `prevent_promotion => geocoder_inherit_score` 

strongly support the name change and can see the case for making it opt-in, though my guess is still that forgetting to turn this on will bite us more often than forgetting to turn it off

> Compares carmen:text and all carmen:text_{language} strings

I had imagined rev2 of this feature (post-`carmen-data`) moving toward an optional property that binds features together without affecting labeling. I chose the default text field for now because its content is a bit arbitrary and it's on our roadmap to decline in importance. I've got a feeling that we'll start seeing ways this system fails when it's purely algorithmic the longer it's in production. I also don't feel like I have enough knowledge of other languages to predict where substring matching will bite us -- I'd feel a lot safer keeping this limited to latin alphabet languages.

No objections to getting this done to better satisfy TWC requirements in the short term, of course.

yhahn:
> I had imagined rev2 of this feature (post-carmen-data) moving toward an optional property that binds features together without affecting labeling. I've got a feeling that we'll start seeing ways this system fails when it's purely algorithmic the longer it's in production.

Definitely.

When I visualize how we'll move to a property in the future I think we'll end up scripting some kind of likely-candidate generation for manual review because there's just so many cases where this logic is appropriate out there. In a sense I'm thinking of this code in carmen as a prototype for that first pass.

Strangely with `carmen:text` I'd say that it counterintuitively is the least reliable across indexes of our current fields because its language is underspecified.

Overall I feel ok about the surface area for false-positives this exposes -- worst case scenario is we reorder features in a resultset that passed all our earlier checks and are tied in relevance.

PR #548, opened 2016-11-30
Handle Un-decodable Text like Emojis

KaiBot3000:
We previously failed to properly decode emojis, but our character set check resulted in these queries becoming `'x'` instead of `''` after decoding. 

This PR:
- Checks the length of the text after the character check, and returns `''` for un-decodable text.
- Tests for correct behavior
- Tidies up some docstrings and a variable name.

Fixes https://github.com/mapbox/api-geocoder/issues/1129

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 17.6.1
+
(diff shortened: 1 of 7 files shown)


yhahn:
:boom: :ship:

PR #550, opened 2016-12-02
Squash more emojis

yhahn:
Continues quest from https://github.com/mapbox/carmen/pull/548 to squash input text that causes more work than necessary.

- Adds a testcase for emoji strings that tokenize to multiple tokens (e.g. have a space between two emojis)
- Adds a testcase for how a long "emoji phrase" behaves with many indexes in play and asserts that carmen can reject the query quickly

### Optimization in `tokenize()`

WIth just the token cleanup we'd end up with tokenized phrases like this:

> ":smile: :smile: :smile: :smile:" => ["", "", "", ""]

That would go straight into subquery permutation generation (https://github.com/mapbox/carmen/blob/6332030c8c9cf80ec4bf42ef10a0fd89a7313bf9/lib/phrasematch.js#L26-L27) so we would still do _some_ work for these queries.

This PR changes this so that when tokenizing we also drop any tokens that unidecode to nothing.

cc @mapbox/geocoding-gang @KaiBot3000


yhahn:
@KaiBot3000 when you have a chance would love a review to make sure this all looks good/jives with what you were seeing.

KaiBot3000:
Looks good, @yhahn. I assumed that the tokenization step would take care of cases where we had multiple emojis but I didn't account for having extra whitespace. Good catch 👍

yhahn:
> I assumed that the tokenization step would take care of cases where we had multiple emojis

me too :sweat_smile:

PR #553, opened 2016-12-05
check if tmpid already in sets

mattficke:
When there are multiple covers with the same tmpid, the lowest-scored one ends up in the `sets` object instead of the highest-scored. This change checks for the existence of `sets[covers[j].tmpid]` to avoid overwriting a higher-scored cover with a lower-scored one.


mattficke:
Working on adding tests for this, and I'm running into trouble generating covers that only match part of the text. For example, in the case of `United States` in our production index, there will be covers with a `text` value of just `States`, due to the frequency of the term `United` in the index.

However, the `addFeauture` script we use for testing adds features one at a time, rather than in the batches of up to 10,000 that `indexdocs` does when we're doing this for real. This prevents us from really testing the effect of term frequency in our unit tests.

I think this is best tackled in 2 broad steps.
1. Allow `addFeature` to index unit test features in batches
2. Add frequency logic similar to our production indexer to `addFeature`

Thoughts on next steps:

- [ ] Modify `addFeature` to accept batches of docs, rather than one at a time as it's set up now
- [ ] Set up a unit test that builds a batch of docs to be indexed and sends them to `addFeature` all at once, make sure this works
- [ ] use `indexdocs.generateFrequency`, `termops.getIndexablePhrases`, and other related functions as part of the indexing process in `addFeature`.

The last step is the one that I'm not quite sure how to do yet, but I'll tackle that once I get batch indexing working.

mattficke:
Turns out the weighting of terms will be handled automatically once features are added in batches, so this is pretty close to being ready.

Done:

* Allow batch adding of features in unit tests
* Spam the index with enough features to affect term weighting

To-Do:
* fool the geocoder: need to create two features with geometry that, from the z/x/y tile coords, look like they stack (and so make it through spatialmatch), but don't when looking at the real geometry.

mattficke:
I think I finally synthesized this in a unit test. The problem occurs when features with similar names overlap in z/x/y space:
```
...xxxxxxx <== United States (country.1)
xxxxxxx... <== United States Minor Outlying Islands (country.2)
...xxxx... <== Midway Atoll (region.3)
.......xxx <== Midway, KY (place.4)
```

A query for `midway united states` should return the region in the first position, with a relevance of 1, since there is no gappy penalty applied. `Midway, KY` should have a very slightly lower relevance since there's an index gap. This doesn't always happen, though.

1. User searches for `midway united states`.
2. In spatialmatch, top result is for `[Midway (region.1), United States (country.1]`. These features stack in z/x/y space but not in the real world, which will cause our problem.
3. The correct stack does not make it into the `results` array, since the incorrect result already claimed that slot. However, it does make it into the `sets` object, which is used for "loose" matching later.
4. The cover that matches the `United States Minor Outlying Islands` feature with the full query text (`united states`) is overwritten in `sets` by a cover that only matches `states`. This cover has a lower relevance since it's only a partial text match.
5. Over in verifymatch, relevance is calculated for Midway Atoll by reverse geocoding the feature centerpoint and comparing the results in `context` against both the `results` array and the `sets` object.
6. Since the correct country feature didn't make it into the `results` array, the `sets` object is checked next. The feature is there! But it's the one with a lower relevance due to the partial text match.
7. `Midway, United States Minor Outlying Islands` is buried in the results due to its low relevance.

This bug is frequently masked since a common situation is for the final cover to be something like a solo `united states` feature, with a relevance of 0.5. This is lower than, say, the 0.6 relevance of a `midway, states` stack, so all is well. However, in situations with a very large number of matches, or when the type filter is used, this final country feature doesn't make it into `sets` and the problem shows up.

mattficke:
🤔 `addfeature.test.js` passes locally for me, but not on CI. Poking at it, but not sure what the difference would be.

mattficke:
Per @yhahn's suggestion, I added a reset function to clear out any in-memory cache (similar to  the one in the [io-reverse test](https://github.com/mapbox/carmen/blob/master/test/geocode-unit.io-reverse.test.js#L64-L78)), which seems to have resolved the inconsistent behavior.

I'm not sure if the underlying cause here is adding features in batches, or adding features with a large number of tiles -- our unit tests typically only have one tile. I'd like to figure out when we'll need to use a reset function like this again in future tests.

yhahn:
@mattficke :heart_eyes: awesome work tracking this down and representing the problem in the testcase. In addition to this fix I think the new batch AddFeature will unlock testing for a bunch of other subtle corner cases. ^_^

:ship:

Issue #549, opened 2016-12-02
not writing our own key-value store


ansis:
*disclaimer: I don't know that much about carmen-cache. just raising some questions*

Carmen-cache is a self-written database. This may have made sense when we were loading parts of the index on demand, but now that we're just getting stuff from disk* we should think about using something else. We don't need to shard by key anymore so carmen-cache becomes a regular key-value store that reads from disk. Databases are hard so we should use one written by someone else instead of writing our own. Something like [RocksDB](http://rocksdb.org/).

At index-time add everything to the db. Transfer the entire db to s3. Download it and use it.

\* *ok, not completely yet*

### I'd hope this would:

- remove a lot of complexity and code from carmen-cache
- make random accesses of features and grid lists significantly cheaper
- reduce complexity in carmen and increase accuracy by letting us do things that used to be expensive:
    - loading features for more/all spatialmatches instead of the 10 best
         - get rid of approximate scores! access the list of languages from the features!
    - if grid lookups are cheaper, you could load grid lists for all phrasematches
        - you could reverse some of the logic of `stackable` and `coalesce` by matching by tile (roughly linear) before creating the stacks. This could be faster and get rid of approximations.
    - this might make it easier to implement degenless/strings-attached
- step towards making it viable to run carmen locally with all the data

### Potential challenges:
- would this be less space-efficient, and if yes, by how much?
- how hard is it to drop s3/sharded sources completely

----

How I started thinking about this:
"Why do we have to have all these approximations? Accessing things used to be really expensive. Now that things are on disk can we get rid of some? Not sure, but it seems like it should be possible to make accesses even cheaper. What would change in carmen if accessing things from the index were cheaper?"

@apendleton @yhahn 
I know there were good reasons for doing it the way it's done, but now that some things have changed do you think using a real key-value store would improve performance and reduce complexity?

apendleton:
Not opposed, but I think it will be hard.

> would this be less space-efficient, and if yes, by how much

If we preserve the current delta-encoding of grids and use protobuf varints for grid lists, probably won't change by that much, but then we're still largely inventing our own database, so we'll need to carefully consider whether it's worth it. If not, I think it will be much less space-efficient.

> this might make it easier to implement degenless/strings-attached

[...]

PR #554, opened 2016-12-05
improve proximity

ansis:
The carmen-cache pr: https://github.com/mapbox/carmen-cache/pull/59
There is an internal branch that adds some proximity integration tests.

## high level stuff

How do we define what the proximity function *should* do? One idealish definition:
Return the feature that people at *location* would most likely be referring to with *query*.

Eventually if we have `(query, expected_feature, user_location)` triples collected from real queries we could use that to tune the implementation but for now we can just manually come up with a bunch of test cases and tune based on those.

Currently Carmen tries to approximate this by combining score and distance. This pr sticks with that approach.

Other APIs bias based on whether a feature is within a bbox or region, instead of based on distance. In some ways this is more limited but it is also more straightforward. So far I've been working on a way to make a scoredist approach better but maybe we should consider something like this?

### improving distance calculations

8efff90, bce379f and https://github.com/mapbox/carmen-cache/pull/59 improve the accuracy of the distance calculations. More accurate distances should mean more accurate scoredists which should mean slightly better results.

### improving scoredist

The main changes will come from changing `scoredist`. The current scoredist function has a couple problems:
- results further than ~40 miles aren't biased based on proximity
- results that are really close are sorted purely based on distance, ignoring score
- the presence of result *c* in intermediate results can influence the order results *a* and *b* are sorted

The functions I've tried on my test cases are `(1 + score) / (1 + distance)` and the slightly more complicated `(score + 1) / (Math.pow(dist, 1.1) + 1)`. They seem pretty promising and I think by tweaking the constant numbers a bit more we could get something good.

### remaining problems

Grids are scored and limited based on approximate scores before the more accurate scores are loaded. Currently the approximate scores can be orders of magnitude different than the actual scores. This can result in approximate scoredists being orders of magnitude different which makes the scoring and sorting very unreliable.

The approximate scores are more problematic for proximity sorting than for score sorting because score sorting only cares about the order of scores while proximity scoring also cares about the relative values of scores.

[Log scale approximate scores](https://github.com/mapbox/carmen/pull/516) should help with this but it's hard to tell how much until that's merged. If that doesn't work well enough, I'm not sure what the next steps are. Maybe switching to more boolean biasing like bbox and region biasing? Maybe trying to get rid of approximate scores (https://github.com/mapbox/carmen/issues/549)?

---

- [ ] wait until https://github.com/mapbox/carmen/pull/516 is merged and re-test accuracy
- [ ] add some more tests and dig into results to see if/what other things can be done while waiting for ^

@mattficke what are your thoughts? does the general direction seem right?


mattficke:
@ansis This is great, I think you're on the right track. It sounds like the log scale score simplification is now the blocker for moving forward with the scoredist improvements, so let's work on shipping that as soon as possible this week. I've lowered the concurrency of the carmen-indexer stack to 25 (from 50) to avoid running out of CPU in the container, so hopefully this will make a global rebuild possible within a reasonable timeframe.

I'm also definitely interested in exploring other approaches like this:
>Other APIs bias based on whether a feature is within a bbox or region, instead of based on distance.

How does this work in practice with the other APIs? What would it look like with ours?

mattficke:
Ok, I think this is pretty close. I stripped out a bunch of the changes so that we can ship more quickly, and iterate from there.

The main change here is that distance from the proximity point is now calculated as the lower of either:
1. Distance to feature centerpoint
2. Distance to the tile cover edge

The companion pr in carmen-cache has been updated to return the tile cover closest to the proximity point when other sorting values are equal.

The radius of effect in the scoreDist calculation is now scaled to zoom level. The status quo holds at zoom 14 (which we use for addresses and POIs), and the constant is doubled at each lower zoom level:
```js
function scoredist(meanScore, dist, zoom) {
    var radius = 40 * (15 - zoom);
    return Math.round(meanScore * (radius/(Math.max(dist,0.0001))) * 10000) / 10000;
}
```
Still need to clean things up, but the substantive work is here.

PR #569, opened 2017-01-10
[ready] don't index degens for synonyms

sbma44:
### Context
Autocomplete is suppressed for text fields other than the default in order to minimize the number of unexpected matches that a user encounters. This was motivated by the high false positive rate and unintuitive behavior delivered by unidecoding everything to a shared character set.

Currently, default text synonyms **are** indexed for autocomplete. This also results in unexpected behavior, as synonyms are never used for display -- only the first term in the default text field is. Consider a feature with the default text `Latveria,Republic of Latveria`. Right now an autocomplete query for `re` can return `Latveria`.

### Fixes/Adds
- Disables degen (autocomplete) indexing for all synonyms.
- Updates tests

### Next Steps
- review (@apendleton / @yhahn, have a sec for 👀 ?)


sbma44:
Tests are green, so barring any additional feedback I'm going to merge and publish tomorrow AM.

apendleton:
Implementation looks good to me. Things to note, though:
* when degenless lands we're going to have to do this again, so I'll need to think through how to do that. And both this change and that one will of course require global rebuilds :(
* long-term, you should be able to search for `Republic of Latveria` (and especially so with places that have two equally natural synonyms, rather than this one where there's one reasonable default and one longer/more formal one). The problem is the display, not the indexing. But I don't know how to deal with that yet either.

Issue #246, opened 2015-04-17
Provide ISO 3166-1 alpha-2 codes for countries


sandstrom:
This is an geocoder!

A small but wonderful improvement would be if it returned iso codes[1] for countries.

[1] http://en.wikipedia.org/wiki/ISO_3166-1_alpha-2

sandstrom:
@yhahn Friendly ping :smile: what are your thoughts on this?

sandstrom:
@yhahn friendly ping :boat:

ingalls:
Hey @sandstrom, this repo is for the geocoding engine that powers both the [mapbox geocoder](https://www.mapbox.com/developers/api/geocoding/) and also for self hosted data. 

Carmen already has the ability to return custom fields. Anything in the `properties` object will be returned - see https://github.com/mapbox/carmen/blob/master/test/util/addfeature.js#L22 If you are indexing your own data simply add a new Key/Value in the `properties` object with the data you want to return. 

We are also working on adding this data to the mapbox geocoding api but don't have a timeline as of yet.

Let me know if you run into any problems

sandstrom:
@ingalls Ah, that makes sense!

Great to hear that it's coming to the Mapbox geocoding API!

sandstrom:
@ingalls Hey Nick! Just wanted to drop a note thanking you for adding the ISO-3166-1 code to the response (though it's outside the scope of this repo). Awesome! :smile: 

Should anyone else have the same question, the response now looks like this:

```
{
    "id": "country.4404326516146160", 
    "short_code": "ca", 
    "text": "Canada"
}
```

ingalls:
Hey no problem! Thanks for waiting and working with us on this one, took a bit of rewriting things under the hood to get it working as expected.

We're hoping to expand the metadata to include WikiData IDs in the future to allow more flexibility in linking with external datasets.

sandstrom:
@ingalls I have some trouble understanding the distinction between carmen and mapbox, but I think this may be the right place.

**Question:** Why do some features omit the region in their context array?

```js
{
  "type": "FeatureCollection", "query": ["birge"], "features": [{
    "id": "address.9808295726271380",
    "type": "Feature",
    "place_type": ["address"],
    "relevance": 0.99,
    "properties": {},
    "text": "Birger-Forell-Platz",
    "place_name": "Birger-Forell-Platz, Berlin, 10713, Germany",
    "center": [13.32, 52.4813968],
    "geometry": {
[...]

PR #572, opened 2017-01-16
Multitype backfill + place_type

yhahn:
This softens the behavior of the previous PR adding multitype behavior (https://github.com/mapbox/carmen/pull/529) and adds the `place_type` key to provide better visibility into multitype features in response data.

- If a multitype feature and other normal features may be returned for the same type level for a given reverse geocode, the normal features will take precedence. This changes the previous behavior where a multitype feature would override other features at the promoted level.
- Include new `place_type` property for enumerating all the possible types of a returned feature. Usually this is an array of just one element -- the type of the feature.

### DIFF
--- a/carmen-geojson.md
+++ b/carmen-geojson.md
@@ -11,6 +11,7 @@ Carmen returns results as a [GeoJSON](http://geojson.org/) `FeatureCollection`:
                 "id": "place.4201",
                 "text": "Austin",
                 "place_name": "Austin, Texas, United States",
+                "place_type": [ "place" ],
                 "bbox": [-97.9383829999999, 30.098659, -97.5614889999999, 30.516863],
                 "center": [-97.7559964, 30.3071816],
                 "geometry": {
@@ -51,6 +52,7 @@ key | description
 id | Id of the feature of the form `{index}.{id}` where index is the id/handle of the datasource that contributed the result.
 text | Text representing the feature (e.g. "Austin").
 place_name | Human-readable text representing the full result hierarchy (e.g. "Austin, Texas, United States").
+place_type | An array of index types that this feature may be returned as. Most features have only one type matching its id.
 bbox | Optional. Array bounding box of the form [minx,miny,maxx,maxy].
 address | Where applicable. Contains the housenumber for the returned feature
 center | Array of the form [lon,lat].
--- a/lib/context.js
+++ b/lib/context.js
@@ -96,12 +96,6 @@ function getSubtypeLookup(types) {
 function stackFeatures(geocoder, loaded, options) {
     if (!loaded.length) return [];
 
-    var typeOrder = Object.keys(geocoder.indexes).reduce(function(memo, id) {
-        var index = geocoder.indexes[id];
-        if (!memo[index.type]) memo[index.type] = memo._counter++;
-        return memo;
-    }, { _counter:0 });
-
     var context = [];
     var memo = {};
     var firstType = false;
@@ -118,11 +112,7 @@ function stackFeatures(geocoder, loaded, options) {
 
         for (var l = feature.properties['carmen:types'].length - 1; l >= 0; l--) {
             var type = feature.properties['carmen:types'][l];
-            var promoted = l > 0 && options.full;
             var conflict = feature.properties['carmen:conflict'] || type;
-
-            // Reconstruct extid based on selected type
-            feature.properties['carmen:extid'] = type + '.' + feature.properties['carmen:extid'].split('.').pop();
 
             // Disallow shifting a feature's type to occupy the maxtype
             // The maxtype is set on forward geocodes by the matched feature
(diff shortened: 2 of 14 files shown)


sandstrom:
@yhahn Is the place array always in the same order?

It seems to be from least specific to most specific, i.e. country first (if available).

```ruby
  {
    "id" => "place.364",
    "type" => "Feature",
    "text" => "Hong Kong S.A.R.",
    "place_name" => "Hong Kong S.A.R.",
    "place_type" => ["country", "region", "place"], # <-----
    "relevance" => 0.49,
    "properties" => {
        "short_code" => "hk", "wikidata" => "Q8646"
    },
    "bbox" => [113.733669, 22.134324, 114.528665, 22.571154],
    "center" => [114.135442, 22.346578],
    "geometry" => {
        "type" => "Point", "coordinates" => [114.135442, 22.346578]
    }
  }
```

yhahn:
@sandstrom yes, and the first item is always the canonical type of the feature

sandstrom:
@yhahn Thanks! Much appreciated 😃  we ran into some bugs with Mapbox after this change, trying to figure out the best way of parsing them. This info was helpful!

***

We parsed type from `id`, but since HK doesn't have any context, which is expected for `place`, our parsing failed. Will update to parse type based on `place_type`.

sandstrom:
@yhahn There is a lingering issue, unfortunately.

In the example above the returned feature was canonically a country (according to your comment above, first item in `type_array` is the canonical type). So I modified our parsing to give precedence to the first entry in `type_array` (if present), and only use first part of `id` as a fallback.

It then makes sense (for the entry above), as a country it's expected that there is no context.

However, this query (below) returns a feature of type `place`, with no `place_type` property to indicate that its canonical type is anything else. As such, we cannot interpret it as anything but a `place`.

However, it doesn't have any context property, which all other places have.

Is this by design or an error?

```ruby
    {
        "id" => "place.364",
        "type" => "Feature",
        "text" => "Hong Kong S.A.R.",
        "place_name" => "Hong Kong S.A.R.",
        "relevance" => 0.99,
        "properties" => {
            "short_code" => "hk", "wikidata" => "Q8646"
        },
        "bbox" => [113.733669, 22.134324, 114.528665, 22.571154],
        "center" => [114.135442, 22.346578],
        "geometry" => {
            "type" => "Point", "coordinates" => [114.135442, 22.346578]
        }
    }
```

https://api.mapbox.com/geocoding/v5/mapbox.places/H.json?types=region%2Cplace%2Caddress&autocomplete=true&access_token=redacted&limit=5&proximity=15.0%2C62.0

mattficke:
@sandstrom if this exact query is one you've tried before, it's possible you're seeing a cached response. You can ensure you're not seeing cached data during testing by appending a parameter such as `&_=1234` to the end of the query.

I just tried your query and am seeing the `place_type` property for the Hong Kong feature. Let us know if you're still running into any issues.

image

sandstrom:
I tried again and now I'm getting HK as a region under China. I'm 100% certain the response I got wasn't cached though (at least not in my http client, perhaps on the server, that's outside my control). I guess the underlying data set has changed.

ingalls:
Hey @sandstrom just seeing your message here! Sorry for the changes as you are talking through a bug, I've been furiously updating our country & region layers over the last couple of days. It's been a bit of an iterative process so lots of different changes going out across deploys.

### Changelog

- `Hong Hong S.A.R.` is now `Hong Kong` for all languages
- `Honk Kong` (And `Macau`) are now a `region` that fall within `China`
- `Hong Kong` is also a `place` feature. HK used to be a country, region & place feature which lead to ugly results like `Hong Kong, Hong Kong, Hong Kong SAR` We added what we call `multitype` features to allow us to merge features across layers. Features that are shared between layers can be identified by a `place_type` array length of >1 in the API results.
- `Taiwan` is untouched for now (`country`) to keep the status quo. 

### TODO

- Good catch on the `short_code` property. I'll get that changed on the data side later today and it should go out sometime next week. 

Huge apologies for causing a headache on your end. It's always a battle on how to best represent complicated country/region politics in an API. If you have any more questions I'm happy to help!

sandstrom:
That explain things, the difference in returned API results has been confusing. The multitype thing is good, agree it cleans up results.

Will test anew next week and see if it works!

Despite what Xi Jinping says, until China invade Taiwan is a de-facto country. Perhaps your new president can bargain away their freedom 😄 

Overall I think the API is pleasant to work with. Awesome with the wikidata and short code additions (we've [talked about that](https://github.com/mapbox/carmen/issues/246) earlier).

sandstrom:
@ingalls Another question on geodata 😄 

A query for "Beijing" currently return a place feature, with a context containing only the country (China). Compare to a query for their old capital "Nanjing" where the region and country are in the context chain.

All is well so far.

But the `place_type` is still only `['place']`, not `['place', 'region']`. Also, there is no `short_code` property on the feature (would expect `CN-11` for Beijing).

According to this document, these four city/regions are both a city and a region:
https://en.wikipedia.org/wiki/Direct-controlled_municipalities_of_China

Let me know what you think! Does my reasoning make sense? 😄 

***

*This may be my last question/issue, then I shall give you peace :) Seems like we may need to migrate off of Mapbox, we need to switch to the enterprise plan but probably cannot afford it. Anyway, I like Mapbox, it's a good product!*

boblannon:
Thanks, @sandstrom!  I'll open a ticket to address this.

sandstrom:
Awesome @boblannon

PR #576, opened 2017-01-20
Fix stats output for scripts/carmen.js

springmeyer:
### Context

The `--stats` flag reports the timing of each major operation: `warmup`, `phrasematch`, `spatialmatch`, and `verifymatch`.

### Fixes/Adds

The `warmup` value was broken: It was reporting a date rather than a delta. Looks like this went missing in https://github.com/mapbox/carmen/commit/9d78a3115d01be874bc6ed1b77c2976c12ae67b3

### Summary

This fixes the `warmup` value by ensuring it represents elapsed time.

### Next Steps

Do we need tests of the command line tools like `scripts/carmen.js`?

### DIFF
--- a/scripts/carmen.js
+++ b/scripts/carmen.js
@@ -94,6 +94,7 @@ carmen.geocode(argv.query, {
     'indexes': true
 }, function(err, data) {
     if (err) throw err;
+    load = +new Date() - load;
     if (data.features.length && !argv.geojson) {
         console.log('Tokens');
         console.log('------');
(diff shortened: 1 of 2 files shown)


mattficke:
>Do we need tests of the command line tools like scripts/carmen.js?

Yes, we have CLI tests in `tests/bin.test.js`. I'll take a pass at adding one so I can take this along with me in the new Carmen@18.0.0 release

PR #577, opened 2017-01-21
use namespaced turf

mattficke:
### Summary
To avoid `npm install` warning fatigue, updates deprecated turf modules with the `@turf` namespaced versions.


sbma44:
Thanks a ton for grabbing this -- had been on my "boy I should really do that, huh" list for a while

PR #578, opened 2017-01-23
Merry easter

ingalls:
### Context

- Update a few old deps.


mattficke:
bringing along with #577

PR #579, opened 2017-01-23
Use Number for hard error on non-numeric input

ingalls:
### Context

If a string loosely fit the format of a reverse query `Number , Number` it could be interpreted as such even when actually being a forward query.

This was due to `parseFloat` ignore non-numeric characters and just grabbing any number it sees.

Ref: https://github.com/DenisCarriere/geocoder-geojson/issues/8#issue-201756354


DenisCarriere:
Woot woot! 🚀 ⭐️

PR #580, opened 2017-01-24
languageMode: strict

yhahn:
### Context

Adds new querytime option `languageMode` which can be set to `strict` to limit returned features to only those that fully match the language specified in the `language` option.

- Refactors/reduces boilerplate in final feature output logic to reduce sprawl for this change
- Adds tests for forward, reverse, reverse (typed) for `languageMode: strict`
- Adds note in README about this option

cc @mapbox/geocoding-gang


yhahn:
I apologize for any painful merge conflicts this refactoring may cause :sweat_smile: ... but I think the code consolidation will feel good after the pain!

index.js
```diff
@@ -77,12 +77,15 @@ function Geocoder(indexes, options) {
...
-            for (var ix = 0; ix < keys.length; ix ++) {
-                if (/geocoder_format_/.test(keys[ix])) source[keys[ix]] = info[keys[ix]]||false;
-            }
+            // Fold language templates into geocoder_format object
+            source.geocoder_format = { default: info.geocoder_format };
+            Object.keys(info).forEach(function(key) {
+                if (/^geocoder_format_/.exec(key)) {
+                    source.geocoder_format[key.replace(/^geocoder_format_/, '')] = info[key];
+                }
+            });
```
yhahn: This moves the template `formats` object building into source initialization. Previously we were doing this at runtime when returning every query.

lib/geocode.js
```diff
@@ -381,62 +339,68 @@ function forwardGeocode(geocoder, query, options, callback) {
...
-                stats.time = (+new Date()) - stats.time;
-                queryData.stats = stats;
-            }
-            if (options.debug) queryData.debug = options.debug;
-            return callback(null, queryData);
-        });
+        return callback(null, queryData);
     }
 }
+
```
yhahn: All of the logic for reverse, reverse (typed), and forward feature formatting have been tucked away into this function.

It takes an array of contexts and returns geojson features, recording used indexes along the way. Any of our `raw => final output` logic will ideally end up here.

lib/util/filter.js
```diff
@@ -101,3 +105,14 @@ function featureMatchesTypes(source, feature, options) {
     return false;
 }
 
+
+// For a feature to match a languageMode: strict filter it must:
+// - Have a text field matching the desired language code
+function featureMatchesLanguage(feature, options) {
+    if (!options.language) return true;
+    if (options.languageMode !== 'strict') return true;
```
yhahn: Note that `languageMode: strict` is still conceptually a little fuzzy around the edges -- it treats the language code as the main thing to match, rather than `{languageCode}_{suffix}`. Refining of this behavior in future versions seems likely.

lib/util/ops.js
```diff
@@ -1,16 +1,18 @@
...
  * toFeature - Reformat a context array into a Carmen GeoJSON feature
  *
  * @param {Object} context Array of GeoJSON features ordered by index level
  * @param {Object} format Objects containing default/langauge specific formatting for `place_name`.
  * @param {String} language code to select language/formatting from context array. Undefined defaults to `carmen:text`
+ * @param {String} languageMode option that can be set to 'strict' filter resutlts strictly to a specific language. Undefined defaults null.
  * @param {Boolean} debug Enable debug mode to expose internal properties in results
  * @return {Object} Carmen GeoJSON feature
  */
-function toFeature(context, format, language, debug) {
```
yhahn: This function needs to change to because `languageMode` strict deals with corner cases where there are features in the context array that need to be filtered out if they do not match the target language. Prior to reaching this point the first feature in the context should already have been filtered down to language consistency -- the changes here prune the context for those features.

PR #582, opened 2017-01-25
Optimize phrasematch

springmeyer:
### Context

I profiled the [setup routine](https://github.com/mapbox/carmen/blob/master/bench/geocode.js#L13-L50) and [geocode routine](https://github.com/mapbox/carmen/blob/master/bench/geocode.js#L64-L69) of `bench/geocode.js`.

With the `setup` routine I noticed:

  - The main event loop was quite busy (98% of time sampled it was busy in sync code)
  - The threadpool was idle
  - The primary bottlenecks in the main event loop were: 1) garbage collection, 2) `unidecode`, and 3) v8 string and array manipulation
  - The main bottleneck within `unidecode` was string allocation, deallocation, and `std::string::append`. 

With the `geocode` routine I noticed:

  - The main event loop was fairly busy (77% busy, 23% idle)
  - The threadpool was idle (with the exception of < 1 % in `carmen::cache::coalesce`)
  - The majority of the 77% of the main event loop was bottlenecked in JSDawg usage:
      - 4 % in deallocating `shared_ptr`
      - the remainder in `JSDawg::Insert` (which itself is slow due to the `std::to_string` usage)

Next I worked to optimize both jsdawg (https://github.com/mapbox/dawg-cache/pull/10) and unidecode (https://github.com/mapbox/node-unidecode-cxx/pull/4). These PRs noticeably speed up the benchmarks in those two repos by reducing memory allocations.

However I noticed they did not make a big impact on the overall performance of carmen's `bench/geocode.js`. So next I took a look at how jsdawg and unidecode were being used in carmen. 

I noticed that `unidecode` and `JSDawg::insert` were being called more times than needed. In many places strings are parsed twice with `unidecode` and the results are not shared. And in several cases `WriteCache.hasPhrase` is called (which is noted to be slow: https://github.com/mapbox/carmen/blob/33e8aa008c3140f7a171eea9c1baa32e154d5afb/lib/util/dawg.js#L58-L59). This `WriteCache.hasPhrase` usage explains why `JSDawg::Insert` was blocking the main event loop in the `geocode` case because it internally [dumps and reloads the entire cache](https://github.com/mapbox/carmen/blob/33e8aa008c3140f7a171eea9c1baa32e154d5afb/lib/util/dawg.js#L19-L30).

### Fixes/Adds

  - Avoids calling `unidecode` twice per subquery
  - Avoids dumping and recreating a dawg `ReadCache` per subquery
  - Speeds up `getIndexablePhrases` by avoided extra `unidecode` calls
  - Speeds up `bench/geocode.js` by `~ 6x`

### Next Steps

 - [ ]  Should we remove [this slow WriteCache functions](https://github.com/mapbox/carmen/blob/33e8aa008c3140f7a171eea9c1baa32e154d5afb/lib/util/dawg.js#L57-L66) to avoid them being called in performance critical code?
 - [ ] Code review of my other PRs: https://github.com/mapbox/node-unidecode-cxx/pull/4 and https://github.com/mapbox/dawg-cache/pull/10

### DIFF
--- a/lib/phrasematch.js
+++ b/lib/phrasematch.js
@@ -1,6 +1,7 @@
 var termops = require('./util/termops');
 var token = require('./util/token');
 var bb = require('./util/bbox');
+var dawgcache = require('./util/dawg');
 
 /**
 * phrasematch
@@ -35,6 +36,16 @@ module.exports = function phrasematch(source, query, options, callback) {
     }
 
     subqueries = termops.uniqPermutations(subqueries);
+    var dawg;
+    // if writeCache, dump data once here to avoid
+    // repeated penalty of dump per subquery
+    if (source._dictcache.dump) {
+        // create a ReadCache once
+        dawg = new dawgcache(source._dictcache.dump());
+    } else {
+        // Already is a ReadCache
+        dawg = source._dictcache;
+    }
 
     loadall(getter, 'freq', [1], function(err) {
         if (err) return callback(err);
@@ -51,10 +62,10 @@ module.exports = function phrasematch(source, query, options, callback) {
             var subquery = subqueries[l];
             var text = termops.encodableText(subquery);
             if (text) {
-                if (!source._dictcache.hasPhrase(text, subquery.ender)) continue;
+                if (!dawg.hasPhrase(text, subquery.ender)) continue;
                 // Augment permutations with matched grids,
                 // index position and weight relative to input query.
-                var phrase = termops.encodePhrase(subquery, options.autocomplete ? subquery.ender : false);
+                var phrase = termops.encodePhraseText(text, options.autocomplete ? subquery.ender : false);
                 var weight = subquery.length / tokenized.length;
                 phrasematches.push(new Phrasematch(subquery, weight, subquery.mask, phrase, scorefactor, source.idx, source._geocoder, source.zoom));
             }
--- a/lib/util/termops.js
+++ b/lib/util/termops.js
@@ -16,6 +16,7 @@ module.exports.getHousenumRangeV3 = getHousenumRangeV3;
 module.exports.encodeTerm = encodeTerm;
 module.exports.encodableText = encodableText;
 module.exports.encodePhrase = encodePhrase;
+module.exports.encodePhraseText = encodePhraseText;
 module.exports.tokenize = tokenize;
 module.exports.feature = feature;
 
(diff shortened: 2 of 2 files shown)


apendleton:
@springmeyer on the dawg side, only readcache is used in production at query time. It's just read-only, so for tests and benchmarks where we want to build a structure first, a write-cache gets built first. I added those methods for convenience to make the tests less clunky (you would otherwise have to build a write cache, dump it, load it into a read cache, and monkey-patch carmen to point at the read cache since it has no mechanism to easily swap one for the other), but that code path is not accessible from the production geocoder because the geocoder caches at runtime are always instantiated with a buffer: https://github.com/mapbox/carmen/blob/33e8aa008c3140f7a171eea9c1baa32e154d5afb/lib/util/dawg.js#L43-L49 . If those slow functions are getting run in the benchmark, the benchmark is testing the wrong thing.

Likewise, the dawg-cache PR looks good to me -- I want to do another read through tomorrow but should be good to merge -- but the entirety of `dawg.cpp` is only used at index build time (it's the mutable version of the structure, not the read-only compact one), so this won't make the runtime geocoder faster, just the indexer.

springmeyer:
> If those slow functions are getting run in the benchmark, the benchmark is testing the wrong thing.

@apendleton yes, the slow functions were getting run during the benchmark. This is visible by doing:

```
DEBUG=1 node bench/geocode.js
```

Which after a while starts spewing `calling hasPhrase on a DAWG WriteCache is very inefficient`. What is the fix that is needed for the benchmark?

apendleton:
The benchmark indexes a bunch of stuff and then geocodes a bunch of stuff. In between, the WriteCache that's attached to the carmen instance for indexing should be replaced by a read cache. To do this, you'd dump the WriteCache, accessible as `._dictcache` on the carmen object, to a buffer using `.toCompactDawgBuffer()` and then instantiate a new Dawg (from `lib/util/dawg.js`) passing that buffer in, then replace `._dictcache` on the carmen with the new Dawg.

springmeyer:
Okay. I think I did what you describe in this PR in https://github.com/mapbox/carmen/pull/582/files#diff-8ef177c6c3f320a8a4c9b7799759c82aR39. Does that look reasonable to do in general? (it covers the benchmark case as well as others perhaps?). Looks like @yhahn already merged this but I'm happy to revise in another PR if you have recommendations on a cleaner approach.

springmeyer:
per chat with @apendleton the change that landed in this PR to reload the dawg cache in phrasematch at https://github.com/mapbox/carmen/pull/582/files#diff-8ef177c6c3f320a8a4c9b7799759c82aR39 is incorrect. Working on fixing this in the benchmark scripts instead over at #584.

PR #591, opened 2017-02-04
Remove old legacy feature doc code paths

yhahn:
### Context

Carmen still supports "v0" spec feature docs -- both as input for indexing and for reverse geocoding only. These docs are the non-geojson format docs that looked like this:

```json
{
  "_text": "Springfield",
  "_center": [0,0]
}
```

We've maintained internal conversions between these docs are the newer geojson format carmen supports. This PR removes all that legacy support and conversion code.

cc @mapbox/geocoding-gang


yhahn:
This'll start carmen@19.x

ingalls:
Oh the joy I feel at finally seeing this go cannot even be described.

PR #595, opened 2017-02-10
Scriptsniff

mattficke:
### Context
Follow-up to #593 & 594.

Even after ensuring language consistency with `languageMode: strict`, mixed-script results can still sneak through in cases where a single language has variants written with differing scripts. This adds a final script consistency check.

### Summary
* add [scriptsniff](https://github.com/mapbox/scriptsniff)
* use scriptsniff to test each feature's `place_name` for mixed script results
* don't include these results when using `languageMode: strict`

### Next Steps
This is a little blunt, and removes any mixed-script results entirely as a final fail-safe. A more sophisticated approach would be to detect the desired script based on the language parameter, and filter out individual features from the context array that don't match the intended script.

### DIFF
--- a/lib/geocode.js
+++ b/lib/geocode.js
@@ -390,6 +390,7 @@ function forwardGeocode(geocoder, query, options, callback) {
 function toFeatures(geocoder, contexts, options) {
     var features = [];
     var indexes = {};
+    var feature;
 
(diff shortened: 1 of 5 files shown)


mattficke:
Going with an alternate approach in #596

PR #596, opened 2017-02-10
exclude loose language matches known to involve mixed scripts

mattficke:
### Context
Follow-up to #593 & #594. Excludes languages with subtypes that use varying scripts from falling back to a potentially mixed-script result.

### Summary
* Add a `digraphic` array of languages known to use multiple scripts
* For these languages, don't attempt a loose match in strict languageMode.


yhahn:
@mattficke PR looks good -- I think :green_apple: once testing IRL/integration looks good?

PR #600, opened 2017-02-23
reverseMode

mattficke:
### Context
Reverse queries currently order features solely by distance. This adds the option to consider score when sorting as well.

### Summary
* adds a `distScore` calculation, similar to `scoreDist` for proximity. Sorts features found with reverse queries based on score + distance, weighted exponentially between 50m - 1000m.

### Next Steps
* need to do a few more performance optimization sweeps. Should probably add this code path to a benchmark test as well.
* Annotate with more comments


index.js
```diff
@@ -102,6 +102,7 @@ function Geocoder(indexes, options) {
             source.geocoder_tokens = info.geocoder_tokens||{};
             source.geocoder_inherit_score = info.geocoder_inherit_score || false;
             source.geocoder_universal_text = info.geocoder_universal_text || false;
+            source.geocoder_reverse_mode = info.geocoder_reverse_mode || false;
```
mattficke: `geocoder_reverse_mode` is enabled on a per-index basis as a boolean tilejson property

lib/context.js
```diff
@@ -99,6 +100,7 @@ function stackFeatures(geocoder, loaded, options) {
     var context = [];
     var memo = {};
     var firstType = false;
+    var reverseMode = options.reverseMode || 'distance';
```
mattficke: `reverseMode=distance` maintains the previous default behavior, this sets it as the default

lib/context.js
```diff
@@ -140,7 +142,12 @@ function stackFeatures(geocoder, loaded, options) {
                 // Reconstruct extid based on selected type
                 feature.properties['carmen:extid'] = type + '.' + feature.properties['carmen:extid'].split('.').pop();
                 break;
-            } else if (memo[type] && loaded[i].properties['carmen:geomtype'] !== 3) {
+            } else if (memo[type] && feature.properties['carmen:geomtype'] !== 3) {
+                if (reverseMode === 'score') {
+                    // don't replace a higher-scored feature with a lower-scored one
+                    if (!feature.properties['carmen:score'] && memo[type].properties['carmen:score']) continue;
+                    if (feature.properties['carmen:score'] && memo[type].properties['carmen:score'] && memo[type].properties['carmen:score'] >= feature.properties['carmen:score']) continue;
+                }
```
mattficke: Previously, less-distant features from lower-numbered indexes would displace features from higher-numbered indexes if they shared the same `geocoder_name` (such as addresses and POIs). This adds the option to only do so if the score is higher.

lib/context.js
```diff
@@ -356,6 +363,21 @@ function contextVector(source, lon, lat, full, matched, language, scoreFilter, c
...
+                    return { index: i, distscore: proximity.distscore(el.distance, el.attributes()['carmen:score'])}
+                });
+
+                mapped.sort(sortByDistScore)
+                resultsSorted = mapped.map(function(el) {
+                    return results[el.index];
+                });
+                results = resultsSorted;
+                resultsSorted = true;
+            }
```
mattficke: Results are sorted by `distScore` if `reverseMode` is set to `score`, it's a reverse geocode, and the index allows reverseMode sorting.

lib/util/proximity.js
```diff
@@ -79,3 +80,14 @@ function scoredist(meanScore, dist, zoom) {
...
+/**
+ * Similar to `scoredist`, used for including score + distance when sorting features returned by a reverse query.
+ * 
+ * @param {Number} dist The distance in meters from the feature to the query point
+ * @param {Number} score The feature's score
+ * @return {Number} distance-adjusted score value
+ */
+function distscore(dist, score) {
+    return Math.round(score * (1000/(Math.max(dist,50))) * 10000) / 10000;
+}
```
mattficke: Weight feature score based on distance from query point. Closer features are weighted more, with a 50m buffer to cap the weighting at 20x.

mattficke:
I think something's up with the github/circle integration, test is 🍏 but still says it's in progress here. Merging.

https://circleci.com/gh/mapbox/carmen/1906

Issue #547, opened 2016-11-30
Multilingual grid proposal


apendleton:
Currently, the grid cache is not language-aware. Doing multilingual autocomplete will likely require us to change that. Here's a proposal to do that. Aims:
* get rid of the current restriction that autocomplete only applies to default text (and remove the dot-suffix)
* make grid operations, coalesce, etc., language-aware, and specifically:
    * prefer matches in the user's requested language
    * still allow matches in a language other than the requested one, but with a relevance penalty
* be minimally invasive (most changes required here will be in `carmen-cache`)

The current grid cache structure works conceptually like:
```YAML
phrase1: [grid1, grid2, grid3, grid4]
phrase2: [grid5, grid6, grid7, grid8]
```

The proposed new format marks grids for a given phrase with a language bitmask:
```YAML
phrase1:
    lang_mask1: [grid1, grid2, grid3]
    lang_mask2: [grid4]
phrase2:
    lang_mask3: [grid5]
    lang_mask4: [grid6, grid7]
    lang_mask5: [grid4]
```

The language bitmask would represent, per index, the list of all known languages at index time. For example (assuming a 64-bit int for the mask, but can/should be bigger):
```
0000000000000000000000000000000000000000000000000000000000000000
|||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||↳ default
||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||└→ ar
|||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||└-→ az
||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||└--→ bg
|||||||||||||||||||||||||||||||||||||||||||||||||||||||||||└---→ bn
||||||||||||||||||||||||||||||||||||||||||||||||||||||||||└----→ bs
|||||||||||||||||||||||||||||||||||||||||||||||||||||||||└-----→ ca
||||||||||||||||||||||||||||||||||||||||||||||||||||||||└------→ cs
|||||||||||||||||||||||||||||||||||||||||||||||||||||||└-------→ da
||||||||||||||||||||||||||||||||||||||||||||||||||||||└--------→ de
|||||||||||||||||||||||||||||||||||||||||||||||||||||└---------→ el
||||||||||||||||||||||||||||||||||||||||||||||||||||└----------→ en
|||||||||||||||||||||||||||||||||||||||||||||||||||└-----------→ es
[...]

PR #602, opened 2017-03-02
[wip] Hyperpolyglot (multilingual autocomplete, etc.)

apendleton:
This is the working branch for the implementation of the carmen and carmen-cache changes described [here](https://github.com/mapbox/carmen/issues/547#issuecomment-277320959), with some slight augmentation to be discussed in a future carmen-cache PR of the `lang-fields` branch.

## Summary of changes:
* As per the previous ticket, carmen-cache now stores language information as well as text and grids. Specifically, there is now one grid cache entry per phrase per combination of languages, rather than one entry per phrase. Grids pertain to that phrase for those languages only.
* Languages are stored one language per bit in a 128-bit integer. The mapping of language to bit position is per-index, so not all indexes have to agree on the language list (we can add languages to individual indexes without globally rebuilding), but the list of allowed languages must now be included in each index's tilejson, and text in language's not in the tilejson's language list will not be indexed.
* Most use of carmen-cache's `get` has been replaced by `getMatching`, which is the successor to `getByPrefix` from the `rocksdb` work. It can either match prefixes or full strings, and also optionally takes a list of language IDs which will be converted into a bitmask in the same format as above, which will be bitwise-AND-ed with the masks of keys matching either the prefix or complete string, respectively, of the cache contents. All entries can potentially be returned regardless of language match, but coalesce will apply a relevance penalty (currently 20%) if the feature does not match a requested language. If no language is supplied to coalesce, an all-1 language mask (so, matches everything with no penalty) will be used.
* at index time, docs that have only `carmen:text` and no language features will also be indexed with an all-1 language field, so they will match any language with no penalty. This will work either for features in untranslated layers (e.g., address layers), or for features with no translations in otherwise-translated layers (because they have not been match to an external source of translation data, etc.).
* Big picture: at query time users who specify a language will see, for both autocomplete and non-autocomplete queries and both complete and degen matches, results matching text from all languages, but text that matches in the language the user requested will be given higher precedence

## Things to have done for the first pass, ideally tomorrow:
* Can index multilingual text and query against it as described above
* Tests of this functionality

## Things to be addressed in the second pass (aiming for Friday):
* Language fallback behavior: ideally if you had text for Spanish but not Portuguese for a feature and, and Spanish was the language we would fall back to when Portuguese was requested, we wouldn't penalize the Spanish text for a Portuguese query. We should emphasize penalizing for a given language if we have known in-language text for a feature and what the user is querying isn't it, but not in situations where we have no text for the language. Put another way: we should aim for the text we're actually going to display to the user (including fallback) matching the user's query as closely as possible. This means we should apply the fallback behavior anticipatorily at index time, and ideally that we assign phrases to all languages in the index for every feature. This shouldn't result in any additional data in the grid cache, but should just involve, e.g., flipping the Portuguese bit on the phrases for the Spanish text on a feature to 1 if we don't have Portuguese text for that feature.
* Autocomplete text similarity: while being able to search for "Deutschland" and get back results even with `language=en` is better than returning nothing (and is a big part of the motivation for this change), we do create the potential for situations where the text we're returning in results doesn't match what the user typed in, as for `language=en` on that feature we'd select "Germany" for display as it's the English text for that feature. We also eliminate any notion of not indexing text for autocomplete, so we regress on [this fix](https://github.com/mapbox/carmen/pull/569) if an in-language match is a synonym rather than the primary text for a feature. I think the fix for this is, in situations where the matched text differs from the display text, to add additional properties to the response with the matched text rather than the ideal-language text so that API consumers can choose which to show to their users as appropriate. `text`, `place_name`, and `language` will still represent a best effort to select display text that best matches what the user requested for the features matched, and where necessary, something like `matched_text`, `matched_place_name`, and `matched_language` will be added if there's disagreement.
* Proper treatment where the user has requested a language the index doesn't include: as currently implemented, if a user requests a language a given index doesn't include at all, the default text is searched instead. We should instead collapse to the nearest language an index includes and use that.

## Open question
As implemented so far, `carmen:text` is treated as just another language (internally referred to as `default`) similar to any other language, and queries without a language specified are searches as if the language were specified as default. This means text from languages other than the default is included but penalized in results, for both autocomplete and non-autocomplete. In current production, for full phrases we search all languages equally, and for partial/degen phrases we search only default text. This is not great behavior in my opinion, but the new behavior is still a fairly significant departure. In the new architecture, we do have the ability to search all phrases with no penalty (by using an all-1 language mask) for non-prefix phrases so we could get part of the way towards matching the current behavior if we wanted to, but the second part would essentially require a behavior of "don't penalize these results if **either** the phrase is a full phrase **or** the language field includes `default`."

I think achieving that at the carmen-cache level would be really gross, so if we wanted to get close to the current behavior we'd probably want, for queries that didn't include a language field, to consider two coalesce stacks for each potential match, one where the final subquery is configured as `{ languages: ALL_LANGUAGES, prefix: false }` and one configured as `{ languages: [ "default" ], prefix: true }`. No idea what the performance ramifications would be, though, and the results still wouldn't be exactly the same (we would still include out-of-language autocomplete matches, they would just be of lower relevance than the features we return now). Open to more suggestions, ideas, etc., on this one.

cc @yhahn


apendleton:
The latest on this:
* everything in the ticket that was proposed, I did, including precomputation of fallback text at index time, and `matched_text`, etc. The implementation for `matched_text` could be done more cleanly in the future, but this will work for now.
* the open question, above, has been resolved with continuing to treat default text as its own language. This works very well for all circumstances that don't involve indexes that lack translations entirely (for which all text is indexed as all-language). For queries that involve both translated and translation-less indexes, translation-less indexes currently might get preferential treatment in that they never receive language penalties while translated indexes might; this means in practice that searching for non-default text without using the language flag will tend to favor translationless indexes. **We've decided we're okay with this for the time being.** A subsequent, runtime-only change could involve applying a relevance penalty to translationless indexes in this circumstance.
* Also, big additional change: **unidecode-cxx is dead**, and has been replaced by a new all-Javascript diacritical mark folder that now lives in [lib/util/remove-diacritics.js](https://github.com/mapbox/carmen/blob/0962df005edb9b35412918dbf4043ff9e03f3322/lib/util/remove-diacritics.js).

Issue #603, opened 2017-03-02
Change fallback behavior for Arabic


apendleton:
### Context
If a user requests text in a language that we don't have for the feature they requested, we attempt to guess what language we should return instead, in the [closest-lang.js](https://github.com/mapbox/carmen/blob/master/lib/util/closest-lang.js) util file. There are a bunch of steps here. First we consult a hand-curated list of fallback candidates (so like, if we know Portuguese should fall back to Spanish we hard-code that), and we try some variants of that if that fails (so like, if you requested 'en-IN` and we don't have that in the fallback list but we do have plain `en`, we'll try that).

If that fails, we try a bunch of heuristics that examine all the languages we do have and assess their similarity to the requested one on criteria like script (so like, if you request Ukrainian, prefer other Cyrillic languages like Russian before falling back to, say, English).

At present, we have no hard-coded fallbacks in place for Arabic, so we rely on the heuristics, which tend to fall back to other Arabic-script languages like Urdu or Farsi. Customer feedback has indicated that this is not desired behavior and that despite script overlap, Arabic speakers are more likely to want English as a fallback option than Farsi. We can accomplish this via hardcoded fallback.

### Next Steps
* [ ] add an Arabic entry to `fallback.json` that falls back from Arabic to English
* [ ] test that Arabic search for features with English but no Arabic text returns the English result
* [ ] add test to pin down what happens if a user searches for Arabic and gets a feature that has neither Arabic nor English text. This will probably still use the heuristics, so I would imagine it will still fall back to Farsi, etc. I think this is okay, but we should probably formalize it via a test anyway.

cc @aarthykc @yhahn

PR #604, opened 2017-03-02
Arabic eng fallback

aarthykc:
### Context
While requesting a feature when `language=ar`,we return the results in Urdu or Farsi, this adds the Arabic fallback language to english and tests that cover cases added here - #603 

### Fixes/Adds
Fixes the language fallback for `arabic => english`

### Summary
- [x] Changes in `test/fallback.js` to add ar => en
- [x] Geocode unit tests to check behavior when there is no english text and check if the fallback behavior works
- [x] add Arabic fallback tests in `test/closest-lang.js`

### Next Steps
-  [x]  Merge @apendleton, @yhahn ?


test/geocode-unit.languageFallback.js
```diff
@@ -0,0 +1,78 @@
...
+            q.defer(function(cb) {
+                buildQueued(conf[c], cb);
+            });
+        });
+        q.awaitAll(t.end);
+    });
+
+    tape('query: United States', function(assert) {
+        c.geocode('United States', { language: 'ar'}, function(err, res) {
+            assert.equal('United States', res.features[0].text, 'Fallback to English');
```
yhahn: @aarthykc it would be great to add an assertion for the `language` property here as well to prove that it wasn't the `carmen:text` default version of `United States` that was returned.

aarthykc: Thank you for reviewing @yhahn, I've added the assertion !

apendleton:
Looks good to me

aarthykc:
@yhahn, @apendleton good to merge ?

PR #606, opened 2017-03-03
Exclude addresses from solely numeric queries

mattficke:
### Context
Numeric queries used to only match a feature if the value was present in the `carmen:text` property. This meant a query for `12345` would match the postcode 12345, but not 12345 Main Street.

This changed during the rocksdb refactor, but the old behavior led to better results. This adds similar behavior back in.

### Summary
* use termops.getPhraseDegens to parse queries for address number component
* don't match address features when only the address number component matches


apendleton:
getPhraseDegens kind of seems like a needlessly expensive operation for this purpose (like, could generate an array with dozens of strings in it that we're just going to immediately throw away), and also I would kind of like to delete it since we don't index degens anymore and its presence is kind of confusing (as evidenced by our befuddlement earlier). Is `termops.getPhraseDegens(text).length === 0` functionally any different from `/#/.test(text.split(' ')[0]) && !(/ /.test(text))` for what we're doing with it? We could keep it in termops if we wanted (like, `phraseHasDegens` or something) and just not actually do the iterating or build the array.

If we do keep it the way it is now, at a bare minimum we should swap so `source.geocoder_address` gets checked first, so we won't generate the phrase degens on every ender for every query, just the ones where we're doing a lookup into an address index.

mattficke:
@apendleton that's a good point. For testing purposes I think I'd prefer to keep it as a standalone termops function, I'll tighten it up there.

PR #607, opened 2017-03-03
Pt v. ITP Limits

ingalls:
### Context

We prefer to return a building point for a feature that has both Points and LineStrings

Sometimes the closest point in a feature is far away from the ITP segment it originally matched. Add sensible limits to how far a pt can be returned.


test/geocode-unit.address-misc.test.js
```diff
@@ -0,0 +1,78 @@
...
+ * (. is pt addr. -- is itp addr)
+ *
+ *      . . . . . . . .
+ *      ------------------------------------
+ *                                  x
+ * We want to prioritze returning a '.' where it is sensible
+ * But in this case it's pretty far away so we should return --
+ * even though the feature has '.'s
+*/
+
```
KaiBot3000: Love the ascii 👏

PR #609, opened 2017-03-21
support multiple values for ?language=en,es,fr,...

yhahn:
### Context

This PR enables passing multiple language codes to the `language` option separated by commas, allowing multiple language results to be returned with a single geocoding request. When multiple langauges are specified

- for forward geocodes, the first language is used for matching resulting features,
- for reverse geocodes, no changes are made to how resulting features are retrieved,
- for all geocodes, the resulting feature geojson includes the following additional properties per language:
  - `text_{code}`: The feature text, localized to `code`
  - `place_name_{code}`: The full place text, localized to `code`
  - `language_{code}`: The language code used for text and place_name. Usually the value of `code` but may differ if a fallback language was used.

### Next Steps

- [x] Add a hard cap on the number of languages that can be requested
- [x] Add tests to demonstrate first-language use on forward geocodes
- [x] Add docs on this feature
- [x] Test for cli support
- [x] Code review


lib/util/ops.js
```diff
@@ -4,124 +4,158 @@ var featureMatchesLanguage = require('./filter').featureMatchesLanguage;
...
+        var formatString = getFormatString(context, format, language);
+        memo[`text${suffix}`] = text.text;
+        if (text.language) memo[`language${suffix}`] = text.language.replace("_", "-");
+        memo[`place_name${suffix}`] = getPlaceName(context, formatString, language, languageMode);
+
+        if (i === 0) {
+            memo.text = memo[`text${suffix}`];
+            if (text.language) memo.language = memo[`language${suffix}`];
+            memo.place_name = memo[`place_name${suffix}`];
+
```
yhahn: @apendleton would love a gutcheck on how `matching_place_name` is set here. I ended up refactoring the code flow quite a bit but _think_ I got the original intent...

apendleton: What I was aiming for was for `matching_place_name` to use the same template as we used for `place_name`, but interpolate `matching_text` instead of `text` into the template for each entry in the context that had it, or `text` otherwise.

I'm not wedded to the current behavior, though, admittedly. For one thing, we currently have a rule where we only use the format string for a particular language if at least one context entry has text in that language, but it's possible with `matching_place_name` for us to use a format string even though none of the `matching_text` strings are in the requested language, which is maybe weird? Also `matching_text` in my implementation totally ignores `languageMode`, which seemed right at the time (you want to include all the text that's similar to what the user typed, even if it's a language mismatch; that's kind of the point), but I could easily be talked out of that.

I think anything that gets at approximately "`place_name`, except where alternate text for a particular feature is a better fit for the user's submitted text" would be fine with me.

yhahn:
### Ready for review!

Suggested review approach:

1. Check out the doc update to the `language` option that define the expected behavior
2. Check out the carmen geojson update for expected output format
3. Check out the new tests that define expected additional output when multiple languages are specified
4. Check out the implementation

The biggest code change is a heavy refactor of `lib/util/ops.js` that splits `getPlaceName()` out of the main `toFeature()` code flow so that it can be called multiple times easily for each different language that is requested.

lib/util/ops.js
```diff
@@ -4,124 +4,158 @@ var featureMatchesLanguage = require('./filter').featureMatchesLanguage;
...
+ * @param {String} language code to select language/formatting from context array. Undefined defaults to `carmen:text`
+ * @param {String} languageMode option that can be set to 'strict' filter results strictly to a specific language. Undefined defaults null.
+ * @param {String} pass Set to 'matching' to enable autoc synonym matching fallback
+ * @return {String} formatted place_name string
+ */
+function getPlaceName(context, formatString, language, languageMode, pass) {
+    var feat = context[0];
+    var place_name;
+    // Use geocoder_format to format output if applicable
+    if (!formatString || formatString === true || formatString === 1) {
```
sbma44: out of curiosity, what circumstances set formatString to `true` or `1`?

lib/util/ops.js
```diff
@@ -4,124 +4,158 @@ var featureMatchesLanguage = require('./filter').featureMatchesLanguage;
...
+}
+
+/**
+ * getPlaceName - get a formatted `place_name` string for a given context stack & language
+ *
+ * @param {Object} context Array of GeoJSON features ordered by index level
+ * @param {String} formatString Templating string for formatting place_name output
+ * @param {String} language code to select language/formatting from context array. Undefined defaults to `carmen:text`
+ * @param {String} languageMode option that can be set to 'strict' filter results strictly to a specific language. Undefined defaults null.
+ * @param {String} pass Set to 'matching' to enable autoc synonym matching fallback
```
sbma44: can you expand on this? from the code I *think* this is in place to preserve tokens specified in the input query regardless of language. but that would make the use of `synonym` here imprecise. and the fact that the param is only used in the absence of a `formatString` makes me think I must be misunderstanding something.

lib/util/ops.js
```diff
@@ -35,18 +35,18 @@ function getFormatString(context, format, language) {
  * @param {String} formatString Templating string for formatting place_name output
  * @param {String} language code to select language/formatting from context array. Undefined defaults to `carmen:text`
  * @param {String} languageMode option that can be set to 'strict' filter results strictly to a specific language. Undefined defaults null.
- * @param {String} pass Set to 'matching' to enable autoc synonym matching fallback
+ * @param {Boolean} matched If set, use matched text in place name
```
sbma44: very helpful!

brsbl:
@yhahn what's the status of this work?

yhahn:
@brsbl will be getting this integrated into Mapbox Geocoding API next!

PR #616, opened 2017-04-10
Pin all them turf modules to 4.0.0 (for now)

yhahn:
### Context

@ingalls's node v6 upgrade push was :green_apple: and then some deps shifted under our feet, throwing off some geometry coordinate values in our fixtures. After comparing `npm ls` of the passing tests (thx to local tests from @ingalls) it looks like the difference is in our turf modules.

Based on my testing locally so far:

- :white_check_mark: 4.0.0
- :x: 4.0.1
- :x: 4.1.0

This PR stopgap fixes our master tests by pinning to `4.0.0` for now. I'll look into what changed upstream with the intention of getting onto the latest turf instead -- probably the case that we'll want the bugfix in the latest turf versions.

cc @mapbox/geocoding-gang


ingalls:
This looks great to me!

PR #618, opened 2017-04-14
Robustify matching text handling when geocoder_format is set

yhahn:
### Context

Desired behavior:

- Include `matching_place_name` on a feature if any part of the query matches part of the result feature's context and adjusts the text to be displayed based on what the user typed (e.g. matching a synonym for part of the context)
  - Example query/matching: **Kansas, Am**erica
  - Canonical placename: Kansas, United States
- Include `matching_text` property on a resulting feature if the feature itself is where the canonical/matching discrepancy lies:
  - Example query/matching: **Ameri**ca
  - Canonical text: United States

This PR fixes up a few cases where this desired behavior ^^ was not being captured fully if 
 `geocoder_format` was set.

cc @mapbox/geocoding-gang

### DIFF
--- a/lib/util/ops.js
+++ b/lib/util/ops.js
@@ -74,7 +74,7 @@ function getPlaceName(context, formatString, language, languageMode, matched) {
                 if (templateProp == carmenProp) {
                     if (templateSubprop == '_name') {
                         // `name` is a hardcoded subproperty that maps to carmen:text
-                        val = closestLang.getText(language, f.properties).text;
+                        val = matched && f.matching_text ? f.matching_text : closestLang.getText(language, f.properties).text;
                         formatString = formatString.replace('{' + templateProp + '._name}',val);
                     } else if (templateSubprop == '_number') {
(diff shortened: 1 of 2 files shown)


lib/util/ops.js
```diff
@@ -144,12 +144,13 @@ function toFeature(context, format, languages, languageMode, debug, geocoder) {
...
 
             // Get matching place name if no language matching occurred.
-            if (!formatString && geocoder) {
+            if (geocoder) {
                 for (var k = 0; k < context.length; k++) {
                     var matched = !!context[k].properties['carmen:query_text'] && getMatchingText(context[k], geocoder, language);
                     if (matched) {
                         context[k].matching_text = matched.matching_text;
                         context[k].matching_language = matched.matching_language;
+                        if (k === 0) feature.matching_text = matched.matching_text;
```
yhahn: This check includes the `matching_text` value on a feature only if the `matching_text` belongs to the feature and not the broader context.

PR #622, opened 2017-05-04
Cleantest

ingalls:
### Context

Update tests with some ES6 Syntax for better readability.

- [x] Standardize test/tape variables across all tests
- [x] Switch to using arrow functions where applicable - greatly helps with `mem()` readability
- [x] Use const/let where possible


KaiBot3000:
I've never written an arrow function, and only barely know the difference between const and let. I'm throwing that out there to flag that `readability` is subjective. I'm down to update, and we should use this opportunity to develop a style-guide moving forward, so that we don't end up with an inconsistent code base. By keeping our code in a similar style we'll make it easy for all of us to switch between different parts of our codebase without extra mental effort or confusion.

PR #625, opened 2017-05-12
collapsing across variants of arabic YEH

boblannon:
### Context

The arabic character "YEH" varies slightly in the Persian writing system, which causes some queries to miss.

| query for "tabriz" | unicode for "YEH"                | matches?   |
|:-------------------|:---------------------------------|:-----------|
| تبريز              | `U+064A ARABIC LETTER YEH`       | ✅          |
| تبریز              | `U+06CC ARABIC LETTER FARSI YEH` | ❌        |‎ |

### Fixes/Adds

New pattern replacements to `removeDiacritics` that collapse three variants of this character into the same character.


boblannon:
tested this solution on staging.  looking good: both queries above return identical results.

PR #626, opened 2017-05-17
Adding Universal Text

KaiBot3000:
### Context
Some text applies across languages: airport codes, postcodes, and to a lesser degree country abbreviations. These values shouldn't be penalized in queries specifying a language, since they can apply in multiple languages. Let's create a `carmen:text_universal` field to hold this text.

### Summary
- [x] failing unit tests for features with `text_universal`
- [x] [hasLanguage](https://github.com/mapbox/carmen/blob/master/lib/util/closest-lang.js#L65) won't accept `universal` as a language option, so we'll need to update it or handle `universal` text separately.
- [ ] ?
- [ ] update docs on carmen geojson


boblannon:
There is a reserved code for "No Linguistic Content" (`zxx`), which would match an entry in model-un.  Not sure if we want to abuse it for this purpose, though...

apendleton:
There's a bunch of special-casing around things that start with `text`, so `text_universal` is probably better than `universal`.

And yeah, special handling is probably best. I would not use the language code for it... I don't think we should support `?language=zxx` as a query parameter, for example.

KaiBot3000:
@apendleton did you see that [I'm `text_universal` already](https://github.com/mapbox/carmen/pull/626/files#diff-bf3c8d2a586c07f633f93ea3e481823aR22)? I have it commented out now so I can check the behavior of the tests without carmen failing on `hasLanguage`. Wasn't sure from your comment whether you were saying that's what I **should** use and didn't realize I was **already** doing that, or were suggesting I use something else.

apendleton:
Yep, my mistake. I saw both "universal" and "text_universal" in the tests and misinterpreted what you were going for here.

yhahn:
@KaiBot3000 @apendleton just making note here of the magic :sparkles: :tophat: :sparkles: moment in the code @apendleton pointed out where some text can be indexed with the special `all` condition that we'll probably want to apply for `text_universal`.

https://github.com/mapbox/carmen/blob/master/lib/util/termops.js#L264-L271

KaiBot3000:
- [x] added `geocode-unit.language-universal.test.js` with universal text features
- [x] added tests to `termops.getIndexableText.js` around universal text indexing
- [x] added a check to `hasLanguage` to handle universal text
- [x] changed `getIndexableText` to account for universal text

We may want to:
- [ ] remove the `geocoder_universal_text` feature we're testing for in `geocode-unit.languageMode-universal.test.js`. This is for layers that are entirely universal text, for example `postcodes`. Now we have a more flexible option.

yhahn:
@KaiBot3000 sounds good re removing the index-level `geocoder_universal_text` option. I think we'll want to keep the existing test which queries using the `languageMode: strict` flag but adapt it to leverage your shiny new feature?

lib/util/termops.js
```diff
@@ -254,27 +254,43 @@ function getIndexableText(replacer, globalTokens, doc) {
...
+        getTokens(texts, langTexts.size ? 'default' : 'all');
+    }
+
+    // get tokens for each of the non-universal, non-default languages' text
+    for (var langText of langTexts) {
+        var lang = langText[1];
+        if (lang != 'universal') {
+            getTokens(doc.properties[langText[0]].split(','), lang);
+        }
+    }
```
yhahn: @KaiBot3000 small thing in terms of code flow -- I wonder if you want to move the universal text `getTokens` call down here so that the code reads in two major chunks:

1. Handling of `text`
2. Handling of all `text_{language}` fields (still with logic around universal as an exceptional case)

Right now where the `text_universal` value is acted upon is a little unexpected (in the default text section, and then skipped in the other text fields section loop).

test/geocode-unit.language-universal.test.js
```diff
@@ -0,0 +1,90 @@
...
+const context = require('../lib/context');
+const mem = require('../lib/api-mem');
+const queue = require('d3-queue').queue;
+const addFeature = require('../lib/util/addfeature'),
+    queueFeature = addFeature.queueFeature,
+    buildQueued = addFeature.buildQueued;
+
+const conf = {
+    // TODO: possibly handle `universal` differently, so we don't have to think of it as a language
+    region: new mem({ maxzoom: 6, geocoder_name: 'poi', geocoder_languages: ['en', 'universal'] }, () => {})
```
yhahn: Is `universal` no longer needed here now that it's always indexed under `all`? :thinking:

lib/util/termops.js
apendleton: @yhahn I think the reason for this ordering was to be able to make a different decision as to how to handle the default text based on whether or not the universal text is present -- the default text gets indexed for all languages if either there are no `text_` fields, or if there's one and it's `universal`.

KaiBot3000: 👍 I was trying to find a way to only check for universal text once, but reworking it as you described makes sense.

test/geocode-unit.language-universal.test.js
KaiBot3000: removed

KaiBot3000:
Per chat we're going to leave `geocoder_universal_text` for now.

- [x] added a test for `languageMode: strict` when used with universal text
- [x] moved `universal` check to `hasLanguage` rather than relying on remembering to check it before calling that function. That fixed a bug where universal text wasn't making it through `termops.decollide`.
- [x] clarified the code flow in `getIndexableText`

PR #627, opened 2017-05-19
Add icelandic language test

ingalls:
### Context

- Experimenting with Icelandic Unicode address names, was having trouble getting a feature indexed and then geocoded again.
- Adds a ITP/PT combined address feature
- Add support for ITP/PT combined feature in test indexer


ingalls:
cc/ @apendleton Not an indexer problem - tests prove it comes out the other end - disregard previous surmising.

PR #630, opened 2017-05-23
[ready] Tokens of my esteem

apendleton:
### Context
This is an attempt at fixing the `main stre` autocomplete issue, where partial matches on phrases that, when present in the entirety, are collapsed by token replacement, fail to match (in other words, because "main street" becomes "main st" at index time, but "main stre" fails to trigger token replacement and also isn't a valid prefix for "main st," we can't match the correct feature.

The strategy here is to try, when making token replacements at index time, to index both the expanded and contracted versions of each token. Query-time replacement remains the same, so when to-be-contracted phrases are present in their entirety, they should still be contracted as per usual, but should now also match the indexed alternate forms when only partially present in autocomplete queries.

### Summary of Changes
* a new variant of `replaceToken` called `enumerateTokenReplacements` has been added. Rather than returning a new string with tokens replaced, it returns an array whose elements are a combination of strings and arrays of strings; in the latter case, these are token replacement opportunities where any option in the array is a possible option in that slot.
* `createReplacer` has been augmented with an indexer-specific mode that still replaces from long to short as per usual, but also creates backwards replacers that go from short to long, so we can index both regardless of which is included in the input data. Reverse replacers by default are only added if they're unambiguous (so if we had both "street" => "st" and "saint" => "st", we would not create an "st" => X replacer), and only if the replacer is simple -- no parentheses are other complex regex features. We also skip global replacers for now. This alternate replacer is only used at index time, not query time, so each carmen now has two replacer objects, one of each kind.
* getIndexablePhrases now uses the enumerate method and the index-mode replacers, and calculates the cartesian product of all possible token substitution options and enumerates the possible alternate strings for each text. It only attempts this full cartesian product if there are three or fewer token replacement opportunities, and only considers up to eight possible options; the fully-contracted version is guaranteed to be indexed regardless.

### Summary of more changes (as of June 16)
* in addition to the automatic index-time unambiguous reverse token replacement, I've now also added a mechanism to manually supply custom reverse token replacements for cases where ambiguity exists. These can be simple strings as is the norm for token replacements, but they can also be functions, which have access to the entire string on which the replacement is being performed, and can intelligently decide what replacement to perform based on context (for example, using position in the string or other cues to decide whether "st" should be reverse-replaced to "saint" or "street"). `enumerateTokenReplacements` has been adapted to accommodate these custom functions, and tests have been added with a simple example. Note that this is all still just index time, and these functions are not used at query time at all.
* I killed decollide. It was a relic of when we were still dependent on hashes and had hash collision troubles, and no longer works in a subset of "main stre"-esque circumstances.
* I incorporated a fix to the way word boundaries are expressed in non-global token replacement which was causing diacritically marked characters to be treated as word boundaries, leading to incorrect application of replacements in some circumstances (see https://github.com/mapbox/carmen/pull/425 )
* I fixed an additional bug in that non-global token replacement strings are not properly escaped, leading to some token replacement strings containing periods (such as https://github.com/mapbox/geocoder-abbreviations/blob/master/tokens/it.json#L61 ) overmatching index-time strings. This was not really a problem before as the behavior, while broken, was the same at index time as at query time. Now, however, the index-time replacement mechanism no longer supports cascading replacement (that is, the same substring being affected by multiple replacements in sequence), so a replacement set that included both "Saint" => "St" and "S." => "S" would behave differently at index time than query time and indexes containing relevant features and those replacements could not be searched correctly. Periods in source token strings are now escaped to compensate.

cc @mapbox/geocoding-gang


sbma44:
> createReplacer has been augmented with an indexer-specific mode that still replaces from long to short as per usual, but also creates backwards replacers that go from short to long, so we can index both regardless of which is included in the input data. Reverse replacers are only added if they're ambiguous (so if we had both "street" => "st" and "saint" => "st", we would not create an "st" => X replacer), and only if the replacer is simple -- no parentheses are other complex regex features. We also skip global replacers for now. This alternate replacer is only used at index time, not query time, so each carmen now has two replacer objects, one of each kind.

I don't want to throw up any roadblocks, but this does take me back to some strategies that were present in early revs of OpenAddresses and which had to ultimately be abandoned. Count this as a "be careful" rather than a "stop!"...

apendleton:
I updated the main ticket to add a bunch more changes added in the second pass of this work

index.js
```diff
@@ -103,10 +103,12 @@ function Geocoder(indexes, options) {
             source.geocoder_address_order = info.geocoder_address_order || 'ascending'; // get expected address order from index-level setting
             source.geocoder_layer = (info.geocoder_layer||'').split('.').shift();
             source.geocoder_tokens = info.geocoder_tokens||{};
+            source.geocoder_reverse_tokens = options.geocoder_reverse_tokens||{};
```
yhahn: Could this options be renamed? `reverse` here means something like "expansion" (?) but I wonder whether readers may confuse this as an option related to reverse geocoding.

PR #642, opened 2017-07-14
Fix behavior of carmen-index.js if it receives a global tokens in the form of a JS file

apendleton:
### Context
In some circumstances it might be desirable to have the global tokens file passed into the carmen-index.js be a Javascript file that can dynamically construct a set of token replacements, rather than a simple JSON file. In the event that such a file is used, it needs to be properly run and the resulting token object retrieved.

### Summary of Changes
- [x] add check in carmen-index.js for if the result of requiring the global tokens file is a function or an object, and if a function, run it to get the object
- [x] add check in the carmen constructor to make sure the tokens it gets are in the form of an object
- [x] add a bin test that runs carmen-index.js with a function-containing tokens JS file, to make sure it clears the above hurdles

cc @mapbox/geocoding-gang

### DIFF
--- a/index.js
+++ b/index.js
@@ -25,7 +25,12 @@ function Geocoder(indexes, options) {
     var q = queue(10);
 
     this.indexes = indexes;
-    this.replacer = token.createGlobalReplacer(options.tokens || {});
+
+    let globalTokens = options.tokens || {};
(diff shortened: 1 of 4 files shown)


ingalls:
Oh shoot, thanks for getting this @apendleton. I had done the main `carmen.js` fxn but didn't consider the carmen-indexer endpoint.

PR #644, opened 2017-07-28
remove russian as fallback from latin script languages

mattficke:
### Context
When a feature lacks text in a requested language, we fall back through a hard-coded list of alternate languages that are similar to the requested language. However, some languages that are written in Latin script include Russian in the fallback list. Returning an entirely different script than was requested is probably incorrect behavior.

### Summary of Changes
* Remove Russian as a fallback from Estonian, Latvian, and Lithuanian
* Add Lithuanian as a fallback to Latvian

### Next Steps
* Polish is the currently second fallback for Lithuanian, but English is the second fallback for Latvian. Given that these two language are most similar to each other, should their fallbacks be the same? Which is the better option?
* Should Russian be a fallback for the Azerbaijani language? It's written in Latin script within Azerbaijan, the Cyrillic version seems to date from the period of Russian/Soviet control.

cc @mapbox/geocoding-gang


boblannon:
> Polish is the currently second fallback for Lithuanian, but English is the second fallback for Latvian. Given that these two language are most similar to each other, should their fallbacks be the same? Which is the better option?

yeah, not sure what the particular rationale was for `lt` -> `pl`, given that it wasn't also applied to `lv`.  the mediawiki fallback chain has `en` as the primary for both.  

the weird thing is that before #555 `lv`'s primary fallback was ....`lv`?!  @sbma44 updated the chain to "reflect feedback from human translators"

> Should Russian be a fallback for the Azerbaijani language? It's written in Latin script within Azerbaijan, the Cyrillic version seems to date from the period of Russian/Soviet control.

according to teh [CLDR survey data](http://www.unicode.org/cldr/charts/latest/supplemental/territory_language_information.html#AZ) `az_Cyrl` is marginal compared to `az` (in Latin script), but it's the most common alternative script.

Wikipedia mentions that Azerbaijani is [still written in Cyrillic in Dagestan](https://en.wikipedia.org/wiki/Azerbaijani_language#Writing_systems).

apendleton:
Polish is not especially closely related to Latvian or Lithuanian (it's a Slavic language whereas the latter two are Baltic), and I wouldn't imaging too many Lithuanian or Latvian speakers would also speak Polish by happenstance, whereas a decent number would speak either Russian or English or both, so that swap sounds good to me.

Note also that the fallback lists are used both at query time in selecting display language and at index time in precomputing fallback penalty. Indexes built before this change will still work fine after it, but upon next rebuild their behavior might change slightly with regard to the relevance of results that go down this particular fallback chain.

mattficke:
@apendleton @boblannon 🙇 thanks. I think what makes sense here, then, is:

* Swap out Polish for English as the Lithuanian fallback
* Remove Russian as the fallback for Azerbaijani
* Add Turkish as first fallback for Azerbaijani, followed by English

boblannon:
> Remove Russian as the fallback for Azerbaijani

I meant to point out that the CLDR survey does say that there's 100% literacy for `az_Cyrl`, which might argue for leaving it.  It's not always clear how accurate those stats are, though.

> Add Turkish as first fallback for Azerbaijani, followed by English

Why Turkish?

boblannon:
> > Add Turkish as first fallback for Azerbaijani, followed by English

> Why Turkish?

oh, duh, related languages, sorry.  was thinking geographically and expected you to consider persian

mattficke:
>there's 100% literacy for az_Cyrl, which might argue for leaving it

I think if a user is requesting `az` they're most likely requesting a Latin script, and returning an unexpected script instead has tended to be the wrong call, even if it's a language that might be understood by the user.

> Why Turkish?

The Wikipedia article says they're closely related and largely mutually intelligible -- do you think this makes sense?

boblannon:
okay, sgtm!

PR #645, opened 2017-08-01
Consistify penalization of default text

yhahn:
This branch revisits how we penalize default text in relation to translated text within an index.

Before, a feature with only default text would never be penalized for missing translation text. Its default text was treated as though it were a translation into every language.

After, a feature's default text is treated consistenly like any other language. It gets penalized for not being an exact language match in every case.

Though indexes generated with this branch are format/schema compatible, I'd like to treat this as a semver breaking change as newly built indexes do not play nice with previously built indexes.

------

### Changes

- Adds unit tests for cross language query scenarios we'd like to support.
- Updates handling of default text in indexer to be more like any other language -- a fallbackable step-child rather than a spoiled only-child.
- Updates index-level language fallback to use a new `unmatched` key rather than `default`. This prevents default text from getting special treatment in indexes without any languages defined at all.

In-context annotations coming to help review.

------

### Next actions

In addition to overall review of approach I'd like to propose adjusting our cross-language penalty in carmen-cache from `* 0.8` to something more gentle like `* 0.95`. This is an arbitrary nerf, but given how much more common this nerf will be after this change this seems appropriate to help anyone who's used relevance scores as a guide to confidence thresholds or other client logic.

- [x] Code review
- [x] IRL testing results
- [x] Gentle-ify carmen-cache penalty

cc @mapbox/geocoding-gang


index.js
```diff
@@ -147,6 +147,7 @@ function Geocoder(indexes, options) {
             lang.hash = crypto.createHash('sha512').update(JSON.stringify(lang.languages)).digest().toString('hex').slice(0,8);
             lang.lang_map = {};
             lang.languages.forEach(function(l, idx) { lang.lang_map[l] = idx; });
+            lang.lang_map['unmatched'] = 128; // @TODO verify this is the right approach
```
yhahn: @apendleton moral compass.

This assigns a language code of 128 (past the bitmask limit?) to the `unmatched` language which makes an appearance below as how indexes that don't support the queried language will proceed when being queried.

**Example:** Index A supports `['en', 'es']`. User searches using `language=ru`. Previously, this would trigger an unpenalized `default` text search. Now this triggers an `unmatched` text search, which can still yield `default` text results but they are now penalized as "cross-language" (`unmatched => default`).

lib/indexer/indexdocs.js
```diff
@@ -70,9 +70,6 @@ function indexdocs(docs, source, options, callback) {
 
     if (!source.lang.fallback_matrix) {
         var matrix = cl.fallbackMatrix(source.lang.languages.filter(function(l) { return l != "default"; }));
-        for (var value of matrix.values()) {
-            value.push("default");
-        }
```
yhahn: Removed `default` from being included as an unpenalized fallback in every language matrix.

lib/phrasematch.js
```diff
@@ -83,7 +83,7 @@ module.exports = function phrasematch(source, query, options, callback) {
                     return source.lang.lang_map[l];
                 } else {
                     var label = cl.closestLangLabel(l, source.lang.lang_map);
-                    return source.lang.lang_map[label || 'default'];
+                    return source.lang.lang_map[label || 'unmatched'];
```
yhahn: Here's where previously language queries that had no index-level entry match would trigger an unpenalized `default` search before.

lib/util/termops.js
```diff
@@ -287,20 +287,7 @@ function getIndexableText(replacer, globalTokens, doc) {
...
-    // alternatively, if this doc has no translations, there are no language-specific texts that could win
-    // so this feature would always get penalized in a language-specific search, which we don't want, so
-    // index it as all-language text
-
-    if ((!langTexts.size) || (langTexts.size === 1) && (doc.properties['carmen:text_universal'])) {
-        getTokens(texts, 'all');
-    } else {
-        getTokens(texts, 'default');
-    }
+    getTokens(texts, 'default');
```
yhahn: This logic removal looks scary but it's not -- just means we always index default text as default text.

Below this in the full file is where universal text is indexed under the `all` key.

index.js
apendleton: Yeah carmen-indexer only supports 0-127 (since they get stored in a 128-bit int)

> Index A supports ['en', 'es']. User searches using language=ru. Previously, this would trigger an unpenalized default text search.

This is not actually exactly the current behavior. Currently, if the user requests a language that the index doesn't have, we run it through fallback to identify the closest language the index **does** have (falling back to default otherwise), and searches with whatever it finds as unpenalized. In general this is pretty important, because it's what makes users who supply language subtags (e.g., "en-UK") work right if all we have is "en." See https://github.com/mapbox/carmen/blob/master/lib/phrasematch.js#L80-L88

I'm a little hazy on how this change would interact with that

lib/phrasematch.js
apendleton: Ah okay, ignore my above comment :)

Issue #574, opened 2017-01-20
Accept antimeridian-crossing bboxes


mattficke:
### Context

Carmen currently implements and enforces bounding boxes per the [2008 GeoJSON spec](http://geojson.org/geojson-spec.html#bounding-boxes):

>The value of the bbox member must be a 2*n array where n is the number of dimensions represented in the contained geometries, with the lowest values for all axes followed by the highest values.

For features with geometries that cross the antimeridian (such as the United States), this leads to flawed bounding boxes.

#### Current US bbox:
![image](https://cloud.githubusercontent.com/assets/9308599/22131472/7b06b558-de68-11e6-9409-5c84b198ad37.png)

#### What it should be:

![image](https://cloud.githubusercontent.com/assets/9308599/22131566/2dedfd5c-de69-11e6-923a-4ac0b7c3c64e.png)

### Next Steps
To fix this, we should implement the updated bbox spec from [RFC 7946](https://tools.ietf.org/html/rfc7946#section-5):
>The value of the bbox member MUST be an array of
   length 2*n where n is the number of dimensions represented in the
   contained geometries, with all axes of the most southwesterly point
   followed by all axes of the more northeasterly point.

BBoxes are used both at indexing time and at query time.

* Indexer: if I'm reading the [turf docs](http://turfjs.org/docs/#bbox) correctly, this can be solved by simply upgrading to a recent version of `turf-bbox`. Turf can calculate spec-compliant bboxes in `west, south, east, north` order.
* Query time: used in the [bbox filter](https://github.com/mapbox/carmen/pull/457), this will need to be modified to accept SW/NE bboxes. Trickiest part will likely be in carmen-cache, where we'll need to properly handle the conversion to z/x/y tile coords in a more sophisticated manner.

In case we need to do our own bbox calculations, capturing @apendleton's notes from chat:

> * if the geometry exists entirely in the eastern or western hemisphere, continue as now
> * if not:
>     * split the geometry at the prime meridian and calculate separate bounding boxes for the eastern half and western half
>     * determine the width of the PM-crossing bounding box by adding (0 - westMinX) + (eastMaxX - 0)
>     * determine the size of the AM-crossing bounding box by adding (westMaxX - -180) + (180 - eastMinX)
>     * pick the smaller of the two, then:
>         * if PM-crossing wins, bbox goes from westMinX to eastMaxX
>         * if AM-rossing wins, bbox goes from eastMinX to westMaxX

[...]

PR #646, opened 2017-08-10
[W, S, E, N] bbox order

mattficke:
### Context
A bounding box property in included for features with polygon geometry, formatted in `[minX, minY, maxX, maxY]` coordinate order. When feature polygons include coordinates close to and on both sides of the anti-meridian (such as Russia, the United States, Fiji, and others), the resulting bounding box spans nearly the entire width of the world.

In these situations we should return the smaller of the two possible bounding boxes, and allow coordinates to be ordered `[W, S, E, N]` rather than `[minX, minY, maxX, maxY]`.

refs https://github.com/mapbox/carmen/issues/574

### Summary of Changes
* Generate anti-meridian crossing bboxes for features that straddle the dateline

### Next Steps
- [x] Still need to add a backwards-compatible migration function to optionally clip anti-meridian crossing bboxes at +-180 degreees

Also note that this does not change the bbox query parameter, which still only accepts bboxes in `[minX, minY, maxX, maxY]` order.

cc @mapbox/geocoding-gang @brsbl


mattficke:
I think this is ready to go, though I'd like another set of eyes on the clipping section in particular to make sure this approach is a good one. I made anti-meridian clipping the default behavior, which you can turn off with an options setting.

mattficke:
>Rather than making clipBBox an index-time option, could this be done at runtime and applied at the final step of returning features?

This is done.

1ec5:
This PR turns out to be consistent with the terminology used [in MapboxGeocoder.swift](https://github.com/mapbox/MapboxGeocoder.swift/blob/6a4645bf8b19a1cb3703894141fc1bfe5d316d12/MapboxGeocoder/MBPlacemark.swift#L390-L392). 👍

mattficke:
@1ec5 glad to hear it! Out of curiosity, how would `RectangularRegion` handle antimeridian-crossing situations? Our current implementation ensures that this doesn't happen in practice, but it's possible this will change in a future API version.

1ec5:
[The documentation](https://github.com/mapbox/MapboxGeocoder.swift/blob/b55d3f1bf0985d8a43098c8bf97a4eb2153bd626/MapboxGeocoder/MBRectangularRegion.swift#L14) currently says:

> `southWest` must be to the southwest of `northEast`. The region may not span the antimeridian. If you need to restrict a query to a region that spans the antimeridian, such as the region that encompasses Fiji, perform two queries each limited to the region on either side of the antimeridian, then combine the results.

MapboxGeocoder.swift has yet to reach v1.0, so we’re free to relax the prohibition against spanning the antimeridian at any point. As a developer, I’d expect to be able to provide a single bbox that straddles the antimeridian in the request or get a similar bbox in the response. That’s how the Mapbox GL libraries work (including the iOS and macOS SDKs). Do we have a ticket tracking antimeridian-straddling bboxes in requests?

PR #647, opened 2017-08-24
Remove stacky, gappy penalties

yhahn:
TL;DR I think it's time to remove the gappy/stacky penalties.

In `master` we apply a very slight penalty and a very slight bonus in these scenarios:

- If you search for something using text that gives it additional context, we give a slight bonus to that feature ("stacking")
  - Example Query: "Atlanta Georgia"
    - "Atlanta Georgia" (the place, region) gets a slight text relev bonus in comparison to
    - "Atlanta Georgia's Best Beer Brewery" (the poi)
- If you search for something where the context is closer in hierarchy to the feature involved, it gets a higher bonus than a feature where the context is very far away in hierarchy
  - Example Query: "100 Main Street United States" << `address <=> country` context distance is far
  - Example Query: "100 Main Street Buffalo" << `address <=> place` context distance is short

These bonuses and penalties were meant to address a variety of corner case issues. One that I remember distinctly is this:

`12345 Georgia`

This could be a postcode search (the postcode `12345` in the state of `Georgia`) or maybe it's the beginning of an address (`12345 Georgia Ave`). In `master` we've decided to favor the former.

---

### A

For the query `69th Ave Oakland` we are currently matching

1. `69th Ave` (address) in `Oakland Gardens` (neighborhood) in `New York` (place)
2. `69th Ave` (address) in `Oakland` (place)

**Explanation:** 1 is beating 2 here because of the `gappy` penalty -- neighborhood and address are closer in our hierarchy than place and address. :disappointed:

**My personal response:** this feels super arbitrary and wrong.

### B

Roughly the same story. For the query `1509 16th St NW` we are currently matching

1. `1509 16th St` (address) in `Northwest Everett` (neighborhood) in `Everett` (place)
2. `1509 16th St NW` (address) in `Washington DC` (place)

**Explanation:** 1 is beating 2 here because of the `stacky` bonus. 2 is a contiguous query, while 1 has "additional context" ("NW" => "Northwest Everett") supplied to the address component.

**My personal response:** this also feels super arbitrary and wrong.

---

## Time to remove

Based on these cases and some IRL testing, the actual cases that stacky and gappy bonus/penalties were addressing seem to have been addressed through other means (e.g. @mattficke's work penalizing directional flips). For example, our stacky and gappy unit tests still pass with this functionality removed :smirk:

cc @mapbox/geocoding-gang

### DIFF
--- a/lib/verifymatch.js
+++ b/lib/verifymatch.js
@@ -244,8 +244,6 @@ function verifyContexts(contexts, sets, indexes) {
 
 function verifyContext(context, peers, strict, loose, indexes) {
     var addressOrder = context[0].properties['carmen:geocoder_address_order'];
-    var gappy = 0;
-    var stacky = false;
     var usedmask = 0;
     var lastmask = -1;
     var lastgroup = -1;
@@ -289,8 +287,6 @@ function verifyContext(context, peers, strict, loose, indexes) {
             } else if (direction === "descending") {
                 backy = lastmask < matched.mask;
             }
-            stacky = true;
-            gappy += Math.abs(indexes[feat.properties["carmen:index"]].ndx - lastgroup) - 1;
         }
 
         usedmask = usedmask | matched.mask;
@@ -303,8 +299,6 @@ function verifyContext(context, peers, strict, loose, indexes) {
         }
     }
 
-    if (!stacky) relevance -= 0.01;
-
     // small bonus if feature order matches the expected order of the index
     if (direction) relevance -= 0.01;
     if (addressOrder === direction) relevance += 0.01;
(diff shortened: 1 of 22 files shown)


lib/verifymatch.js
```diff
@@ -244,8 +244,6 @@ function verifyContexts(contexts, sets, indexes) {
```
yhahn: This is the main substance of the PR, removing the stacky & gappy logic in verifymatch. The rest is all test update fallout (e.g. relevance scores that were 0.99 => 1.00).

Note that `geocode-unit.stacky.test.js` and `geocode-unit.gappy.test.js` still exist and pass.

aarthykc:
This looks good to me @yhahn ! 🌮 , thank you for the explanation!

PR #648, opened 2017-08-25
The Numnum Bug

yhahn:
Ran into this debugging a wacky address query in Sweden:

```
115 39 Sweden
```

Can currently return something like 

```
# this has relevance 1
115 115, 269 76 Torekov, Skåne län, Sweden
```

What's going on here where we end up with the house 115 on street 115 but no sign of 39 to be found?

### What's going on

- Query gets numtokenized and normalized to these two permutations: `## 115` `1## 39` (we always index the data with housenum first).
- `## 115` matches data
- On the way back out, the `maskAddress` is `11` (both term positions are potential house numbers) and maskAddress is called against the original _query_ which has tokens in this order `115 39`
- It snags the first housenum-like value, `115`
- And pairs it with the street that's been found, `115`

And so we end up with `115 115`.

### The short-term fix (in this branch)

Make `maskAddress` a bit smarter by looking at both the coverText and query to determine if it's about to reuse a housenum that was really originally interpreted as a street.

### The long-term fix

When we `numTokenize` at the initial phrasematch stage, extracting the housenumber early may make more sense rather than masking and then doing forensics later to try to guess at what happened.

---

For now have put the stopgap short term fix in place.

cc @mapbox/geocoding-gang

### DIFF
--- a/lib/util/termops.js
+++ b/lib/util/termops.js
@@ -151,7 +151,7 @@ function tokenize(query, lonlat) {
  * is a valid address
  *
  * @param {Array} query tokenized query
- * @param {Array} tokenized cover text
+ * @param {String} cover text
  * @param {Integer} mask a mask for the given query
  * @return {Object} returns an address object or null
 */
--- a/lib/util/termops.js
+++ b/lib/util/termops.js
@@ -4,7 +4,6 @@ var idPattern = /^(\S+)\.([0-9]+)$/;
 var token = require('./token');
 var permute = require('./permute');
 var cl = require('./closest-lang');
-var levenshtein = require('fast-levenshtein');
 
 module.exports.id = id;
 module.exports.terms = terms;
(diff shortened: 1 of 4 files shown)


lib/util/termops.js
```diff
@@ -151,14 +151,20 @@ function tokenize(query, lonlat) {
  * is a valid address
  *
  * @param {Array} query tokenized query
+ * @param {Array} tokenized cover text
```
mattficke: I think this needs the `coverText` parameter name added?

lib/util/termops.js
```diff
@@ -151,14 +151,20 @@ function tokenize(query, lonlat) {
...
  * @param {Integer} mask a mask for the given query
  * @return {Object} returns an address object or null
 */
-function maskAddress(query, mask) {
+function maskAddress(query, coverText, mask) {
+    var coverTokens = new Set(tokenize(coverText));
```
mattficke: Ooh. I haven't used Set before, this is a nice example.

lib/util/termops.js
yhahn: :+1:

lib/util/termops.js
yhahn: learned from @apendleton

PR #654, opened 2017-09-11
Unintended cross-language penalties

mattficke:
### Context
We apply a relevance penalty to queries that don't match the requested language (or, if no language is specified, the default text). This causes undesirable behavior in some situations (including numeric postcodes and common abbreviations used in many languages), so we added a `text_universal` language property that can contain values to which this penalty won't be applied.

There's a problem, however, when the same value is used in `text_universal` and a language-specific property. In these ambiguous situations, Carmen thinks the query is matching the language-specific value rather than the universal one, and applies the penalty. I've added a failing test to illustrate the problem, not sure the best way to fix this yet.

### Summary of Changes
* add unit test

### Next Steps
Fix the thing

cc @mapbox/geocoding-gang


mattficke:
This was an indexer bug, we would only treat a token as universal text if it wasn't shared with another language. We were checking if something was universal text with:

```
if (langList == 'all')
```

This failed if the langlist was something like `['sv', 'all']` though. I've updated it to check for the presence of `all` instead with
```
if (langlist.includes('all')
```

Tip of the cap to @apendleton for assistance tracking this down.

PR #655, opened 2017-09-20
Multivariate calculus

apendleton:
### Context
This branch builds on the [normalization cache](https://github.com/mapbox/carmen/pull/652) work by introducing the ability for an index to represent a phrase that might normalize to more than one alternative phrase, or might map to both an alternative phrase and itself. The DAWG building mechanism has also been modified to track, when phrases are added to the DAWG, whether or not they are added as primary phrases or only as part of normalization mappings. In the event that they are added as both in the course of indexing different documents, the normalization cache created at dump time will preserve the original phrase as well as the alternate phrase as a target in the normalization mapping.

### Summary of Changes
- [ ] switch to https://github.com/mapbox/carmen-cache/pull/113
- [ ] modify the DAWG cache build steps to track additional information to detect phrases that should map both to themselves and to an alternate
- [ ] add tests of the new indexing and search behavior
- [ ] modify the merge logic to support the new formats
- [ ] improve merge tests to explicitly test this circumstance

### Next Steps
none

cc @mapbox/geocoding-gang


apendleton:
These tests pass locally but fail on circle because of https://github.com/mapbox/node-pre-gyp/issues/298 and will continue to as long as this branch is pointed at a URL rather than a published version of carmen-cache . Once that's released though, I'll bump this branch to the released version and the tests should fix themselves.

PR #656, opened 2017-09-21
Support duplicate address numbers in address clusters in forward queries

apendleton:
### Context
At present, if an address cluster contains multiple entries for a single address number, only one of them is considered in verifymatch. This PR alters the address cluster logic to return an array rather than a single point, of all matching numbers, so that verifymatch can consider any of them as potential responses.

Note that the deduplication mechanism that runs before results are returned will still collapse down to a single response if the duplicates end up having identical `place_name` values. There are a couple of circumstances where this change is still important, though:
* when a bbox excludes the value that would otherwise have been the one arbitrarily chosen as the response (in current production you get no results in this circumstance)
* when other layers in the stack end up giving different addresses within the cluster different `place_name` values, for example if a single address cluster spans multiple postcodes

### Summary of Changes
- [x] alter addressCluster.forward to return an array of all responsive values rather than a single value
- [x] alter verifymatch to expect an array and to potentially add multiple results per loaded feature
- [x] add tests of the addresscluster function as well as to geocodes over clusters with duplicate values

### Next Steps
None

### To-do for later
As of this PR there will be an inconsistency in the interfaces between the addressCluster.forward function, which returns an array, and the addressItp.forward function, which still returns a single object. Eventually we should adjust addressItp also, probably in a conceptually similar way so if more than one interpolation line in the same cluster could include the same address number, we return multiple candidate ITP points. I'm not sure how exactly this should work though, so I'm just doing this half for now in the interests of expediency.

cc @ingalls @mapbox/geocoding-gang

### DIFF
--- a/lib/pure/addresscluster.js
+++ b/lib/pure/addresscluster.js
@@ -13,9 +13,15 @@ function forward(feature, address) {
     for (var c_it = 0; c_it < cluster.length; c_it++) {
         if (!cluster[c_it]) continue;
 
-        // Check is cluster is pt geom
+        // this code identifies all the indexes of cluster[c_it] that have a1 or, if that fails, a2
+        // equivalent approximately to python [i for i,e in enumerate(cluster[c_it]) if e == a1]
+        // expressed as a reduction starting with an empty array `a` that, if the value e at each step
+        // `e` is equal to `a1`, gets the current index `i` appended to it
+        // more info at https://stackoverflow.com/questions/20798477/how-to-find-index-of-all-occurrences-of-element-in-array#comment69744472_20798754
         var a_index = cluster[c_it].reduce((a, e, i) => (e === a1) ? a.concat(i) : a, []);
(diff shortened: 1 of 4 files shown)


lib/pure/addresscluster.js
```diff
@@ -14,15 +14,18 @@ function forward(feature, address) {
...
-        if (a_index > -1 && feature.geometry.geometries[c_it].type === 'MultiPoint') {
-            return {
-                type:'Point',
-                coordinates: [
-                    Math.round(feature.geometry.geometries[c_it].coordinates[a_index][0]*1e6)/1e6,
-                    Math.round(feature.geometry.geometries[c_it].coordinates[a_index][1]*1e6)/1e6
-                ]
-            };
+        var a_index = cluster[c_it].reduce((a, e, i) => (e === a1) ? a.concat(i) : a, []);
+        if (!a_index.length) a_index = cluster[c_it].reduce((a, e, i) => (e === a2) ? a.concat(i) : a, []);
```
yhahn: Want to add a comment here that explains how this pulls out the indices of *n* matching address points? The reduce in ES6 format is super terse.

PR #657, opened 2017-09-21
Disable autocomplete when numerical token moved to beginning

lizziegooding:
### Context

Queries in the format `place`, `state abbreviation` `postal code` sometimes return address features where the state abbreviation has been autocomplete. For example, the query `Quincy, MA 02169` returns results like `02169 Maine St, Quincy, Illinois 62301, United States`.

### Summary of Changes

- [X] add unit test for `quincy, ma 02169`
- [X] set `ender` to false if numerical token is moved to the beginning of the query to prevent autocomplete

### Next Steps

- [ ] run tests in api-geocoder -- 🍐 with @mattficke tomorrow

per @yhahn 's comment, 
> We'll want to keep an eye out for how this affects German autocomplete (where house numbers appear last).

cc @mapbox/geocoding-gang


lib/util/termops.js
```diff
@@ -383,6 +383,7 @@ function permutations(terms, weights, all) {
         // leading-numtoken version of a phrase.
         if (permutation[permutation.length-1].indexOf('#') !== -1) {
             permutation.unshift(permutation.pop());
+            permutation.ender = false;
         }
```
mattficke: One thing to look at here is what happens when there's only one token present (i.e., when `permutation.length === 1`). Does it still make sense to set `ender` to false in that case?

PR #658, opened 2017-09-24
Perf optimization for enumerateTokenReplacements

yhahn:
### Context

Token replacements are one of the more expensive things in our indexing process. @apendleton's recent work adds a way to generate many permutations (within reason) of strings that have multiple occurrences of replaceable tokens, e.g.

```js
q.deepEqual(
    token.enumerateTokenReplacements(tokens, 'fargo street northeast, san francisco'),
    [
        'fargo St NE, sf',
        'fargo St NE, san francisco',
        'fargo street NE, sf', 'fargo street NE, san francisco',
        'fargo St northeast, sf',
        'fargo St northeast, san francisco',
        'fargo street northeast, sf',
        'fargo street northeast, san francisco'
    ]
);
```

To support tokens that occur multiple times (e.g. `St Martin St`) we have a rather complex code path.

This branch adds a separate code path that is much simpler for occasions when there is only one match for a target replacement.

### Bench comparison

Using a sample of Australia OpenAddresses data

```sh
### master

node bench/index.indexdocs.js short.geojson 
# indexdocs
index short.geojson x 8.51 /sec (45 runs sampled)

### fasterly

node bench/index.indexdocs.js short.geojson 
# indexdocs
index short.geojson x 10.57 /sec (54 runs sampled)
```

### Next steps

- [x] @apendleton can you give this a review to make sure this optimization approach is conceptually sound?

cc @mapbox/geocoding-gang


lib/util/token.js
```diff
@@ -131,16 +131,36 @@ module.exports.enumerateTokenReplacements = function(tokens, query) {
     let outSet = new Set([query.trim().toLowerCase()]);
 
     for (let token of tokens) {
-        // escape hatch on the check to make sure we don't run away with infinite recursion
-        for (var i = 0; i < Math.min(out.length, 64); i++) {
+        for (var i = 0; i < out.length; i++) {
+            // escape hatch on the check to make sure we don't run away with infinite recursion
+            if (i > 64) break;
```
yhahn: This is a small optimization - previosly `Math.min()` would be run on each iteration of the for loop.

lib/util/token.js
```diff
@@ -131,16 +131,36 @@ module.exports.enumerateTokenReplacements = function(tokens, query) {
...
+
             // we want to enumerate each possible replacement of each token, unless it's an XRegExp, where this whole trick doesn't work
             var occurrenceCount = token.named ? 1 : (out[i].match(token.from) || []).length;
+            if (occurrenceCount === 0 && i === 0) break;
```
yhahn: If the target replacement can't be found at all in the initial string, we can skip this token for subsequent permutations.

apendleton: Hm, I'm not sure about this one, because there's the possibility that it could be introduced by a subsequent replacement. This *probably* wouldn't happen, but I'm having a hard time fully imagining all the permutations here. Like, suppose we had a token replacement that expanded a short string to a longer string that contained a letter with an umlaut on it, and we wanted to also index both the umlaut and plain+e versions. I guess with this change, in the event that such a replacement were necessary, it would be incumbent on geocoder-abbreviations to supply both versions?

yhahn: To make this concrete...

```
über => ub
ü => ue
ü => u
```

these are the permutations we'd generate now:

```
0 ub car company
1 über car company
2 uber car company
3 ueber car company
```

with this optimizations we would miss the latter permutations:

```
0 ub car company
1 über car company
```

Let me write a testcase to show this to confirm we're thinking about this correctly? And then yank this particular optimization?

apendleton: That's the case I'm concerned about, yes. Though to be clear, at least as relates to the umlauts and the other German replacements currently defined in geocoder-abbreviations, this is a theoretical concern -- we don't introduce or remove any umlaut'ed words at the moment as far as I can tell. I'm having difficulty thinking through whether or not something similar might happen in other circumstances, though, or might be caused by complex interactions between more than one set of tokens applied to the same text.

PR #660, opened 2017-10-02
try out prettier

mattficke:
### Context
Just an experiment to see what would happen if we used [prettier](https://github.com/prettier/prettier) on a pre-existing project.

Done with `prettier --single-quote --tab-width 4 --write "./**/*.js"`

I kinda like it 😄 

cc @mapbox/geocoding-gang


ingalls:
I think this is a great idea and highlights some features we should hard enable in ESLint. That said I'm not psyched on prettier as a whole as it basically an configurable ESLint instance.

Things I really like
- Enforce 4 space tabs across the project including test/bench/lib
- Use single quotes within carmen lib
- Consistent spacing between fxn args and braces/brackets

What I'm not psyched about:

- Strange multi line string declarations

```+const WORD_BOUNDARY =
+    '[\\s\\u2000-\\u206F\\u2E00-\\u2E7F\\\\\'!"#$%&()*+,\\-.\\/:;<=>?@\\[\\]^_`{|}~]';**
```

- Multline if statements

```
+ if (
+                    from === stripped ||
+                    replacementOpts.skipDiacriticStripping
+                ) {
```

- Enforching single quotes in test means that we can't paste test fixutres into GeoJSON without converting all of them back to double quotes as single quotes are invalid JSON.
- Arrays are being forced into the expanded line delimited form, this makes browsing test files a headache as you have to skip hundreds of lines of array elements.
- Callbacks without parenthesis are a pain as you have to re-add them every time you work with or add to the callback (same reason I support `{}` on single line if/else statements. Hey at least I'm consistent.
- All functions on newlines can be needlessly verbose where the team as a whole tends to collapse.
```
q.defer(
  function()
````

- Multiline console.logs

```
console.log(
   'help doc here'
);
```

- Expanded function returns

That said I think this is a great way to kickstart the conversation around updating some of our ESLint defaults to mirror things that the majority of the team already do.

apendleton:
Yeah, the "make all the test data a million lines long" was the one that really jumped out at me as being obnoxious.

mattficke:
I think there's an option to disable it for specific blocks, if we want to preserve the original format for things like test fixtures.

As far as I understand it, the idea with prettier is to use it in conjunction with something like ESlint -- use the linter for substantive checks (no unused variables, no modifying native objects, etc), and prettier for stylistic issues.

mattficke:
Sounds like we should stick with eslint rather than adding a new tool. Let's update that instead.

PR #664, opened 2017-10-10
[DONOTMERGE] Refactored token replacement (reference)

yhahn:
### Context

@apendleton explored some of the ideas you mentioned last week of doing token replacement similarly to how we handle diacritics - mash them together into a single `/(WORD BOUNDARIES)(token|token|token|token)(WORD BOUNDARIES)/` regex for performance.

I took more of a PoC/bluesky refactor approach here because I could see a path toward much less code complexity/craziness -- will annotate below.

I think the approach hinges on being able to eliminate named token replacement/move to using custom callbacks. I also don't know what other changes you have planned so just want to put this up as a reference.

Despite using real recursion for enumerating variations this version is still faster than master by 50% or so (0.80hz to 1.2hz for me locally).

### Next Steps

- :eyes: from @apendleton, incorporate into his new work if makes sense

cc @mapbox/geocoding-gang

### DIFF
--- a/index.js
+++ b/index.js
@@ -115,9 +115,11 @@ function Geocoder(indexes, options) {
             source.token_replacer = token.createReplacer(info.geocoder_tokens||{});
             source.indexing_replacer = token.createReplacer(info.geocoder_tokens||{}, {includeUnambiguous: true, custom: source.geocoder_inverse_tokens||{}});
 
+/*
             if (tokenValidator(source.token_replacer)) {
                 throw new Error('Using global tokens');
             }
+*/
 
             source.maxzoom = info.maxzoom;
             source.maxscore = info.maxscore;
--- a/lib/util/token.js
+++ b/lib/util/token.js
@@ -1,3 +1,4 @@
+var uniq = require('./uniq');
 var removeDiacritics = require('./remove-diacritics');
 var XRegExp = require('xregexp');
 
@@ -10,7 +11,6 @@ module.exports.createReplacer = function(tokens, inverseOpts) {
     inverseOpts = inverseOpts || {};
     inverseOpts.includeUnambiguous = inverseOpts.includeUnambiguous || false;
 
-    var replacers = [];
     let isInverse = {};
 
     if (inverseOpts.includeUnambiguous) {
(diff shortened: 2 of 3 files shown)


index.js
```diff
@@ -115,9 +115,11 @@ function Geocoder(indexes, options) {
             source.token_replacer = token.createReplacer(info.geocoder_tokens||{});
             source.indexing_replacer = token.createReplacer(info.geocoder_tokens||{}, {includeUnambiguous: true, custom: source.geocoder_inverse_tokens||{}});
 
+/*
             if (tokenValidator(source.token_replacer)) {
                 throw new Error('Using global tokens');
             }
```
yhahn: This would need updating since the format of replacers changes with this.

lib/util/token.js
```diff
@@ -37,27 +35,24 @@ module.exports.createReplacer = function(tokens, inverseOpts) {
...
 
-        var replacementOpts = {};
-        if (typeof to == 'object' && to.text) {
-            replacementOpts = to;
-            to = to.text;
-        }
+    const groups = {
+        skipBoundaries: { fromList:[], tokens:{} },
+        withBoundaries: { fromList:[], tokens:{} }
+    };
```
yhahn: First pass groups tokens into these two categories (withBoundaries, skipBoundaries) since we'll need to generate two separate MEGA regexes for these cases.

lib/util/token.js
```diff
@@ -71,42 +66,43 @@ module.exports.createReplacer = function(tokens, inverseOpts) {
...
-                // count groups in original regex, using the trick from https://stackoverflow.com/a/16046903
-                let count = new RegExp(from + "|").exec('').length - 1;
-
-                // increment replacements indexes in `to`
-                if (typeof to == "string" && !replacementOpts.skipBoundaries) {
-                    to = to.replace(/\$(\d+)/g, function(str, index) { return '$' + (parseInt(index)+1).toString();});
-                    entry.to = '$1' + to + '$' + (count + 2).toString(); // normalize abbrev
-                } else {
-                    entry.to = to; // normalize abbrev
-                }
```
yhahn: Code below generates 1 or 2 replacers with the word-bounded and unbounded MEGA regexes.

lib/util/token.js
```diff
@@ -122,75 +118,38 @@ module.exports.replaceToken = function(tokens, query) {
     return abbr;
 }
 
-module.exports.enumerateTokenReplacements = function(tokens, query) {
+module.exports.enumerateTokenReplacements = function enumerateTokenReplacements(tokens, query, offset, out) {
```
yhahn: This whole process gets a lot simpler with a recursive process where each found token replacement spawns two paths of recursion:

- A: we replace the token and then recurse through the rest of the string
- B: we don't replace the token and then we recurse through the rest of the string

You'll see this ends up generating the same permutations in the tests, just in a different order.

PR #665, opened 2017-10-12
Unsquish countries

mattficke:
### Context
In https://github.com/mapbox/carmen/pull/538 & https://github.com/mapbox/carmen/pull/546, we added a "squishy" option that allowed more-granular features to beat out identically named, but larger features within which they were contained. It's relatively common for large cities to have the same name as their containing region (`New York, New York` being a notable example), and in these cases the more granular feature tends to be the more notable one. So we let the smaller feature inherit the score of the parent feature, and it ends up on top of the results.

Turns out this same pattern exists in cities named after their containing _country_ -- Mexico City, Panama City, etc. In these cases, a query for `Mexico` should probably return the country first, but the current squishy logic promotes the city instead.

Countries should not participate in this score inheritance, so this adds an option to exclude individual indexes from participating.

### Summary of Changes

- add a new index-level option, `geocoder_grant_score`, that lets individual indexes opt-out from passing on their scores to child features. Defaults to `true`, opt-out by setting to `false`.

cc @mapbox/geocoding-gang


mattficke:
Closing for now, can revisit if needed.

PR #667, opened 2017-10-17
Fuzzworld

aaaandrea:
[Context](https://github.com/mapbox/geocoding/issues/158#issuecomment-337028078)
#### Next Steps

- [x] @aarthykc - Uniform places where normalization is being called to change with the uniform API etc. 
- [x] @aarthykc  - Opt into fuzzy matching in phrasematch 
- [x] @aarthykc  - In Phrasematch we will conditionally pass flag extra flag based on contents of the string. 
- [x] @aarthykc - Latin letters checking. 
- [x] @aaaandrea - Look at the response to see if the match string is different from the one we turned in. we will need to pass in the real string and pass into coalesce or the response string and pass that into coalesce. 
- [x] both -Tests

coming soon: check relavence calculations: communicate to carmen-cache or verify match to penalize fuzzy matches. either generic or not so carmen-cache can apply


aaaandrea:
This work was done in the [Fuzzy Phrase](https://github.com/mapbox/fuzzy-phrase) repo.

Issue #674, opened 2017-12-31
[routablequest] 2. Overview & Docs


yhahn:
Walk through the background and history of the problem space in person with @mattficke and @apendleton, ideally grabbing a whiteboard. You'll want to cover:

- The use case we're trying to address for navigation applications - snapping a forward and reverse geocode to the associated nearby road, not necessarily the closest by distance
- Why possible standalone solutions require developers to do a lot of the work of a geocoder (e.g. fuzzy text matching, or associating address points to roads, or...)
- How address features are represented in carmen's input data as single features with both multipoint clusters and multilinestring segments for the same road
- Why reverse geocoding a user's location to the nearest housepoint and then snapping to the associated road may yield better results than just snapping to the nearest road
- Why it may not and what we'll learn if it doesn't

### Documenting your goal

Once you're done, covering the problem space background, open up `carmen-geojson.md` which documents the geojson format of results from carmen geocodes. Think through where and how you'd like the `routable_point` property to be added and add documentation of the behavior you'll be trying to implement in carmen.

- [ ] Discuss overview/problem space with @mattficke @apendleton
- [ ] Create a new branch `routablequest-docs`
- [ ] Update `carmen-geojson.md` documenting the `routable_point` property you'll be adding
- [ ] Create a PR describing your change and ask for a review

Once you're done with those steps, close this issue.

cc @nmargolis @mattficke @apendleton

Nmargolis:
A few thoughts on:
> Why reverse geocoding a user's location to the nearest housepoint and then snapping to the associated road may yield better results than just snapping to the nearest road

We discussed a few tricky cases with reverse geocoding:
1. When the point returned by geolocation is closer to a part of the road network that is not a useful road (e.g. an alley behind a house). If we reverse geocode to a housepoint first, we can take advantage of the functionality we will build with forward geocoding that will use the associated road that is part of the same feature.
2. Unlike in forward address geocoding where the street name is given explicitly, so we know what the most relevant street is, it's possible that the housepoint result from reverse geocoding could be on a different street than where the user is (especially if the point is near a corner). Reverse geocoding first will help avoid case #1 above, since it will force a point on a road associated with an address, but it may be a different road than intended if the resulting address is on a different street.
[...]

PR #677, opened 2018-01-02
Add documentation for routable_point for Carmen GeoJSON

Nmargolis:
### Context
Proposed new spec for routable_point feature as described in #674 

### Summary of Changes
- [ ] Add routable_point as a top-level key on the Carmen GeoJSON spec

### Next Steps
I have a few open questions:
1. Not sure if it makes sense for routable_point to be a top-level key with the other carmen keys (in addition to standard GeoJSON), or if it should be nested under properties. I have it here as a top-level key because it seems to be the most consistent with the way things are done now.  It seems that as it is now, relevant, standardized location data is all top-level (center, bbox, etc), and properties seem to be non-standardized byproducts of data sources. Adding it top-level will mean a bigger change to the Carmen spec (and more deviation from standard GeoJSON), but hopefully adding properties shouldn't break any existing functionality. Also, are there other use-cases where information like this that will be consumed by API users is nested in properties? If not, adding to the properties would mean accessing a new nested field that API consumers may not be currently accessing.

2. Does the order matter? I've added it to the end, but if there's a reason it should be in a different order, let me know!

cc @mapbox/geocoding-gang

### DIFF
--- a/carmen-geojson.md
+++ b/carmen-geojson.md
@@ -58,6 +58,7 @@ key | description
 `address` | Where applicable. Contains the housenumber for the returned feature
 `center` | Array of the form [lon,lat].
 `context` | Array representing a hierarchy of parents. Each parent includes `id`, `text` keys.
+`routable_point` | Optional. Array of the form [lon,lat]. Closest point on the relevant road, if the feature has an associated road.
 
 For geocodes that include one or more language codes set by `options.language`, the following keys will also be returned for each language requested:
 


mattficke:
>I have it here as a top-level key because it seems to be the most consistent with the way things are done now.

👍

>Does the order matter?

Nope! Order is not guaranteed and shouldn't be relied upon for JS objects as a general rule.

Looks good! On `properties`, you've flagged something we need to think about, we don't have a really rigorous definition of what goes in there at the moment. Let's punt on that for now :)

PR #681, opened 2018-01-17
Skip irrelevant indexes earlier when using country param

mattficke:
### Context
Currently, the country param only filters out features that are not in the requested country or countries relatively late in the process, in verifymatch. We end up doing a fair amount of work finding and sorting results from indexes that will have all their results thrown away at the end.

We can skip indexes that don't contain any features in the requested country earlier in the process in phrasematch to improve performance.

### Summary of Changes
* When using the country param, don't consider phrasematch results from indexes that don't contain features in the requested country.

cc @mapbox/geocoding-gang


mattficke:
This also adds support for the `bbox` and `languageMode` params to the carmen.js script.

PR #682, opened 2018-01-19
Update tokenize script to better display tokens

ingalls:
### Context

Rewrite the `tokenize` script to be more consistent in it's output and get rid of `split` who's functionality is now duplicated by core `readline`

cc @mapbox/geocoding-gang


aarthykc:
lgtm! 🌮

PR #683, opened 2018-01-22
Fix bbox in scripts/carmen.js

lizziegooding:
### Context
https://github.com/mapbox/carmen/pull/681 added support for `bbox` arguments from the carmen CLI. This fixes error handling for that change and adds tests to confirm it works as expected.

### Summary of Changes
- [x] check the length of the `bbox` arg and parse `bbox` rather than `proximity` parameter
- [x] add unit tests for cli runs with valid and invalid bboxes

### Next Steps
- [ ] review
- [ ] merge & release new carmen version

cc @mapbox/geocoding-gang

### DIFF
--- a/scripts/carmen.js
+++ b/scripts/carmen.js
@@ -75,13 +75,13 @@ if (argv.proximity) {
 }
 
 if (argv.bbox) {
-    if (argv.bbox.split(',') !== 4)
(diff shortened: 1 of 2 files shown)


mattficke:
🤦‍♂️ thanks for catching the typo and adding a test @lizziegooding

PR #684, opened 2018-01-25
Token concat

ingalls:
### Context

- Updates to token tests
    - Refactor to avoid use of anon fxn
    - Ensure all assertions are `result, expected`
    - Add tests around per-index regex tokens (sweden)

*Bug Fix*
On the first pass of generating a `from:to` pair, the original `to` string would be mutated to include the numbered replacement groups referencing the word boundaries.

So `$1something` would be mutated to become `$1$2something$3`

If the diacritic removal code was run it then removed the diacritics from the `from` replacer and ran the `to` creator again. Since the `to` had been mutated by the first pass it would then incorrectly calculate the numbered replace groups. 

Fix: Avoid mutating the original `to` text, instead cloning on write.

cc @mapbox/geocoding-gang

### DIFF
--- a/test/token.test.js
+++ b/test/token.test.js
@@ -37,6 +37,25 @@ tape.test('token#concatenated single token', t => {
     t.end();
 });
 
+tape.test('token#concatenated single token - diacritics', t => {
+    let tokens = {
+        '([a-z]+)vägen': {
+            'text': '$1v'
+        }
+    }
+    let tokenReplacer = tokenize.createReplacer(tokens)
+    let expected = [
+        { from: /([\s\u2000-\u206F\u2E00-\u2E7F\\'!"#$%&()*+,\-.\/:;<=>?@\[\]^_`{|}~]|^)([a-z]+)vägen([\s\u2000-\u206F\u2E00-\u2E7F\\'!"#$%&()*+,\-.\/:;<=>?@\[\]^_`{|}~]|$)/gi, inverse: false, named: false, to: '$1$2v$3' },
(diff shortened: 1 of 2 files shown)


test/token.test.js
```diff
@@ -1,57 +1,83 @@
...
+    }
+    let tokenReplacer = tokenize.createReplacer(tokens)
+    let expected = [
+        { from: /([\s\u2000-\u206F\u2E00-\u2E7F\\'!"#$%&()*+,\-.\/:;<=>?@\[\]^_`{|}~]|^)([a-z]+)vägen([\s\u2000-\u206F\u2E00-\u2E7F\\'!"#$%&()*+,\-.\/:;<=>?@\[\]^_`{|}~]|$)/gi, inverse: false, named: false, to: '$1$2v$3' },
+        { from: /([\s\u2000-\u206F\u2E00-\u2E7F\\'!"#$%&()*+,\-.\/:;<=>?@\[\]^_`{|}~]|^)([a-z]+)vagen([\s\u2000-\u206F\u2E00-\u2E7F\\'!"#$%&()*+,\-.\/:;<=>?@\[\]^_`{|}~]|$)/gi, inverse: false, named: false, to: '$1$2v$3' }
+    ]
+
+    t.deepEquals(tokenReplacer, expected, 'created a regex')
+
+    t.equals(tokenize.replaceToken(tokenReplacer, 'Samuelsvägen'), 'Samuelsv', 'replaced token');
```
aarthykc: great catch!

PR #685, opened 2018-02-05
Waffle Matching Places

ingalls:
### Context

A customer requested the ability to have more insight into how synonyms are surfaced.
Ref: https://github.com/ingalls/pt2itp/pull/313

@apendleton rightly pointed out we already do this, but practical testing of the PT2ITP backed indexes showed the `matching_*` tags were not being output.

This PR adds support for waffle indexed features as a fallback.

cc @mapbox/geocoding-gang

### DIFF
--- a/lib/util/ops.js
+++ b/lib/util/ops.js
@@ -1,8 +1,9 @@
 module.exports.toFeature = toFeature;
-var closestLang = require('./closest-lang');
-var featureMatchesLanguage = require('./filter').featureMatchesLanguage;
-var termops = require('./termops');
-var bbox = require('./bbox');
+
+const closestLang = require('./closest-lang');
+const featureMatchesLanguage = require('./filter').featureMatchesLanguage;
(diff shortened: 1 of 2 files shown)


lib/util/ops.js
```diff
@@ -218,64 +219,77 @@ function toFeature(context, format, languages, languageMode, debug, geocoder, cl
...
         if (regex.exec(candidate)) {
-            for (var phraseKey of allPhrases.get(candidate)) {
+            for (let phraseKey of allPhrases.get(candidate)) {
                 if (!matches.hasOwnProperty(phraseKey)) matches[phraseKey] = candidate;
             }
         }
     }
 
-    var best;
+    //Fallback Behavior to match addresses ie: 12## Main st
```
apendleton: Just noting for posterity that this is the operative section of the PR and most of the rest of it is ES6-ifying.

PR #687, opened 2018-02-07
Reduce index size

aarthykc:
### Context
Reducing index size by dropping interpolation for features that have long addressnumber ranges. For example, in France, we noticed addressnumber array that looked something like `[1,2,3....5000]`. This is causing our indexes to be bulky and hence, the approach to reduce the index size is to drop interpolation for features where outliers like 5000 are detected. 

#### Outlier detection
The outlier detection formula used is [Tuckey's fences](https://en.wikipedia.org/wiki/Outlier#Tukey's_fences) where we determine the lower limit and an upper limit in a sorted arrray (quartiles - q1 and q3) below and above which (respectively) the address numbers are counted as outliers. 

#### Dropping interpolation
The way to drop interpolation is by setting these values:
```
carmen:rangetype	
carmen:lfromhn	
carmen:ltohn	
carmen:rfromhn	
carmen:rtohn	
carmen:parityl	
carmen:parityr
```
To null, which would mean that the address cluster is not interpolated.

### Summary of Changes
- [x] adds outlier detection to indexdocs
- [x] drops the interpolation in standardize function in indexdocs

### Next Steps
- [x] Reviewed + feedback from @ingalls || @apendleton 
- [x] test on indexes and see changes
- [x] merge, release

cc @mapbox/geocoding-gang


lib/indexer/indexdocs.js
```diff
@@ -213,6 +214,24 @@ function standardize(doc, zoom) {
         });
     }
 
+
+// if an outlier is detected in address numbers for example in [1,2,3,5000], 5000 is considered an outlier, then drop interpolation for it
+    const outlier = isOutlierDetected(doc.properties['carmen:addressnumber']);
+        if (outlier) {
```
ingalls: The identation is a bit wonky here - could this become 

```
if (isOutlierDetected(doc.properties['carmen:addressnumber'])) {
```

lib/indexer/indexdocs.js
```diff
@@ -213,6 +214,24 @@ function standardize(doc, zoom) {
...
+            console.warn('address cluster contains addressnumbers outside the threshold, drop interpolation for address cluster');
+            var interpolationProperties = ['carmen:lfromhn','carmen:ltohn', 'carmen:parityr', 'carmen:rfromhn','carmen:rtohn', 'carmen:parityl'];
+
+//set interpolation properties values to null, for example: "carmen:parityr":[["O", "O" ,null ,null ,null], null] would become "carmen:parityr":[[null, null ,null ,null ,null], null]
+            interpolationProperties.forEach(function(p) {
```
ingalls: We're going to be enforcing arrow functions in the next week or so, want to swich this to 

```
forEach(p => {
```

apendleton: Actually, pretty sure our lint rules require parens even for single argument. So I think it should be:

    forEach((p) => {

but we can confirm.

ingalls: 👍

lib/indexer/indexdocs.js
```diff
@@ -138,11 +134,11 @@ function runChecks(doc) {
     } else if (geojsonErr.length) {
         for (var err_it = 0; err_it < geojsonErr.length; err_it++) {
             if (geojsonErr[err_it].level == 'message') {
-                //It's good practice to follow these warnings but required
+            //It's good practice to follow these warnings but required
```
ingalls: *not required

ingalls: * not required

I think you meant to say not required here

aarthykc: oh i didn't write that, but can make the change, only changed the indentation!

PR #689, opened 2018-02-13
Prefer non-ghost features when building context

mattficke:
### Context
"Ghost" features (typically alternate place names, denoted by a score of
-1) are excluded from a feature's context array unless they were
included in the original query. This allows an alternate place name to
returned if it was explicitly requested, but otherwise the default place
name will be used.

However, in situations where a query token matches both a ghost and a
non-ghost feature (for instance, a POI with the same name as a ghost
place within which it is located), we will return the first feature that
happens to be returned when querying the vector tile, which is
arbitrary and non-deterministic between indexer runs.

### Summary of Changes

This adjusts the context builder to use a priority order:

1. Non-ghost features that match a query token
2. Ghost features that match a query token
3. Non-ghost features that do not match a query token

### Next Steps
There is still some non-determinism when matching ghost features, since
this preserves a portion of the old behavior and returns only the first
ghost feature that is also a forward phrasematch. However, this is a
smaller problem than the prior status quo, and doesn't have as obvious a
solution since all features in this category are identically scored.

cc @mapbox/geocoding-gang

### DIFF
--- a/test/geocode-unit.ghost.test.js
+++ b/test/geocode-unit.ghost.test.js
@@ -1,0 +1,106 @@
+const tape = require('tape');
+const Carmen = require('..');
+const context = require('../lib/context');
+const mem = require('../lib/api-mem');
+const queue = require('d3-queue').queue;
+const addFeature = require('../lib/util/addfeature'),
+    queueFeature = addFeature.queueFeature,
+    buildQueued = addFeature.buildQueued;
+(() => {
+    const conf = {
+        region: new mem(null, () => {}),
+        city: new mem(null, () => {}),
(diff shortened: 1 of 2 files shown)


test/geocode-unit.ghost.test.js
```diff
@@ -0,0 +1,106 @@
...
+            }
+        };
+        queueFeature(conf.city, city, t.end);
+    });
+
+    tape('index city 2', (t) => {
+        let city = {
+            id:3,
+            properties: {
+                'carmen:text':'Tatooine',
```
aarthykc: Haha nice 👌

PR #693, opened 2018-03-10
Routablequest

Nmargolis:
### Context

This is a work in progress for adding routable point functionality for forward geocoding of addresses. Since address points currently are either a parcel centroid, building centroid, or possibly a building entrance, they may be significantly offset from the road. The new routable_points property in the geocoding results will contain point(s) on the nearby relevant road.

This approach takes advantage of the fact that address geojson features already have the associated road geometry as a result of the interpolation process, and uses that to find the nearest point on the relevant road.

See previous progress in #678 

**For example:**
Routable point example
 A request for the address on U Street shown here (red marker) will return a routable_points array with the point on the correct road (green marker), and NOT on the _technically closer_ alley behind the address.

### Summary of Changes

- [ ] Use Turf.js nearest-point-on-line to find the routable point
- [ ] Add routable_points attribute to geocoding results for forward geocoding
- [ ] Add unit tests and integration tests for routable point functionality

➡️ **carmen-geojson.md**

- Documents the new routable_points property as an array of points

➡️ **lib/pure/routablePoint.js**

- Accepts a point object or coordinate array and an address feature, and returns a routable point coordinate array for correctly formed inputs, and null for incorrect inputs
- Checks to make sure inputs are correct and routable_point-worthy. ⚠️ routable_point worthiness is defined here as non-interpolated address points (since interpolated points are already on the roads used as interpolation lines), and features that don’t already have routable_points pre-defined.
- Uses Turf’s nearest-point-on-line on the address point and the address feature’s  (Multi)LineString geometry

➡️ **lib/util/ops.js**

- Add routable_points (⚠️ plural, array for now) to the returned feature

➡️ **lib/verifymatch.js**

- For forward geocoding of addresses, call the routablePoint functionality

➡️ **test/geocode-unit.address-routable-point.test.js**

- Tests that a forward geocode of a non-interpolated address returns the expected routable_points
- Tests that a forward geocode of an interpolated address does not return routable_points
- Tests that a forward geocode of a feature without a LineString does not return routable_points

➡️ **test/routablepoint.test.js**

- Tests that features without LineStrings return null
- Tests that empty point inputs return null
- Tests the following cases: (some of these are not realistic since we’re skipping routablePoint for interpolated results, so may not be worth testing 🤷‍♀️  )
```
    Example point (x) feature point coords (.) and linestring coords (-):
    Address is one of the known address points
     .   x   .
     - - - - -
    
    Point already on the linestring should return itself (unless its interpolated) 
    .   .   .
    - - x - -
    
    Point in between points on the linestring
         x
     .   .   .
     - -   - -
     
     Point on a diagonal to a zigzag line
              x
      .        .
     /  \ /  \
         .
    
    Point equidistant from two sides of a cul-de-sac
      |         |
       <-- x
      | _ _ _ _ |
      
    Point closer to one side of a cul-de-sac
      |         |
            x -->
      | _ _ _ _ |
      
      Point closer to different side of cul-de-sac
      |          |
      |          |
      |      x   |
        _ _ _ _ _
```

### Next Steps

- [ ] Address the TODOs in the code
- [ ] Add more test cases. Specifically, test non-addresses, and update routablePoint.test.js with more realistic features (that have the `carmen:types` attribute
- [ ] Add benchmarks to test performance
- [ ] ⚠️ Confirm decision that a routable_points array in the results makes sense to future-proof for eventually allowing features with multiple routable points
- [ ] ⚠️ Confirm expected behavior for interpolated points (which are already on the road/interpolation line)
- [ ] ⚠️ Potentially add a distance threshold to ensure results aren’t far from the original point
- [ ] ⚠️ Determine what routable_points set at index time could look like, in order to properly check for it in routablePoint.js
- [ ] ⚠️ Add reverse geocoding functionality

** ⚠️ flag means this requires larger discussion

cc @mapbox/geocoding-gang


lib/geocoder/ops.js
```diff
@@ -114,9 +114,12 @@ function getPlaceName(context, formatString, language, languageMode, matched) {
  * @param {String} language code to select language/formatting from context array. Undefined defaults to `carmen:text`
  * @param {String} languageMode option that can be set to 'strict' filter resutlts strictly to a specific language. Undefined defaults null.
  * @param {Boolean} debug Enable debug mode to expose internal properties in results
+ * @param {Object} geocoder Geocoder object
+ * @param {boolean} clipBBox Whether or not to clip BBox
+ * @param {boolean} routingMode Whether or not geocoding request was made for routing
```
apendleton: Small nit: that this is a boolean is surprising to me, given that the other `_Mode` things are strings. "What's the language mode? Strict. What's the routing mode? Yes." I think the intention is that we're *in* routing mode, or not, but that uses the term differently than we do for the others (language and reverse).

lib/geocoder/routablepoint.js
```diff
@@ -0,0 +1,85 @@
...
+
+    const nearestPoint = featureLineString ? nearestPointOnLine(featureLineString, point) : null;
+
+    if (!nearestPoint) {
+        return defaultResult;
+    }
+
+    // Round coordinates to 6 decimal places
+    const nearestPointCoords = nearestPoint.geometry.coordinates.map(
+        (coord) => Math.round(coord * Math.pow(10, 6)) / Math.pow(10, 6)
```
apendleton: If we're doing this a lot, we should calculate `Math.pow(10, 6)` once and reuse it. Here, we're calculating it not once but twice, for each element in the `map` iteration.

lib/geocoder/verifymatch.js
```diff
@@ -154,6 +163,11 @@ function verifyFeatures(query, geocoder, spatialmatches, loaded, options) {
             } else {
                 feats[0].properties['carmen:address'] = null;
             }
+        } else if (options.routingMode) {
+            feats.forEach((feat) => {
+                const routPoints = routablePoints(feat.properties['carmen:center'], feat);
+                feat.routable_points = routPoints;
+            });
```
apendleton: So if I'm reading this right, if routingMode is set, we're going to try calculating a routable point for every feature, even if it's, like, a country? Does this make sense?

apendleton: To expand here per voice: I think it makes sense for there to be a semantic distinction between "this is the kind of thing that can have routable points, but we failed to find one in this instance" vs. "this is the kind of thing that doesn't have routable points." Having the field in the latter case strikes me as unexpected, and as a general matter, we don't add every possibly property a feature could conceivably have on every feature, set to null. We don't have null phone numbers on countries, either.

lib/geocoder/ops.js
Nmargolis: Ah, that does seem confusing. What if it was just `routing`? I do still think it makes sense for it to be a boolean, similar to `autocomplete`

lib/geocoder/verifymatch.js
Nmargolis: Updated to only add `routable_points` if the `geocoder_routable` flag is set on the tilejson for the source of a feature (currently addresses and POIs)

Issue #697, opened 2018-03-16
Istanbul is deprecated and seems to not support some es6


Nmargolis:
### Context

Carmen currently uses [Istanbul](https://github.com/gotwarlost/istanbul) to run code coverage as part of the CI process.

Istanbul seems to be deprecated as of 2 years ago, and doesn't support some es6, specifically destructuring. It [errored out](https://circleci.com/gh/mapbox/carmen/2568) on the latest push of routablequest.

![image](https://user-images.githubusercontent.com/5750602/37547203-7221a66c-292d-11e8-9039-004717ea9265.png)

The offending line is ```const { geometry: { geometries: featureGeom } = {} } = feature;```
which may not be the easiest to read, but shouldn't cause an error.

### Next Steps

The Istanbul readme links to a "version 2.0" called [nyc](https://istanbul.js.org/).

I've tested nyc locally by npm installing nyc and changing `yarn coverage` to `TESTING=true nyc tape test/*.js  && codecov` and it seems to work AND output this nifty summary 😍 :

![image](https://user-images.githubusercontent.com/5750602/37548347-68a7911c-2934-11e8-8f85-4242f9662a98.png)

Not sure how this will work with Codecov but seems promising!

- [x] Create PR with a switch to nyc
- [x] Test if the CI process and Codecov work

PR #691, opened 2018-02-16
Project Windex: docs edition

boblannon:
### Context

This PR is focused on updating and improving source code documentation. The goal is total coverage of the source as it currently exists.

JSDoc comment blocks throughout the codebase to make carmen's source code more approachable. Carmen is very complex, and the goal is to make it easier to understand the context for a given function, with some indication of typical usage.

In addition, the [documentation.js](http://documentation.js.org/) library will be used to generate a markdown version of the parts of the carmen API that are public. We're taking "public" to mean the parts of the API that are typically engaged with when using carmen as a geocoding library (rather than developing on carmen itself).

### Next Steps

@boblannon
- [x] documentation.js setup
- [x] add "build-docs" command
- [x] Indexer js docs
- [ ] Text-processing js docs

@aarthykc 
- [x] geocoder js docs
- [x] Bin js docs
- [x] Template for tests in the test/README.md

cc @mapbox/geocoding-gang

PR #698, opened 2018-03-17
Update Istanbul to nyc for code coverage

Nmargolis:
### Context
#697 
tl;dr istanbul is deprecated and breaks on some features of es6, like destructuring.
nyc is the maintained version 2.0 of istanbul

### Summary of Changes
- [x] Remove istanbul as a dependency
- [x] Add nyc as a dependency
- [x] Update coverage script to use nyc
- [x] Ignore the auto-generated nyc output

### Next Steps

- [x] Get insight into how we use Codecov to make sure this doesn't break anything

cc @mapbox/geocoding-gang


boblannon:
I'm not sure this is working properly.  for whatever reason, the checks for codecov.io aren't being triggered on this PR.  compare with #691:

2018-03-22 at 2 15 pm

Looking at [the codecov dashboard](https://codecov.io/gh/mapbox/carmen/commit/5ef23361d7904002d146eed07cd444c1902f61d1), I think it's having trouble parsing what you're giving it:

2018-03-22 at 2 16 pm

i think we're going to want something more similar to the example that codecov [has in their docs](https://github.com/codecov/example-node#nyc):

```
nyc report --reporter=text-lcov > coverage.lcov && codecov",
```

boblannon:
looking at the nyc docs, it seems like you might just be able to add the `--reporter` flag to the `nyc` invocation itself, so you'd have 

```
TESTING=true nyc --reporter=lcov --reporter=text-lcov tape test/*.js && codecov
```

Nmargolis:
thanks! I'm going to try pushing that up and seeing if it works.

boblannon:
2018-03-26 at 9 41 pm

overiding to merge.  the difference here seems related to how istanbul and nyc calculate coverage.

PR #699, opened 2018-03-19
changed the Croatian fallback to Serbian latin from Serbian

Fa7C0n:
### Context

Croatian script is similar to Latin script, so instead of returning Cyrillic Serbian `sr`  it makes more sense to return Serbian with latin script `sr_Latn`.

### Summary of Changes
- [ ] changed the fall back to `sr_Latn`

cc @mapbox/geocoding-gang


mattficke:
Looks good. There's a small tweak that needs to be made to the digraphic exclusion test (https://github.com/mapbox/carmen/blob/master/test/geocode-unit.languageMode.test.js#L280-L398) to get it passing, and I think it would be useful to add another similar test.

Background: this test was added in https://github.com/mapbox/carmen/pull/596, and ensures that `sr` results are excluded when the `languageMode=strict` param is used, since the script family is ambiguous. This was needed because we consider the language-only portion of a full language tag (e.g, the `en` part of `en_US`, or `sr` in `sr_Latn`) as an acceptable fallback in strict `languageMode`. This is almost always a safe bet, except in the case of Serbian.

To make sure we're still properly testing the behavior of the `languageMode` parameter, can you add an additional test that's essentially the same as the existing test I link to above, except without a `carmen:text_sr_Latn` property for the `United States` feature? In this case, we'll want to make sure that the `United States` feature is still being stripped out, rather than falling back to English.

test/geocode-unit.languageMode.test.js
```diff
@@ -364,6 +364,125 @@ const addFeature = require('../lib/util/addfeature'),
...
+        });
+    });
+
+    tape('teardown', (t) => {
+        context.getTile.cache.reset();
+        t.end();
+    });
+})();
+
+// digraphic exclusion test without sr_Latn fallback in United States
```
Fa7C0n: Added this test, based on this comment - https://github.com/mapbox/carmen/pull/699#issuecomment-374248050

PR #702, opened 2018-03-27
Strip out acute accent character ( ́  ) from all the queries 

bsrinivasa:
### Context

Based on user feedback, queries for Moscow in Russian labels doesn't work when `Москва́`(with the accent character) is used instead of `Москва`. Based on discussion,  removing the character from all the queries with this character is a good solution Adding a line to strip out the accent character from all the queries. 

Used [this](https://www.fileformat.info/info/unicode/char/0301/index.htm) to refer for the unicode encoding to replace the accent character with null.

### Summary of Changes
- [ ] Added a `.replace` statement to strip out the character

### Next Steps
- Test
- Review and merge 
- Bump version

cc @mattficke  @mapbox/geocoding-gang

### DIFF
--- a/lib/util/termops.js
+++ b/lib/util/termops.js
@@ -110,6 +110,8 @@ function tokenize(query, lonlat) {
         // all other ascii and unicode punctuation except '-' per
         // http://stackoverflow.com/questions/4328500 split terms
         .replace(/[\u2000-\u206F\u2E00-\u2E7F\\'!"#\$%&\(\)\*\+,\.\/:;<=>\?@\[\]\^_`\{\|\}~]/gi, ' ')
+        // removes the acute accent character  ́
+        .replace(/[\u0301]/g,'')
         .split(/[\s+]+/gi);
 
     let pretokens = [];


mattficke:
@bsrinivasa looks good, can you add tests demonstrating that this works? I think we'll need two, you can add them to the existing test files:

1. A termops.tokenize.test.js unit test to test that the accent is stripped out
2. A geocode-unit acceptance test (probably in geocode-unit.unicode.test.js) that tests that a query for `Москва́` returns `Москва`

apendleton:
Several concerns:
* we already have a diacritic removing mechanism, here: https://github.com/mapbox/carmen/blob/master/lib/util/remove-diacritics.js . If it doesn't remove this diacritic, we should fix it, rather than adding a separate mechanism for this specific diacritic
* only performing this operation at query time and not at index time creates the potential to render some labels unsearchable (which the approach above addresses ^ as it's symmetric)
* ... but that potentially means we'd want to globally reindex in case this occurs in practice
* regardless of which approach we take, we introduce the possibility of rendering a non-empty string empty (if it's comprised only of this character), which the current diacritical mark removal operation explicitly does not allow; we need to test that case thoroughly, because normalizations of this kind were what caused the "emojibomb" crasher we used to have in Carmen

apendleton:
Thinking about this more on the way home, I think the way we should solve the empty-string problem is to only strip the combining diacritical mark if it follows a printable character, and not strip combining marks that stand alone (either as the only character in a query or the only character in a token). Also, from a maintainability/non-hackyness perspective, we should handle combining diacritical marks generally, rather than just the acute accent. They [exist in a single Unicode block](https://en.wikipedia.org/wiki/Combining_Diacritical_Marks), so this should be straightforward.

(We could also consider ditching our current diacritical mark removal code altogether, and replacing it with a [unicode normalization step](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/normalize) that decomposes all precomposed characters with diacritical marks into `(unmarked character) + (combining diacritical mark)` pairs, and then stripping them all in one go, but I'm less certain about whether or not this is a good ideal/whether it would work properly for the set of languages we currently support -- cc @boblannon for thoughts -- but this is probably a stretch goal that we can punt on for now.)

So, proposal: instead of the termops, change, we add an extra `replace` step after the main one in `removeDiacritics` here: https://github.com/mapbox/carmen/blob/master/lib/util/remove-diacritics.js#L35 that looks like: `.replace(/(\S)([\u0301-\u036F]+)/g, '$1')`. In other words, "match a non-whitespace character followed by one or more characters in the combining diacritical mark range, and replace them with just the first character."

And then, in addition to the tests @mattficke proposes, we also test:
* what happens when one combining diacritical mark makes up the entirety of a query
* what happens when one combining diacritical mark makes up the entirety of a token in a query that contains normal tokens
* ditto for those two if it's multiple combining diacritical marks in sequence
* what happens when multiple combining diacritical marks are applied to the same character
* what happens when one or more combining diacritical marks is applied to a character that already has a precomposed diacritical mark (e.g., "é").
* what happens when a combining diacritical mark character applies to a non-whitespace token-boundary character, like punctuation

This will almost definitely be impactful enough that we'll want to think about a reindex, which I realize expands the scope of your plans. In the interests of counterbalancing my desire to not introduce one-off isolated quick fixes into carmen with your interest in getting something out the door, you could also consider skipping the carmen work altogether for now and addressing your feature of interest with a synonym on the data side. Diacritics are due for an overhaul later this year anyway most likely, so it could wait if you'd prefer.

boblannon:
> We could also consider ditching our current diacritical mark removal code altogether, and replacing it with a unicode normalization step that decomposes all precomposed characters with diacritical marks into (unmarked character) + (combining diacritical mark) pairs, and then stripping them all in one go

you'd still have the problem of the "pre-composed" versions of accented characters, meaning we'd still have to map at least some of the accented characters.

bsrinivasa:
Thank you everyone for your thoughts. 🙂 
Holding on this change here. No next actions right now here.

PR #705, opened 2018-04-06
[WIP] Be explicit

Nmargolis:
### Context
The intention of this PR is to add an option to enable routable point functionality to release in beta without disturbing existing API users, and to make the response more explicit (return `null` for the points instead of returning nothing at all, if routable points are requested

### Summary of Changes
- [x] Adds routingMode option to geocode, and only returns routable_points if true
- [x] Update signature of response to allow for adding more info in the future
- [x] Update documention

### Next Steps
- [ ] Update api-geocoder to handle new option

cc @mapbox/geocoding-gang

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,4 @@ npm-debug.log
 /coverage
 /.nyc_output/
 coverage.lcov
+yarn-error.log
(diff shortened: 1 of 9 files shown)


Nmargolis:
Update:

Summary of Changes

PR #716, opened 2018-06-20
[fuzzy-phrase] remove normalization cache

boblannon:
### Context

This PR removes the normalization cache. The immediate objective is to simplify the `DawgCache` wrapper so that it can be more easily replaced with components from node-fuzzy-phrase.

### History

A brief history on the normalization cache's origins and function:

#### Fixing autocomplete

Originally, we identified an issue where token replacement was an impediment to autocomplete. At index time, "main street" would get token-replaced to "main st". This is fine for a straightforward query including "main street" because it would also undergo token-replacement and be converted to "main st". However, in an autocomplete setting, a user would get no completions for `main stre`, because the index only contained the abbreviated "main st".

The strategy was to add both the abbreviated and non-abbreviated versions of the string at index time, which allows the autocomplete to match.

A simplified example: say we have the phrase `"north main street"` and the token replacements `{"north": "n", "street": "st"}`. At index time, we would index all of the versions of that phrase obtained by applying those replacement mappings:

```
n main st
north main st
n main street
north main street
```

The dawg-cache's performance isn't impacted much by the addition of new entries, because of the efficiency of graph structures.

There would be significant cost introduced, though, because the grid cache would get bigger: each variant would have to be stored separately there. The growth could be huge in cases where a feature's text had multiple replacement opportunities. This was limited by:

- only considering three or fewer replacement opportunities.
- considering only the first 8 phrases generated by taking the product of the token replacements
 
However, this was an incomplete solution because there would be cases where not all tokens were being represented in their full and abbreviated forms. No matter what, though, we were guaranteed not to drop the version of the phrase that contained the fully-abbreviated versions of each token. 

Example: say we have the phrase "North Newtown Square Street Road" and the following replacement rules:

```
{ 
    "north": "n",
    "road": "rd",
    "square": "sq"
    "street": "st",
}
```

This would potentially net us `2**4 = 16` different phrases In that case, we would ignore the fourth rule and therefore not include any of the versions of the query that included the un-abbreviated "street". We would, however, include the fully abbreviated version `"n newtown sq st rd"`
 
#### Allowing more combinations with a normalization cache

Ignoring some token replacements was less than a complete solution. Considering all replacements could get very expensive in some contexts, though: cases like german umlaut variations (eg `ü -> ue`) could end up producing lots of variations, since they might happen in many words, and even multiple times in one word.

```
Möbelträger
Möbeltraeger
Moebelträger
Moebeltraeger
```

Nevertheless, in some contexts (German being a motivating example) it's important to make sure all variants are available for autocomplete matches. The normalization cache, added in #652, was intended to make that possible, while avoiding an explosion in the size of the grid cache.

The strategy was to add all possible variants of a phrase in the dawg-cache, but maintain a normalization cache that mapped from most of those variants to a single, canonical phrase. That canonical phrase would be stored in the grid cache.  chooses the first variant of a list of possible variants.

```
canonical, stored in dawg-cache and grid-cache: 
  "n main st"

variants, stored in the dawg-cache and also mapped to the canonical via normalization cache: 
"north main st"
"n main street"
"north main street"
```

So at query time, a query like "north main stre" would match the dawg cache, and then get mapped to the grid-cache entry associated with "n main st".

### Summary of Changes
- [ ]

### Next Steps

cc @mapbox/geocoding-gang


aarthykc:
TKTKTKTK

PR #717, opened 2018-06-26
Replace dawg cache with fuzzyPhraseSet builder

aarthykc:
### Context
TKTKTK

### Summary of Changes
- [ ] TKTKTK

### Next Steps

cc @mapbox/geocoding-gang


boblannon:
Got a bunch more tests passing, but I've run into a stumper.  It seems to have something to do with the `'lazy'` mode that `geocode-unit.translation-noauto.test.js` runs.  Not sure why this would be impacted by our changes, though...

```
/Users/bob/dev/carmen/test/acceptance/geocode-unit.translation-noauto.test.js:58
                cache[type].pack(rocksdb);
                            ^

TypeError: rocksdb file is already loaded read-only; unload first
    at TypeError (native)
    at forEach (/Users/bob/dev/carmen/test/acceptance/geocode-unit.translation-noauto.test.js:58:29)
    at Array.forEach (native)
    at Test.tape (/Users/bob/dev/carmen/test/acceptance/geocode-unit.translation-noauto.test.js:55:30)
    at Test.bound [as _cb] (/Users/bob/dev/carmen/node_modules/tape/lib/test.js:76:32)
    at Test.run (/Users/bob/dev/carmen/node_modules/tape/lib/test.js:95:10)
    at Test.bound [as run] (/Users/bob/dev/carmen/node_modules/tape/lib/test.js:76:32)
    at Immediate.next (/Users/bob/dev/carmen/node_modules/tape/lib/results.js:71:15)
    at runCallback (timers.js:672:20)
    at tryOnImmediate (timers.js:645:5)
```

apendleton:
@boblannon the purpose of this particular part of this test is to confirm that this behavior works for both backends of carmen-cache (the in-memory one used in indexing, and the on-disk one used at runtime). It makes a new carmen at the start and puts some data in it, assuming that carmen won't find an already-present rocksdb file when it first starts and so will auto-select the memory backend, and tries querying that, then bakes the data (the `pack` operation) in that cache onto disk and reloads the data in carmen and tries it again.

What looks to be happening here is that when it tries to do that pack, it finds that that rocksdb file is already loaded someplace -- maybe when it was first creating the carmen, it ended up selecting the on-disk one instead of the memory one because it *was* already there when it shouldn't have been? We ensure the not-there-ness generally by selecting a random filename on each carmen instantiation if one isn't supplied, but somehow that isn't working here. It does look like the mechanism for doing that [has changed](https://github.com/mapbox/carmen/blob/add-fuzzy-phrase-set/index.js#L289-L290), and it's not clear just from looking at it why, so... that's where I'd look first.

apendleton:
I should say also: possible debugging strategies would be to console.log `cache[type]` at that failure point to see what it is (should be a memory cache, not a rocksdbcache), and `rocksdb` (which should be a filename of a thing that does not yet exist; it'd get created in pack, ideally). If those turn out not to be as expected next step is to figure out why not.

boblannon:
>  We ensure the not-there-ness generally by selecting a random filename on each carmen instantiation if one isn't supplied, but somehow that isn't working here. It does look like the mechanism for doing that has changed, and it's not clear just from looking at it why, so... that's where I'd look first.

ah okay, yeah, now `getBaseFilename()` caches its result after it's first returned. this was necessary to be able to instantiate a new 	`FuzzyPhraseSet` in `buildQueue`'s callback.

test/acceptance/geocode-unit.multiload.test.js
```diff
@@ -6,8 +6,10 @@ const mem = require('../../lib/sources/api-mem');
 const queue = require('d3-queue').queue;
 const { queueFeature, buildQueued } = require('../../lib/indexer/addfeature');
 
-const country = new mem(null, () => {});
-const conf = { country: country };
+const conf = {
+    country: new mem({ maxzoom: 6 }, () => {})
+};
+
```
boblannon: @apendleton this change (and the one below) allows this test to pass, but @aarthykc and I aren't really sure why. Therefore, we weren't sure whether it was a confound to the test itself.  Any ideas?  Before the change, we were failing this check:

```
t.equal(country._dictcache, a.indexes.country._dictcache
```

...because `country._dictcache` (the original source object's dictcache) remained unchanged even after `a.indexes.country._dictcache` (the clone's dictcache) was reassigned with a `new FuzzyPhraseSet`.

apendleton: Yes, I think that confounds the test. When does that reassignment occur?

boblannon: well, at first we had it happen in `buildQueue`, but we recently moved it to `store` in order to support a benchmark script that ran into difficulty because it wanted to instantiate a `Carmen`, build an new index and then immediately query the same `Carmen` (without using `queueFeature`/`buildQueue`.

That's where it happens now, per b82cede

boblannon: of course, what's frustrating is that the tests and benchmarks require this dual-natured `Carmen` instance, which is something we don't actually require in any real-world use case.  I know the long-term plan is to split concerns into indexer and geocoder, but I think that, in the short term, we're going to have to bring back something analogous to the `DawgCache`/`ReadCache`/`WriteCache` design.

apendleton: The benchmark is probably just something I missed in the initial round. It should be updated to use `buildQueue` and the transformation should happen there. Agreed that this is not a real-world usecase, but as long as we're gonna support it, we should support it all in one place, and have all test-ish (bench, etc.) stuff work the same way.

I'm disinclined to do the facade. We're going to have to fight this battle at some point, and it sounds like it's already mostly fought. If we backtrack now we'll have to start all over again when we actually do it for real, and the facade facilitates new code or tests being added that continue to depend on a behavior we know we no longer want to support. Is it your impression that there are lots of failures remaining of this kind, or is it a final few corner cases? If the former, I guess we do what we have to do get the thing out the door, but if the latter, I think we should press on and finish out the split.

boblannon:
Everything is passing in my local environmeent.  The tests are currently failing on circle because `node-fuzzy-phrase` needs a version bump and a new prebuilt binary.  Once those are available, tests should pass.  Should happen shortly after https://github.com/mapbox/node-fuzzy-phrase/pull/1 is closed.

PR #719, opened 2018-07-19
[wip] fuzzy search

apendleton:
### Context
This is an omnibus PR tracking the work to incorporate https://github.com/mapbox/fuzzy-phrase into carmen, replacing the current dawg-cache text index with a new mechanism that allows for approximate string matching. We're doing a bunch of other stuff while we're at it though.

### Summary of Changes
- [x] ditch dawg-cache
- [x] introduce dependency on node-fuzzy-phrase
- [x] node-fuzzy-phrase has no index merge operation, so we've eliminated the merge operation altogether from carmen along with all attendant tests
- [x] we've also eliminated the normalization cache mechanism, which was tied to dawg-cache's internal mechanics; for now, we're multiply indexing textual variants in all cases
- [x] node-fuzzy-phrase also supports node 8, and we've moved to a new release of carmen-cache that does as well, so this carmen branch works on node 8 (with the caveat that it still doesn't do so on circle; see https://github.com/mapbox/carmen-cache/issues/124)
- [x] there's been some miscellaneous cleanup as well, like es6-ifying the bench code

### Next Steps
- [x] carmen's current strategy for lookup up multiple subqueries within a query in the fuzzy-phrase backend is not algorithmically well suited to the way the new backend works; as of the landing of https://github.com/mapbox/fuzzy-phrase/pull/40 we have a couple of much more performant strategies we can use instead, and we should adapt phrasematch to take advantage of them
- [x] we should pin down opt-in vs. opt-out in terms of API shape for using the fuzzy search functionality
- [x] fuzzily matched items should incur a relevance penalty in matches; we get the necessary information back from the fuzzy-phrase query to be able to do this, but don't propagate it forward in the geocoding pipeline yet
- [x] figure out the deal with carmen-cache and node 8
- [x] move to released fuzzy-phrase and node-fuzzy-phrase

cc @mapbox/geocoding-gang


yhahn:
congrats!!!!

PR #722, opened 2018-07-21
Fuzzy multi-lookup in phrasematch

apendleton:
### Context
This branch advance the fuzzy matching work by moving to more efficient lookup strategies in the fuzzy phrase graph, plus other odds and ends.

### Summary of Changes
- [x] switch the default for the fuzzy option to true -- I'd be curious to know what the motivation for making it false was; maybe it was to match autocomplete, which is false by default, but in general I think the decision around autocomplete was a mistake -- having the defaults be different than our defaults in production means we don't test very well in our standard configuration, since most tests won't opt out of the defaults, and I'm quite sure we'll want this to be on by default in production
    - [x] a few tests relied on not matching things that were similar but not identical, so these ones broke with this change and I fixed them by explicitly disabling fuzzy matching
- [x] pretty much rewrite phrasematch, to move from the old model of calculating permutations JS side and looking them up one by one to a new model:
    - [x] for non-address indexes, where we're essentially just looking up windows within the query, do a single lookup passing in the whole query, using the windowed lookup mechanism in node-fuzzy-phrase, and then emulating the creation of bitmasks as would have been created by the permutation calculator on the responses
    - [x] for address indexes, where additional permutations are generated, generate them all using the old mechanism but look them all up at once with multi-lookup
    - [x] simplify the `variants` notion and instead just only have subqueries, all of which are formatted the same
- [x] fix some longstanding annoyances along the way, like eliminating a couple of functions from termops that did literally nothing (one was just `function encodeTerm(text) { return text; }`), and eliminate their corresponding tests
- [x] move to a newer version of node-fuzzy-phrase
- [x] commented out node 8 on circle for now, until the `illegal instruction` issue is resolved

### Next Steps
I may add the relevance-penalty work, which is next on my agenda, to this branch, or may do a new one. Will depend on how invasive that ends up being.

cc @mapbox/geocoding-gang


aarthykc:
@apendleton yeah, don't think we paid attention while setting fuzzy to false, pretty sure it was to match autocomplete.

>termops that did literally nothing (one was just function encodeTerm(text) { return text; }), and eliminate their corresponding tests 
👏 thank you! 

🌮 pending the node 8 fix!

PR #726, opened 2018-07-23
Simplifications to text processing for indexing

apendleton:
### Context
This branch makes a couple of changes meant to simplify index-time computation and speed up builds a bit. Nothing here is earth-shattering, but should be incrementally useful.

### Summary of Changes
- [x] only generate housenumber ranges once per document, rather than once per synonym and token replacement variant -- the properties used to decide what to generate here don't change along these axes and we were generating the same housenumber range over and over for no reason
- [x] introduce `termops.getMinimalIndexableText` with accompanying tests -- this is basically a forward port of what `getIndexableText` used to be before the introduction of language-aware text indexing and multiple indexing of token variants. In the course of indexing we end up calling `getIndexableText` twice, once at the beginning to generate a term frequency model which we later use to generate text variants of feature names with high-frequency words dropped from them, and again to actually index the text (using the term frequency model from the first run to decide which additional phrases to generate for each document in the second). Most of the complexity that has been added to this function over time has been to benefit the second run -- the first run pays no attention to languages, and doesn't benefit from multiple-indexing of variants (and in fact may suffer somewhat from it). Here we reintroduce the old way of performing this operation and use it on the frequency pass.

### Next Steps
Nope

cc @mapbox/geocoding-gang

### DIFF
--- a/lib/indexer/indexdocs.js
+++ b/lib/indexer/indexdocs.js
@@ -442,10 +442,10 @@ function generateFrequency(docs, replacer, globalTokens, maxScore) {
         // set max score
         freq['__MAX__'][0] = maxScore || Math.max(freq['__MAX__'][0], docs[i].properties['carmen:score'] || 0);
 
-        const texts = termops.getIndexableText(replacer, globalTokens, docs[i]);
+        const texts = termops.getMinimalIndexableText(replacer, globalTokens, docs[i]);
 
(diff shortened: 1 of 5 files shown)


apendleton:
Absent-mindedly opened this PR with broken tests. Should be fixed now.

aarthykc:
Very cool! 🌮

PR #727, opened 2018-07-25
[wip] more efficient stacking

apendleton:
### Context
Testing carmen@fuzzy in practice revealed the possibility that certain kinds of queries could crash stall out the server process, consume a ton of memory, and potentially crash the server process. This branch is a first attempt at fixing it, though it's not all the way there yet.

In carmen's first lookup step, `phrasematch`, we identify all the substrings of the query that can be found in any of the indexes. In our second phase, `spatialmatch`, we explore all the ways in which different matches from different indexes can go together (in other words, come from indexes that cover the same countries, and don't overlap in terms of tokens in the query they cover). Once this set has been narrowed down, we take the 30 best combinations, and we perform a more complex geospatial matching operation (`coalesce`) with them.

The fuzzy work has dramatically increased the number of phrasematches we identify, since we now identify exact as well as approximate matches. In spatialmatch, even though we truncate at 30 stacks, even the process of enumerating all the stacks in the first place (before truncation) can apparently be extremely computationally expensive if the number of phrasematches is very high. The upside, though, is that the only things that matter for determining stacking are mask (that is, which tokens are consumed) and relevance (a function of the number of tokens in a match and, now, the edit distance), so in circumstances where the same subquery produces many fuzzy matches, they'll all behave identically for purposes of stacking. This means we can deduplicate them along these lines, perform the stacking operation (which is combinatoric and expensive), truncate *that* to 30, and then re-expand it again along the combinations we previously collapsed (and re-truncate to 30 again).

This prevents the crashing, but is unfortunately still pretty slow, and also often produces bad results, so there's more work to be done.

### Summary of Changes
- [x] introduce new operations into spatialmatch, `collapseToArchetypes` and `expandFromArchetypes`, which collapse multiple matches from the same index with the same mask and weight together, and then expand them back apart again, respectively, and apply these in spatialmatch

### Next Steps
- [x] the 30-element cutoff seems to often not produce the results we want under fuzzy circumstances -- it seems fairly common in queries with lots of short tokens for their to be preliminary (pre-coalesce) matches with 100% relevance because every very-short token occurs in some index or other; they ultimately won't align spatially, but we don't know that yet, and they end up crowding out the correct (but fuzzy) results that would otherwise have won. In general we prefer first the most relevant stacks, then the shortest (that is, if it can achieve high relevance with fewer matches of more tokens apiece, that's better than lots of short matches) -- in the circumstance described above, we want the length to boost the fuzzy-matched item even though it's not of perfect relevance. What I'd like to try is calculating two relevance totals, unpenalized and penalized, and then sorting the stacks on the (unpenalized, length, penalized) tuple, so that the relevance penalty only matters when features are otherwise tied on both unpenalized relevance and length.
- [x] I'm pretty sure this still won't be enough, so the other thing I think we'll have to do is not try to fuzzy-match very short phrases at all. We had considered not attempting to fuzzy-match very short *words*, but this will be less restrictive than that; we'll continue to be able to correct short words as part of longer phrases, but phrases comprised of a single short word (<= 2 chars) will no longer be correctable. I'm pretty sure this will dramatically improve our performance since a ton of the noise we're seeing so far is from the combinatoric explosion brought about by these super-short matches.

cc @mapbox/geocoding-gang


apendleton:
Did both of those other things. Ended up with an exclusion rule that's essentially "don't allow edit_distance > 0 if the phrase is exactly one word long and either that original or corrected word is less than four characters long." So, getting rid of all the 2-3 char problem matches, hopefully.

apendleton:
This turns out to have worked much better than I initially thought because I made a silly typo: https://github.com/mapbox/carmen/pull/727/commits/e2503dffab7ca71ecf9f4b5c240e14d1dd90779e . The dangers of late night hotfixes...

apendleton:
@boblannon and I tried a couple strategies for further improvements on sorting to balance penalized and unpenalized relevances. I've since merged what I think will be my final attempt here at least for now: calculating a single composite score that incorporates all three signals (relevance, length, edit distance) into a single value and sorting on that, so that edit distance and length can compete on equal footing rather than having to explicitly prioritize one over the other.

PR #735, opened 2018-08-13
Error when carmen:text contains the word 'constructor'

samely:
### Context
`carmen-indexer` returns an error while setting tokens

```
indexableText[key].add(language);

^
Error: indexableText[key].add is not a function
```

@mattficke  helped to find that this error is happening because `constructor` is actually calling to the constructor function of a [`new Set()`](https://github.com/mapbox/carmen/blob/master/lib/text-processing/termops.js#L306).

### Summary of Changes
- [x] Add test for `carmen:text` that contains the word `constructor`.

@apendleton @mattficke 
cc @mapbox/geocoding-gang


samely:
@mattficke can you review it?

samely:
@apendleton could you review this? :pray:

lib/text-processing/termops.js
```diff
@@ -303,8 +303,8 @@ function getIndexableText(replacer, globalTokens, doc) {
 
         const key = tokens.join(' ');
 
-        indexableText[key] = indexableText[key] || new Set();
-        indexableText[key].add(language);
+        if (indexableText[key] instanceof Function || !indexableText[key]) indexableText[key] = new Set([language])
```
apendleton: Seems like this should have a semicolon at the end?

mattficke: @samely this overwrites the original `constructor` property of the object, reassigning inherited properties like this usually isn't a good idea.

Another option would be to replace `const indexableText = {};` above with `const indexableText = Object.create(null);`, which creates an object that doesn't inherit any properties from its prototype. I think this would be a one-liner that wouldn't require any other changes.

samely: Yes. It works :bowing_woman: Thank you! Can you review?

apendleton:
This still fails linting. Can you confirm that eslint passes before you merge?

samely:
@apendleton I would prefer to take a look at this in a diff PR :pray:

samely:
:bowing_woman: could you approve this PR @apendleton? Thank you!

PR #739, opened 2018-08-15
Switch to carmen-cache@reduce-x-language-penalty

apendleton:
### Context
This PR switches to carmen-cache@reduce-x-language-penalty, which decreases the penalty applied to out-of-language matches in coalesce from 10% to 4%, and update tests accordingly for new, higher relevance values.

The motivation was newfound interactions between the previous penalty and new, lower penalties introduced as part of the fuzzy search push. This allowed for circumstances where fuzzy matches would beat correctly spelled matches that were out of language, and in testing this seemed not to conform with user expectations.

### Summary of Changes
- [x] switch to carmen-cache@reduce-x-language-penalty
- [x] update tests with new relevance numbers

### Next Steps
- [x] test with upstream packages that depend on carmen
- [x] if test results look okay, merge mapbox/carmen-cache#128, publish a new binary, and cut a release to npm
- [x] switch to released carmen-cache
- [ ] merge this PR, and release

cc @mapbox/geocoding-gang

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,7 +6,7 @@
   "author": "Mapbox (https://www.mapbox.com)",
   "license": "BSD-2-Clause",
   "dependencies": {
-    "@mapbox/carmen-cache": "0.21.4",
+    "@mapbox/carmen-cache": "https://github.com/mapbox/carmen-cache/tarball/66407c9be35d599a4cb876e40a4daf06df293bcf",
     "@mapbox/geojsonhint": "^2.0.1",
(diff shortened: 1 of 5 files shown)


mattficke:
carmen-cache is merged. Hitting the `[publish binary]` race condition (possible fix ticketed here: https://github.com/mapbox/carmen-cache/issues/129) over there, but between both the commit build and the tag build everything passed, so I'm going to publish.

PR #741, opened 2018-08-21
Proximity test fix

lizziegooding:
### Context

This PR fixes proximity.scoredist.test.js unit tests to call the `scoredist` function with all required arguments. Previously, the tests only passed in the score and distance parameter, not the zoom or radius parameters, which meant the calculated `scoredist` value was always `NaN`. As part of that bug fix, we also moved the `meanScore` function, which calculates the geometric mean of carmen geojson features' `carmen:score`, to util/proximity.js and added unit tests.

### Summary of Changes
- [x] refactor proximity.scoredist.test.js tests to call `scoredist` function with zoom level and `PROXIMITY_RADIUS` constant
- [x] move `meanScore` function to util/proximity.js
- [x] write unit tests for `meanScore` function

### Next Steps
- [x] review by @yhahn 

cc @mapbox/geocoding-gang

### DIFF
--- a/lib/geocoder/verifymatch.js
+++ b/lib/geocoder/verifymatch.js
@@ -107,7 +107,6 @@ function verifymatch(query, stats, geocoder, matched, options, callback) {
 * @returns filtered - sorted and verified features
 */
 function verifyFeatures(query, geocoder, spatialmatches, loaded, options) {
-    let meanScore = 1;
     const result = [];
     let feats;
     for (let pos = 0; pos < loaded.length; pos++) {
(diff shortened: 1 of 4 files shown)


yhahn:
@lizziegooding :+1: good change and writeup explaining context for the update.

Would recommend staging this work prior to an official carmen release per our conversation around the `isNaN()` check and throw.

lizziegooding:
Successfully deployed to staging and verified downstream tests passed using this branch. Going to merge and release.

PR #742, opened 2018-08-22
Added #!/usr/bin/env node

pakastin:
Can't run .js without node hashbang


ingalls:
Thanks for the addition!

PR #743, opened 2018-08-23
[Experiment] OpenLocationCodes

ingalls:
### Context

Google has for the past few years been working on the OpenLocationCode spec. It's now slowly popping up across their mapping products.

### Summary of Changes
- [x] Add ability to do a OpenLocationCodes Lookup
- [ ] Include PlusCode in returned responses
- [ ] Unit Tests

### Next Steps
No concrete next steps, just a proof of concept to see what an integration would look like.

cc @mapbox/geocoding-gang

### DIFF
--- a/lib/geocoder/geocode.js
+++ b/lib/geocoder/geocode.js
@@ -148,7 +148,7 @@ module.exports = function(geocoder, query, options, callback) {
     if (pluscode.isValid(query)) {
         const code = pluscode.decode(query);
 
-        return reverseGeocode(geocoder, [ code.longitudeCenter, code.latitudeCenter ], options, callback);
+        return reverseGeocode(geocoder, [code.longitudeCenter, code.latitudeCenter], options, callback);
     }
 
     // Reverse geocode: lon,lat pair. Provide the context for this location.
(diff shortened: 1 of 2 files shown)


ingalls:
@missinglink Ya this was mostly just playing around with the concept, would require more work to get it into shape and hash out the final details of how best to implement.

We do return the converted coordinates in the `query` array so someone could use the API just to do the conversion in addition to the full geocode.

Also have to support queries like `<PlusCode>, place` which google seems to favour and is more difficult with our current infrastructure.

Adding plus codes to returned results however should be fairly trivial. Are you planning on integrating with Pelias?

missinglink:
honestly, we're still on the fence. I went to meet them in their Zurich office on Monday and they're nice people, I think their intentions are good.

I would prefer that PlusCodes were purely programmatic, supporting the locality lookup thing currently requires a proprietary data service, which, for obvious reasons, I'm not a big fan of :)

my ideal would be to put it all in the client lib, and we could give them the open data to do that, plus some src code similar to what https://github.com/pelias/placeholder does in order to do the lookup entirely client-side.

IMHO that would be a truely Open standard.

ingalls:
@missinglink yeah I was a little disappointed to see the `localcode + place` as well. But I understand that they are doing it to reduce the full code size and allow it to be memorable.

Like you said however - This just adds the complexity of place data and all the fuzzy/spelling problems that come with it. I am however very excited to see an open standard form, too many public entities already consider their address data closed. We don't need more closed address data in any form.

PR #729, opened 2018-07-31
Slightly penalize address matches with no housenumbers, and update te…

apendleton:
…st fixtures accordingly

### Context
This PR introduces a slight penalty for matches within address indexes that don't contain housenumbers. Given a match for `bakery`, it's more likely to be intended to refer to bakeries than `bakery street`.

### Summary of Changes
- [x] add penalty
- [x] update tests

### Next Steps
No

cc @mapbox/geocoding-gang

PR #750, opened 2018-10-12
Street fallback

ingalls:
### Context

If an address result get's to verify match but fails to be matched with a Point or Interpolated address - fallback to returning the street level result

### Problem

Addresses that don't match our point or ITP data do the following
```
Query:
123 Left Turn, Albuquerque, CA, USA

Response:
Albuquerque, CA, USA
```

We should investigate returning the street level result where possible instead as closer fallback to the user's intended feature

```
Query:
123 Left Turn, Albuquerque, CA, USA

Response:
Left Turn, Albuquerque, CA, USA
```

cc @mapbox/search


lib/geocoder/phrasematch.js
```diff
@@ -174,13 +174,6 @@ module.exports = function phrasematch(source, query, options, callback) {
             );
         }
 
-        // we expect addresses typically to have house numbers
-        // if this is an address but has no house number, penalize it with respect
-        // to other kinds of matches
-        if (source.geocoder_address && !/#/.test(phrase)) {
-            editMultiplier *= .99;
-        }
```
apendleton: I'm pretty nervous about cutting out this penalty. Adding it fixed a whole bunch of category test failures, where bogus address-index matches were pushing valid categories off the stack limit cliff (especially, "bakery" matching "bakery street" as a prefix match). See https://github.com/mapbox/carmen/pull/729 .

ingalls: Per voice, working on https://github.com/mapbox/carmen/pull/758 as a precursor to this change

ingalls:
This PR is unblocked now that https://github.com/mapbox/carmen/pull/758 is merged.

Issue #747, opened 2018-10-11
More heavily bias local results in verifymatch


lizziegooding:
### Context
When we calculate `scoredist` in verifymatch.js, the highest value we can return is "up to 100x times the geometric mean". To start, we should experiment with this max value/multiplier to determine an appropriate threshold that would more heavily bias local results. In the future, we should consider completely rewriting this algorithm to achieve the mixture of local/high relevance global results we want.
https://github.com/mapbox/carmen/blob/f831f2dfc9eb965752939f5809b18315b38b7659/lib/util/proximity.js#L69-L90

### Next Steps
- [ ] increase constant multiplier and update tests
- [ ] publish dev version
- [ ] see how it changes pass rates downstream

cc @miccolis @apendleton @aarthykc @boblannon

PR #752, opened 2018-10-12
More heavily bias local results

lizziegooding:
### Context
Ref https://github.com/mapbox/carmen/issues/747. To start, increase constant multiplier from 100 to 300 so that `scoredist` can be up to 300x the geometric mean score.

### Summary of Changes
- [x] increase constant multiplier in `scoredist` function to 300

### Next Steps
- [ ] publish a dev release
- [ ] run against downstream tests


lizziegooding:

Issue #748, opened 2018-10-11
Use a different curve to bias local results in verifymatch


lizziegooding:
### Context
We currently weigh features along an exponential curve based on their distance from a given proximity point, with 1 being on proximity point and 0 being at or beyond the proximity bias radius. This means that even if a user sets a very large proximity bias radius, say 500 miles, the affect of that bias extends for a much smaller geographic area. This bias is often so low as to fail to return local, relatively low score features. We should experiment with using a different curve, such as a linear or sine curve instead.

https://github.com/mapbox/carmen/blob/f831f2dfc9eb965752939f5809b18315b38b7659/lib/util/proximity.js#L69-L90

### Next Steps
- [ ] weight features along a linear curve
- [ ] publish dev version
- [ ] see how it changes pass rates downstream
- [ ] weight features along a sine curve
- [ ] publish dev version
- [ ] see how it changes pass rates downstream

cc @miccolis @apendleton @aarthykc @boblannon

lizziegooding:
Not going with this approach, moved to https://github.com/mapbox/carmen/pull/768

PR #753, opened 2018-10-12
Use a different curve to bias local results

lizziegooding:
### Context
Ref https://github.com/mapbox/carmen/issues/748. To start, weigh results along a sine curve when calculating `scoredist`.

### Summary of Changes
- [ ] weigh results along a sine curve

### Next Steps
- [ ] publish dev release
- [ ]  run against downstream tests


lizziegooding:
Capturing from chat, we may want to use a different, but similarly shaped curve rather than a sine curve, as `Math.sin` could negatively impact performance. @apendleton proposed using a hyperbola (`y = 1/x`) with limited max/min values.

lizziegooding:

Issue #746, opened 2018-10-10
Parameterize cutoffs


apendleton:
There are a bunch of places in carmen or carmen-cache where we arbitrarily cut off the number of things we consider for performance reasons. In general the cutoff amounts are hard-coded. Let’s expose them as configurable parameters with defaults, to get ready for letting Carmen consumers configure them and experiment with the tradeoffs of result quality vs. performance.

# Overall carmen changes
- Carmen’s constructor takes an `options` argument: https://github.com/mapbox/carmen/blob/master/index.js#L54 that right now we only use for global tokens. We’ll be adding more stuff to it. It may make sense to group all the things we’re adding into a child object, so, `options.cutoffs = {MAX_STACKS: 30, MAX_WHATEVER: 50}` etc. The Carmen constructor should save these cutoffs at construction time.
- All the cutoffs will affect phrasematch, spatialmatch, or verifymatch. Two of the three don’t have access to the whole geocoder object, but all three take an `options` parameter, so we’ll want to copy whatever cutoffs are set into the `options` parameter before each function is called within `forwardGeocode`:
  - phrasematch: https://github.com/mapbox/carmen/blob/master/lib/geocoder/geocode.js#L367
  - spatialmatch: https://github.com/mapbox/carmen/blob/master/lib/geocoder/geocode.js#L367
  - verifymatch: https://github.com/mapbox/carmen/blob/master/lib/geocoder/geocode.js#L413
- We should continue to use the current hardcoded values as defaults for any cutoffs we make configurable. We should **either:**
  - have the Carmen constructor check for the presence of every configurable value and set a default value instead, such that every carmen object has a fully populated set of configurable cutoffs (so, move all the default cutoff values into the Carmen constructor, and have *match count on them always being included in `options`), **or**
  - leave the cutoff values where they are, and have *match set them if they’re not in the options supplied
  I have a slight preference for the latter if only because it will involve changing less stuff, but I don’t really care as long as we’re consistent.

# Specific limits
### Maximum phrase length for fuzzy search
Number of words in a query beyond which we don’t try to do fuzzy matching.
**Currently:** 8
**Maybe change to?:** making this higher is bad for performance but matches more stuff
**Set/used on line:** https://github.com/mapbox/carmen/blob/master/lib/geocoder/phrasematch.js#L43

### Minimum word length for single-word fuzzy queries
Number of characters that must be in a single-word query for us to try fuzzy-matching it
**Currently:** 4
[...]

PR #755, opened 2018-10-15
[first cut] parameterized cutoffs

aarthykc:
### Context
Per #746, 
This allows users to toggle limits that have been currently hard coded for performance benefits. The defaults have been added to the `constants.js` file and are overriden if the user provides a value through the options object.

### Summary of Changes
- [ ] changes to `constants.js` to include the default values 
- [ ] changes to verify, spatial and phrase match use options when provided
- [ ] tests to illustrate differences when the cutoffs are provided vs when the default is used

### Next Steps
- [ ] Bench marks
- [ ] js docs for the new options in `geocode.js`

cc @mapbox/geocoding-gang

### DIFF
--- a/lib/constants.js
+++ b/lib/constants.js
@@ -9,12 +9,8 @@ module.exports = {
     PROXIMITY_RADIUS: 200,
     MAX_TEXT_SYNONYMS: 10,
     MIN_CORRECTION_LENGTH: 4,
-    // STACKABLE_LIMIT: https://github.com/mapbox/carmen/blob/master/lib/geocoder/spatialmatch.js#L277
     STACKABLE_LIMIT: 100,
-    // SPATIALMATCH_STACK_LIMIT: https://github.com/mapbox/carmen/blob/master/lib/geocoder/spatialmatch.js#L40
     SPATIALMATCH_STACK_LIMIT: 30,
-    // MAX_CORRECTION_LENGTH: https://github.com/mapbox/carmen/blob/master/lib/geocoder/phrasematch.js#L43
     MAX_CORRECTION_LENGTH: 8,
-    // VERIFYMATCH_STACK_LIMIT: https://github.com/mapbox/carmen/blob/master/lib/geocoder/verifymatch.js#L48
     VERIFYMATCH_STACK_LIMIT: 20
 };
--- a/lib/constants.js
+++ b/lib/constants.js
@@ -8,5 +8,13 @@ module.exports = {
     MAX_QUERY_TOKENS: 20,
     PROXIMITY_RADIUS: 200,
     MAX_TEXT_SYNONYMS: 10,
-    MIN_CORRECTION_LENGTH: 4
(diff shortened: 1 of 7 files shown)


lib/geocoder/spatialmatch.js
```diff
@@ -5,6 +5,7 @@ const coalesce = require('@mapbox/carmen-cache').coalesce;
 const bbox = require('../util/bbox.js');
 const termops = require('../text-processing/termops');
 const constants = require('../constants');
+let limit = constants.stackable_limit;
```
apendleton: This should be const, if we need to set it at all... but can't we just use the constant directly?

(I know it's `let` because we're mutating it later, but... we shouldn't mutate it later.)

lib/geocoder/spatialmatch.js
```diff
@@ -26,6 +27,8 @@ module.exports.sortByZoomIdx = sortByZoomIdx;
 
 function spatialmatch(query, phrasematchResults, options, callback) {
     let stacks;
+    limit = options.stackable_limit;
```
apendleton: We shouldn't be mutating a global variable based on the options for a particular query. We should be passing this through into stackable.

lib/geocoder/spatialmatch.js
```diff
@@ -274,8 +277,6 @@ function stackable(phrasematchResults, memo, idx, mask, nmask, stack, relev, adj
 
     // Though spatialmatches are sliced to 30 elements after this step
     // leave some headroom as this step does not include type filtering.
```
apendleton: This comment refers to the 100. If we keep the comment, we should move it to where the default is now defined.

lib/constants.js
```diff
@@ -8,5 +8,13 @@ module.exports = {
     MAX_QUERY_TOKENS: 20,
     PROXIMITY_RADIUS: 200,
     MAX_TEXT_SYNONYMS: 10,
-    MIN_CORRECTION_LENGTH: 4
+    MIN_CORRECTION_LENGTH: 4,
+    // STACKABLE_LIMIT: https://github.com/mapbox/carmen/blob/master/lib/geocoder/spatialmatch.js#L277
```
apendleton: Nitpick, but these line numbers are presumably all going to change, so I'm not sure if putting them in the comments is that useful.

lib/geocoder/geocode.js
```diff
@@ -52,6 +53,10 @@ module.exports = function(geocoder, query, options, callback) {
     options.bbox = options.bbox || false;
     options.reverseMode = options.reverseMode || 'distance';
     options.routing = options.routing || false;
+    options.stackable_limit = options.stackable_limit || constants.STACKABLE_LIMIT;
```
apendleton: Another nitpick, but can we separate the things that are how carmen is being set up from the options coming in from the users? I was thinking maybe we'd put the cutoffs in a separate object, but even just some whitespace or a comment or something would be good, since these feel like categorically different kinds of things.

Issue #749, opened 2018-10-11
Sort final results with equal relevance by distance rather than scoredist


lizziegooding:
### Context
We currently sort several properties, most significantly relevance and `scoredist`, which at this point in verifymatch is whatever larger, a feature's `score` or calculated `scoredist`.

https://github.com/mapbox/carmen/blob/b9c2bb783e98a5149e3af85c7216e4f8948ae0da/lib/geocoder/verifymatch.js#L389-L399

We should instead try sorting results with equal relevance by distance to see if it improves the quality of search results.

### Next Steps
- [x] sort results by relevance, than distance
- [ ] publish dev version
- [ ] see how it changes pass rates downstream

cc @miccolis @apendleton @aarthykc @boblannon

lizziegooding:
Not going with this approach, moved to https://github.com/mapbox/carmen/pull/768

PR #756, opened 2018-10-15
Sort final results with equal relevance by distance

lizziegooding:
### Context
Ref https://github.com/mapbox/carmen/issues/749. Sorts results with equal relevance by distance and disregards score entirely. I originally tried stripping out all the places in verifymatch where we use `score` or `scoredist` and found it ended up being more trouble than it was worth. We should still run this branch against our downstream acceptance tests, but it's very unlikely we'd want to take this approach in production. It completely disregards _any_ sorting based on score/prominence and would mean that lower prominence, local results would be surfaced before local prominent results.

### Summary of Changes
- [ ] sort by `distance` instead of `scoredist` in `sortFeature` function
- [ ] sort by `distance` instead of `sortContext` in `sortFeature` function
- [ ] update acceptance and unit tests

### Next Steps
- [ ] publish a dev release
- [ ] run against downstream tests

### DIFF
--- a/lib/geocoder/verifymatch.js
+++ b/lib/geocoder/verifymatch.js
@@ -206,6 +206,7 @@ function verifyFeatures(query, geocoder, spatialmatches, loaded, options) {
         }
     }
 
+
     // Sort + disallow more than options.limit_verify of
     // the best results at this point.
     result.sort(sortFeature);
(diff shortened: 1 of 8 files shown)


lizziegooding:

PR #760, opened 2018-10-17
[WIP] More detailed time logging in `verifymatch` and `spatialmatch`

boblannon:
### Context

To get a better understanding of where our time is spent in forward geocodes, this PR adds more fine-grained timing breakdowns to `spatialmatch` and `verifymatch` when `options.stats=true`.  

Hoo lordy is this going to break a lot of tests.  if we want to keep this stuff around (and maybe that's not so crazy) we'd have to update a bunch of tests because this changes the signatures of `verifymatch` and `spatialmatch`.

### Summary of Changes
- [x] separate `options.stats` from `stats_data`
- [x] separate `options.debug` from `debug_data`
- [x] add more time measurements in `spatialmatch` and `verifymatch`
- [x] alter `bin/carmen.js` to display additional stats.

### Next Steps

- [ ] if we're going to keep this, need to fix tests

cc @mapbox/geocoding-gang


boblannon:
closing this for now but it's there if we need it.  no obvious advantages over using standard profiling tools.

PR #761, opened 2018-10-17
Don't bother stacking if a query is a single word

miccolis:
### Context
I'm wondering if stackable really needs for run for single word queries. PR attempts this shortcut. I'm pretty sure there are still missing pieces here, but tests pass. 

### Next Steps
- [ ] See what happens with a larger dataset
- [ ] Enforce the same limit that stackable does

cc @mapbox/geocoding-gang

### DIFF
--- a/lib/geocoder/spatialmatch.js
+++ b/lib/geocoder/spatialmatch.js
@@ -27,7 +27,18 @@ module.exports.sortByZoomIdx = sortByZoomIdx;
 function spatialmatch(query, phrasematchResults, options, callback) {
     let stacks;
 
-    if (phrasematchResults.length) {
+    if (phrasematchResults.length === 0) {
+        stacks = [];
+    } else if (query.length === 1) {
+        stacks = phrasematchResults.reduce((m, r) => {
+            if (r.length === 0) return m;
+            for (let i = 0; i < r.phrasematches.length; i++) {
+                // Todo probably missing relev & adjRelev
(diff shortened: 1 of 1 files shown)


apendleton:
This looks fine to me. I do wonder if it solves an actual problem (are one-word queries slow enough that they need to be special-cased?), but assuming yes, no objection. Would want to confirm passrates on a downstream PR before approving/merging though, just to make sure we're not missing any corner-case behavior changes this might introduce.

lib/geocoder/spatialmatch.js
```diff
@@ -27,7 +27,20 @@ module.exports.sortByZoomIdx = sortByZoomIdx;
...
     let stacks;
 
-    if (phrasematchResults.length) {
+    if (query.length === 1) {
+        stacks = phrasematchResults.reduce((m, r) => {
+            if (r.length === 0) return m;
+            for (let i = 0; i < r.phrasematches.length; i++) {
+                const v = [r.phrasematches[i]];
+                v.relev = v[0].weight;
+                v.adjRelev = v[0].weight * v[0].editMultiplier;
```
miccolis: @apendleton do I need to copy scoreFactor forward here too. Didn't that just change in master?

miccolis:
Totally agree that it's unclear what affect this would actually have. It does avoid some work, so if single word queries are a sizable portion of traffic the cumulative difference may be noticeable, or maybe not. If not, the extra complexity here isn't worth it.

PR #762, opened 2018-10-18
[dnm] Dist experiments

apendleton:
For now this is a test branch to experiment with some ideas around distance and proximity. I'll edit this section to have the traditional summary/next steps if it seems like it's viable, but I suspect that ultimately we'll take some ideas from here and combine them with other folks' experiments.

Major things going on:
* I ditched the geometric mean contraption for scaling distance values into something on the same order of magnitude as scores, since per @lizziegooding 's work it's super unstable and swings wildly depending on which features happen to make it over the cutoff into the consideration bucket. Instead, my goal is to come up with a constant scaling factor that we can use for all features and queries, but that's still somewhat derived from the data rather than being baked into Carmen. Current strategy: bucket all the indexes into types, take max scores of all indexes within each type and average them, then take the geometric mean of those across types. I think this will give us a number that's pretty stable over time and resistant to both outliers and to adding more indexes of any given type.
* I'm playing with a gaussian curve rather than the current quadratic for the distance calculation. There will still be a notion of a radius, but because of the shape of the curve my suspicion is that we'll want them to be smaller numbers than we have now, and the effect won't stop working past that radius, it'll just significantly decrease
* I sort of became skeptical of using the zoom as-is as a direct input into the calculation. They're not distributed in a way that gives the behavior we want in practice: some layers that seem relatively similar are far apart, and some that we'd expect to be far apart are pretty close. Instead, I'm bucketing the zoom levels and manually assigning an effective radius to each bucket
* I'm going to combine distance and score by sum instead of max. Product is also a viable option to try.

Tests don't pass, and I haven't done much work with it with real data. Pretty sure at a minimum that some constants will have to change. Also I think I left a console.log in there someplace.

### DIFF
--- a/index.js
+++ b/index.js
@@ -251,6 +251,17 @@ function Geocoder(indexes, options) {
             this.byidx[i].bmask = bmask;
         }
 
+        // come up with a constant that represents the rough order of magnitude
+        // of all the scores. Strategy: look at each type, find the mean score
+        // for that type, and then take the geometric mean of the means of the types
+        let scoreAvgs = [];
+        for (const typename of Object.keys(this.bytype)) {
+            const scores = this.bytype[typename].map((source) => source.maxscore || 0);
+            scoreAvgs.push((scores.reduce((a, b) => a + b) / scores.length) || 1);
+        }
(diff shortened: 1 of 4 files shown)


apendleton:
Subsumed by #768

PR #765, opened 2018-10-18
Consider nmask earlier in stackable

miccolis:
### Context
Another potential speedup, this PR avoids checking the `nmask` or the index and match being evaluated more than once per index. Not totally clear what this gets us, but it should avoid unnecessary iteration.

### Next Steps
- [ ] Verify that we want this, and remote the extraneous throw :)

cc @mapbox/geocoding-gang

### DIFF
--- a/lib/geocoder/spatialmatch.js
+++ b/lib/geocoder/spatialmatch.js
@@ -265,6 +265,8 @@ function stackable(phrasematchResults, limit, memo, idx, mask, nmask, stack, rel
 
     const phrasematchResult = phrasematchResults[idx];
 
+    if (nmask & phrasematchResult.nmask) return;
+
     // For each stacked item check the next bmask for its idx.
     // If the bmask includes the idx these indexes cannot stack
     // (their geocoder_stack do not intersect at all).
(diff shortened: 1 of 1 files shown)


boblannon:
played this against master for 992 forward queries (pulled from logs).  this looks like it has a measurable impact.  measurements here are in milliseconds:

|       | stackable (baseline) | stackable (early_nmask) |
|-------|----------------------|-------------------------|
| count | 992.000000           | 992.000000              |
| mean  | 11.310484            | 8.506048                |
| std   | 40.982096            | 32.443788               |
| min   | 0.000000             | 0.000000                |
| 25%   | 0.000000             | 0.000000                |
| 50%   | 1.000000             | 1.000000                |
| 75%   | 9.000000             | 6.250000                |
| max   | 632.000000           | 602.000000              |

apendleton:
Want to confirm my own understanding here: superficially it looks like the way the code is in master, if the check that's inside the loop returned true, it would `continue` through all the iterations of the loop and then either implicitly return `undefined` if `idx != 0` or do a thing to prep actual output if `idx == 0`. The new branch always returns `undefined` if the check succeeds, regardless of `idx`.

I *think* that's safe because when `idx == 0`, `nmask` would also be `0` such that that `&` in the check would always return false, so there would never be a condition where we'd `continue` through the loop but still do the things in the `if idx == 0` block. Is that your understanding as well?

If so, looks good to me, contingent on downstream passrates.

miccolis:
@apendleton yes, you are correct. However, I agree that this idx check after a pair of mask based returns (for `bmask` and now `nmask` too) is totally confusing. If that assumption hadn't existed before I would have done this differently. As written is the minimal change, which I figured was preferable.

PR #766, opened 2018-10-19
Use proximity point and index bboxes as an additional stack sort hint

apendleton:
### Context
#758 resulted in some changes in behavior for proximity-sensitive queries such that in some circumstances, the ideal response no longer made it above the 30-stack cutoff in spatialmatch. This PR introduces an additional hint for proximity queries to increase the odds that the right one makes it above the cutoff, and also fixes a bug I incidentally found along the way.

### Summary of Changes
- [x] for queries that have proximity points, at phrasematch time, check for each index whether or not the proximity point is inside the index's bbox, and attach the result to the phrasematch object
- [x] use bbox/prox-point overlap as an ordering hint, to prefer matches from indexes that contain the prox point over ones that don't
- [x] ensure that prox info is copied over in `collapseToArchetypes` in spatialmatch, and also do the same for `scorefactor`, which was supposed to be copied but wasn't, rendering that sort hint ineffective

### Next Steps
- [x] Tests. Definitely for the new proximity behavior, and also probably for the scorefactor behavior, since that was broken and nobody realized.

cc @mapbox/search

### DIFF
--- a/lib/geocoder/phrasematch.js
+++ b/lib/geocoder/phrasematch.js
@@ -152,6 +152,10 @@ module.exports = function phrasematch(source, query, options, callback) {
     // cross-index comparisons.
     const scorefactor = (source._geocoder.freq.get('__MAX__') || [0])[0] || 1;
 
+    const proxMatch = options.proximity ?
+        bb.inside(options.proximity, source.bounds) :
+        false;
+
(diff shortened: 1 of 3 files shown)


apendleton:
@ingalls test looks good. 🌮 to your changes.

PR #764, opened 2018-10-18
[dnm] refactor verifymatch scoredist implementation

lizziegooding:
### Context

With proximity enabled, we should consider distance from the proximity point as well as feature prominence when deciding the order of results to return to users. The simplest way to implement a score that's a composite of two factors is to scale each factor to a bounded range of values and combine them either by multiplying or subtracting. If they're of the same importance, we would bound them to the same scale, say 0 - 1, and if they're of different importances, we can manipulate that scale so the more important factor is scaled to a higher range of vales than the less important factor. 

For our implementation, `score` is already scaled roughly along an exponential curve. I started experimenting with reducing `score` to the point where it can be compared along roughly the same scale (but not necessarily the same curve) as `distance` and multiplying the product of those values. I scaled distance along a very extreme curve (hyperbola `y = 1/x`) where as the distance approaches 0, `y` approaches `∞` The distance is scaled by the proximity bias radius to extend the bias for larger features that are higher on the admin hierarchy. Because I'm combining with multiplication, the minimum value for either value is 1. If the feature is beyond the bias radius, the distance weighting defaults to 1 so these features are only sorted by score.

Here's an example of the affect this change has had on sort order for a particular test case, querying `missi` near the Mission in San Francisco:

| order | master                                                      | proximity-algorithm                                         |
|-------|-------------------------------------------------------------|-------------------------------------------------------------|
| 1     | Mississippi                                                 | Mission                                                     |
| 2     | Mission                                                     | Mission Branch Library                                      |
| 3     | Mission Branch Library                                      | Mission Wishing Tree                                        |
| 4     | Mission Wishing Tree                                        | Mission Pie,cafe, coffee, tea, tea house                    |
| 5     | Mission's Kitchen,fast food                                 | Mission's Kitchen,fast food                                 |
| 6     | Mission Cultural Center for Latino Arts,college, university | Mission Terrace                                             |
| 7     | Mission Pie,cafe, coffee, tea, tea house                    | Mission Bay                                                 |
| 8     | Mission Critter                                             | Mission Cultural Center for Latino Arts,college, university |
| 9     | Mission Gastroclub                                          | Mission Critter                                             |
| 10    | Mission Tires & Service Center                              | Mission Gastroclub                                          |
| 11    | Mission Skateboards                                         | Mission Tires & Service Center                              |
| 12    | Mission Dental Care                                         | Mission Skateboards                                         |
| 13    | Mission Pet Hospital                                        | Mission Dental Care                                         |
[...]

Issue #763, opened 2018-10-18
Research and implement a different `scoredist` calculation in verifyMatch


lizziegooding:
Based on internal test cases, it doesn't seem like our current approach to calculating proximity is tenable. Cutting this ticket to track/explore alternatives.

lizziegooding:
Tracking in https://github.com/mapbox/carmen/pull/764 and https://github.com/mapbox/carmen/pull/762

lizziegooding:
Done in https://github.com/mapbox/api-geocoder/pull/2348#issuecomment-433119145

PR #768, opened 2018-10-23
[dnm] scale calculate distance and score to the same range

lizziegooding:
### Context
Ref https://github.com/mapbox/carmen/issues/763. 

### scoredist
`scoredist` is calculated as the product of a feature's `scoreWeight` and `distWeight`. Both `scoreWeight` and `distWeight` have the same range between 1 and 11, meaning that `scoredist` will always be a value between 1 and 121. This range is arbitrary, but given the decimal precision we're working with, scaling values to a range larger than 0 and 1 makes final values a little more human readable. The minimum value of both is 1 so one weight doesn't cancel the other out (for example, if a feature had a score of `0` but a distance of `0.01` miles).

### scoreWeight
`scoreWeight` is calculated as the feature's score divided by the maximum score of all features in the index scaled to a value between 1 and 11. This proportional score means that differences in score remain proportional while magnitude is reduced so `scoreWeight` and `distWeight` are weighed evenly in the final `scoredist` calculation. I also experimented with using a logarithmic scale, similar to https://github.com/mapbox/carmen/pull/764 but found that it over emphasized differences in low scores (say a score of `1` vs `3`) and minimized differences in high scores (say `33,700` vs `660,000`). 

### distWeight
`distWeight` is calculated as the distance between the feature and the proximity point normalized by the proximity bias radius (`dist / radius`) for that feature type scaled along a gaussian curve. I experimented with a few different types of curves but ultimately chose a gaussian/normal distribution because it provides a slight "table top" near the origin (distance 0), decays rapidly at middling distances, and then decays slowly again towards the edge of the proximity radius. 

image
_From the elastic search docs, gaussian vs linear vs exponential curves_

I chose a relatively large variance (σ²) so the curve would be more spread out and the decay less severe.
image

### Resources
- https://www.elastic.co/guide/en/elasticsearch/guide/master/decay-functions.html
- https://www.elastic.co/guide/en/elasticsearch/guide/master/boosting-by-popularity.html
- https://www.elastic.co/guide/en/elasticsearch/reference/master/query-dsl-function-score-query.html#function-decay

### Next Steps
- [ ] publish dev release
- [ ]  run against downstream tests
- [ ] (maybe) keep playing with the dist curve / variance


apendleton:
In case it's useful: what I ended up with in my branch was a super simple incarnation of a Gaussian curve that made the math really easy:

image

(that's described as the Stigler simplification on the [normal distribution wikipedia page](https://en.wikipedia.org/wiki/Normal_distribution))

and then just divided x by an effective-radius number to control how far to spread the range. It's not a hard cutoff, but most of the effect is bigger than the radius number, and it falls asymptotically out from there.

So like, for an effective radius of 2, the graph/formula looks like this:

image

and then I did different radii for different feature types. Might be less involved than the whole generalized formula?

Also worth noting, though: as a matter of organizing philosophy/practice, in Carmen we treat the names of specific types as arbitrary, and leave it up to individual consumers of the library to configure whatever types they want; we don't hardcode the names of specific feature types that Mapbox happens to use in its particular implementation, and `poi`, `locality`, etc., have no special significance in the existing codebase. For mine I keyed it on zoom levels, but if that doesn't give us enough granularity, we should probably instead make this radius a directly configurable option in the tilejsons themselves.

lizziegooding:
Noting per chat w/ @apendleton, I'm going to experiment with a few iterations of this implementation and see the effect on downstream passrates:

- current implementation. If we do decide to go with this approach, we'll need to add proximity bias as an option to our tilejsons, which will require updating additional projects
- set radius by zoom level instead of type
- use the simplified gaussian curve calculation @apendleton outlined in https://github.com/mapbox/carmen/pull/768#issuecomment-432077303
- scale features by distance beyond the proximity radius cut off by removing https://github.com/mapbox/carmen/blob/205eef40877df4612b03c9e249a527edbba30682/lib/util/proximity.js#L115-L116 similar to https://github.com/mapbox/carmen/pull/762

lib/util/proximity.js
```diff
@@ -67,26 +71,103 @@ function center2zxy(center, z) {
...
+ * proximity radius.
+ * @param {Number} dist The distance in miles from the feature to the proximity point
+ * @param {String} type The feature type. If the feature has multiple types,
+ *  use the last value
+ * @return {Number} scaled score value
+ */
+function distWeight(dist, zoom) {
+    const radius = scaleRadius(zoom);
+    // set gaussVal to 0.5 when normalized distance is 3/4 of the proximity radius (e.g. 75 / 100 miles)
+    const gaussVal = gauss(dist / radius, variance(0.75, 0.5));
```
apendleton: `variance(0.75, 0.5)` looks potentially complicated/computationally-expensive and always returns the same output. Can we calculate it outside the function and store the result as a global rather than computing it every time?

lib/util/proximity.js
```diff
@@ -67,26 +71,103 @@ function center2zxy(center, z) {
...
-    // flip distance to a float between 0-1, 1 being closest to the center point
-    let distVal = 1 - Math.min(dist / weightedRadius, 1);
-    // square to weight nearer features higher
-    distVal = Math.pow(distVal, 2);
-    // heuristic: the closest features can be up to 100x times the geometric mean
-    return parseFloat((100 * meanScore * distVal).toFixed(4));
+function scoredist(score, maxScore, dist, zoom) {
+    const scoreVal = scoreWeight(score, maxScore);
+    const distVal = distWeight(dist, zoom);
+    return parseFloat((distVal * scoreVal).toFixed(6));
```
apendleton: I'm still confused as to why we round this instead of just using it (not a criticism since we did it before, but still weird). If we're looking for more operations to shave off for speed, we should probably drop this -- it allocates and then immediately destroys a string for no apparent reason.

index.js
```diff
@@ -270,6 +270,9 @@ function Geocoder(indexes, options) {
             this.byidx[i].bmask = bmask;
         }
 
+        // Find the max score of all features in all indexes
+        this.maxScore = this.byidx.reduce((max, source) => Math.max(max, source.maxscore), 0) || 1;
```
apendleton: Flagging that this potentially leaves us pretty sensitive to jitter if the one index with the highest-score features in it ever changes. I think we can live with that though.

lib/util/proximity.js
```diff
@@ -67,26 +71,103 @@ function center2zxy(center, z) {
...
+ * Set proximity radius by feature type
+ * @param {String} type The feature type. If the feature has multiple types,
+ *  use the last value
+ * @return {Number} scaled radius in miles
+ */
+function scaleRadius(zoom) {
+    let radius;
+    if (zoom <= 6) {
+        // country, region
+        radius = 1800;
```
apendleton: I put these here in my branch for expediency, but we may want to move them into `constants.js` eventually. We can leave them alone now if we're revisiting this anyway,  but at the very least we should probably get rid of the radius constant that's in there now that doesn't do anything anymore.

test/acceptance/geocode-unit.scoredist.test.js
```diff
@@ -62,17 +62,17 @@ const queue = require('d3-queue').queue;
         });
         q.awaitAll(t.end);
     });
-    tape('geocode proximity=10,10 => superscored', (t) => {
+    tape('geocode proximity=10,10 => nearest', (t) => {
```
apendleton: Capturing per voice: I think we should have one of these where we get the closest one, and another where we get a higher-scored one even though it's farther away, if it's really high-scored and we're not that close to the closest one. I *think* that's the intent of this test?

lizziegooding:
After testing downstream, I ended up making the following changes to the original implementation:
- set radius by zoom level instead of type. I'd like to change this sooner rather than later so we can set custom radiuses per feature type, but am prioritizing shipping this PR first.
- use the same gaussian curve calculation for distance but different `scale` and `decay` values to flatten and extend the curve
screen_shot_2018-10-24_at_1_42_58_pm

Noting while we're here, that if we were to go back to using the original curve, we should use @apendleton 's suggestion of Stiegler's simplified normal distribution, as it it almost identical to the gauss curve in https://github.com/mapbox/carmen/pull/768/commits/205eef40877df4612b03c9e249a527edbba30682
screen_shot_2018-10-23_at_5_18_34_pm

- removed the previous limit so features beyond the proximity radius cut off can have a slightly higher distance weighting than 1. This affect disappears when the feature is roughly 4x the proximity radius away from the proximity point (so if the radius were 100 miles, the feature would receive distance weighting slightly above 1 between 100 and 400 miles from the feature, but no distance bump beyond 400 miles). This change didn't affect downstream proximity passrates, but did, result in a more intuitive sorting order for unit test cases in carmen like https://github.com/mapbox/carmen/blob/6ea84c1d293ab5350da706249cf69bf439d94ddb/test/unit/util/proximity.scoredist.test.js#L140-L143
where for features with roughly the same score, results outside the proximity radius (>100 miles distance) but within the same country return before results thousands of miles away.

lib/util/proximity.js
lizziegooding: 👍 can we still keep/test the variance function even if we don't call it in production? or does that feel overly messy? I could see wanting to change that value in the short term and the tests demonstrate pretty clearly what it does/the affect of changing it.

lib/util/proximity.js
lizziegooding: Yeah, I'm fine with removing this. I'll round these values in tests so we don't have issues with decimal precision.

lib/util/proximity.js
apendleton: Yeah totally. I just meant, like, above/outside the function put `const VARIANCE_CONSTANT = variance(0.75, 5);` and then inside the function, `const gaussVal = gauss(dist / radius, VARIANCE_CONSTANT);` or whatever. So, still use the function (and test it if we want), but just call it once when carmen loads instead of every time.

lib/util/proximity.js
lizziegooding: Good call, I can move these over to constants.js. I kept the original proximity radius constant because it's passed to carmen-cache in spatialmatch.js https://github.com/mapbox/carmen/blob/8ab8c3e0ce8933a256975920666076c9b40fd2ac/lib/constants.js#L9.

apendleton: Oh good catch. Forgot about that.

lib/util/proximity.js
lizziegooding: Ah cool, thanks for the clarification 👍

PR #776, opened 2018-11-02
Don't deduplicate interpolated points into non-interpolated cousins

miccolis:
### Context
Noticed a case where two same addresses exist in two cities on either side of a river. One was interpolated and one was not. The non-interpolated one always killed the other from the search. It shouldn't.

In talking to @ingalls it seems like this check predates changes to how addresses are generated that should make it entirely unnecessary. So this PR removes it.

### Summary of Changes
-  Don't de-duplicate address based on interpolation.

### Next Steps

- [x] Take a closer look at the tests that are commented out, they can probably be removed.
- [x] Run these changes against a larger set of test cases.

cc @mapbox/search


miccolis:
My test case here has been a range of addresses that exist in Manhattan and Brooklyn; `1-5 Bond Street`. While interpolation is no long the reason these fail, in the data I'm looking its now the distance that triggers the de-deduplication... The code calculates a distance of 4.3149275866570616 km, this falls [below the 5 km threshold](https://github.com/mapbox/carmen/blob/84252775fb002214e562c91eea2b42f8bad2820b/lib/geocoder/dedupe.js#L42) and the lower ranked point is removed from the list.

I'm trying not to let this one edge case bias me. I still think removing the interpolation code is a good idea, but should we also lower / remove this limit? Or detect something like this another way?

ingalls:
PT2ITP should handle this on the data side fairly efficiently now, as long as test results look good, we can tweak this later from PT2ITP if we find an edge case.

boblannon:
context for that force-push: we typically update the changelog and bump the version after the merge to master.

Issue #771, opened 2018-10-29
Revisit result sorting to sort on a single composite incorporating relevance and score/proximity


apendleton:
### Context
Currently, carmen sorts first on relevance and then, as a tie-breaker, on proximity and score. For results with large relevance differences, this is good: results that match most or all of the user's query clearly better match the user's intent than ones that only match parts of it (e.g., a complete match for a fully qualified address rather than a partial match for only the city or only the housenumber and street), even if those partial matches are nearby. However, for two matches that are of very similar but not identical relevance, often due to a small penalty applied for an out-of-language match or a fuzzy match (i.e., a typo), we often end up matching very far-away things that share that typo or happen to be in that language when the user wanted something that was almost as good a much but was closer (or more salient).

We should experiment with a new approach where, rather than sorting first on relevance and then on proximity/score, we construct a single composite value that's dominated by relevance but allows small relevance differences to be overtaken by large proximity or score differences. To start, we could consider a weighted average of, say, 90% relevance + 10% prox/score, and see what happens.

We'll probably want to test on three classes of queries where we know this is a problem:
* language penalties
* typos
* word frequency penalties (e.g., matching variants that have been auto-indexed at reduced relevance without "the")

each of these might have a slightly different magnitude of penalty, and also gets its penalty applied in a different place, so we may not have equal success with all of them.

### Next Steps
- [ ] first let's try just doing this in verifymatch: https://github.com/mapbox/carmen/blob/master/lib/geocoder/verifymatch.js#L398
- [ ] we may also have some features that don't even make it to verifymatch because they get lost in a similar sort in carmen-cache, so we may need to address it there as well

lizziegooding:
Done in https://github.com/mapbox/carmen/pull/777

PR #777, opened 2018-11-02
Sort final results w/ proximity enabled based on relevance and score/proximity composite 

lizziegooding:
### Context
Ref https://github.com/mapbox/carmen/issues/771. First pass at creating a composite score with `relevance` _and_ `scoredist` when `proximity` is enabled rather than using `scoredist` as a tie breaker for `relevance`. The current implementation gives `relevance` a weight of 1 and `scoredist` a weight of 0.3. These thresholds are arbitrary; I chose them because they create a final score that  overcomes a single 4% penalty on an element in the stack when the feature doesn't have a `carmen:text_<language code>` property in the user-specified language. The final result is stored in the `carmen:relevance` property (we should probably revisit this naming) and is a value between 0 and 1.

### Summary of Changes
- [x] add `carmen:relevance` property to features when proximity is enabled
- [x] sort by `carmen:relevance` first if it exists

### Next Steps
- [x] test against downstream passrates. I chose these weight thresholds based on a single test case https://github.com/mapbox/carmen/compare/bump-language-penalty?expand=1#diff-3170dfb751a08557bb216e0c49b2cea1R49 and will likely tweak
- [ ] make sure we have test cases that capture this behavior for language, typo, and word frequency penalties
- [ ] document https://github.com/mapbox/carmen/compare/bump-language-penalty?expand=1#diff-d3439e250bdd6a8098bb9bd4095efcc1R185 and make constants/math more transparent
- [x] move https://github.com/mapbox/carmen/compare/bump-language-penalty?expand=1#diff-3170dfb751a08557bb216e0c49b2cea1R49 to its own test file
- [ ] use `carmen:relevance` as the final `relevance` displayed to the use  in the response when `proximity` is enabled

cc @apendleton @aarthykc @boblannon @miccolis

### DIFF
--- a/lib/geocoder/verifymatch.js
+++ b/lib/geocoder/verifymatch.js
@@ -407,8 +407,8 @@ function verifyContext(context, peers, strict, loose, indexes, options, geocoder
                 feat.properties['carmen:spatialmatch'].relev,
                 feat.properties['carmen:scoredist'],
                 feat.properties['carmen:address'],
-                feat.geometry && feat.geometry.omitted ? 1 : 0,
-                feat.properties['carmen:score'] < 0 ? 1 : 0
+                feat.geometry && feat.geometry.omitted,
+                feat.properties['carmen:score'] < 0
             );
         } else {
             feat.properties['carmen:scoredist'] += squishy;
--- a/lib/geocoder/verifymatch.js
+++ b/lib/geocoder/verifymatch.js
@@ -207,6 +207,10 @@ function verifyFeatures(query, geocoder, spatialmatches, loaded, options) {
                 feat.properties['carmen:distance'],
                 feat.properties['carmen:zoom']
             );
+            feat.properties['carmen:relevance'] = proximity.relevanceScore(
+                feat.properties['carmen:spatialmatch'].relev,
+                feat.properties['carmen:scoredist']
+            );
         } else {
             feat.properties['carmen:scoredist'] = feat.properties['carmen:score'];
         }
@@ -396,7 +400,8 @@ function verifyContext(context, peers, strict, loose, indexes) {
 * sortFeature - sorts the best results in the features according to the relevance, scoredist, position or if the address exists
 */
 function sortFeature(a, b) {
-    return (b.properties['carmen:spatialmatch'].relev - a.properties['carmen:spatialmatch'].relev) ||
+    return ((b.properties['carmen:relevance'] ? b.properties['carmen:relevance'] : 0) - (a.properties['carmen:relevance'] ? a.properties['carmen:relevance'] : 0)) ||
+        (b.properties['carmen:spatialmatch'].relev - a.properties['carmen:spatialmatch'].relev) ||
         ((a.properties['carmen:address'] === null ? 1 : 0) - (b.properties['carmen:address'] === null ? 1 : 0)) ||
         ((a.geometry && a.geometry.omitted ? 1 : 0) - (b.geometry && b.geometry.omitted ? 1 : 0)) ||
         ((b.properties['carmen:scoredist'] || 0) - (a.properties['carmen:scoredist'] || 0)) ||
--- a/lib/geocoder/verifymatch.js
+++ b/lib/geocoder/verifymatch.js
@@ -211,8 +211,8 @@ function verifyFeatures(query, geocoder, spatialmatches, loaded, options) {
                 feat.properties['carmen:spatialmatch'].relev,
                 feat.properties['carmen:scoredist'],
                 feat.properties['carmen:address'],
-                feat.geometry && feat.geometry.omitted ? 1 : 0,
-                feat.properties['carmen:score'] < 0 ? 1 : 0
+                feat.geometry && feat.geometry.omitted,
+                feat.properties['carmen:score'] < 0
             );
         } else {
             feat.properties['carmen:scoredist'] = feat.properties['carmen:score'];
--- a/lib/geocoder/verifymatch.js
+++ b/lib/geocoder/verifymatch.js
@@ -267,7 +267,7 @@ function loadContexts(geocoder, features, sets, options, callback) {
 
     q.awaitAll((err, contexts) => {
         if (err) return callback(err);
-        callback(null, verifyContexts(contexts, sets, geocoder.indexes));
+        callback(null, verifyContexts(contexts, sets, geocoder.indexes, options, geocoder));
     });
 }
 
@@ -279,7 +279,7 @@ function loadContexts(geocoder, features, sets, options, callback) {
 * @params {Object} indexes - indexes in the instance of the geocoder for e.g: {country: {}, region: {}, place:{}}
 * @returns {Object} contexts - feature hierachy sorted by the score value
 */
-function verifyContexts(contexts, sets, indexes) {
+function verifyContexts(contexts, sets, indexes, options, geocoder) {
     const peers = {};
     for (let c = 0; c < contexts.length; c++) {
         peers[contexts[c][0].properties['carmen:tmpid']] = contexts[c][0];
@@ -309,8 +309,8 @@ function verifyContexts(contexts, sets, indexes) {
             }
         }
 
-        const strictRelev = verifyContext(context, peers, verify, {}, indexes);
-        const looseRelev = verifyContext(context, peers, verify, sets, indexes);
+        const strictRelev = verifyContext(context, peers, verify, {}, indexes, options, geocoder);
+        const looseRelev = verifyContext(context, peers, verify, sets, indexes, options, geocoder);
         context._relevance = Math.max(strictRelev, looseRelev);
     }
     contexts.sort(sortContext);
@@ -320,7 +320,7 @@ function verifyContexts(contexts, sets, indexes) {
 /**
 https://github.com/mapbox/geocoding/issues/302
 */
-function verifyContext(context, peers, strict, loose, indexes) {
+function verifyContext(context, peers, strict, loose, indexes, options, geocoder) {
     const addressOrder = context[0].properties['carmen:geocoder_address_order'];
     let usedmask = 0;
     let lastmask = -1;
@@ -391,8 +391,29 @@ function verifyContext(context, peers, strict, loose, indexes) {
     // contained "gaps" in continuity between index levels --
     // EXCEPT in the case of an eligible higher-level feature that shares an
     // exact carmen:text value with something in its context (e.g. New York, New York)
-    if (squishy > 0)
-        context[0].properties['carmen:scoredist'] += squishy;
+    if (squishy > 0) {
+        const feat = context[0];
+        // Recalculate carmen:scoredist and carmen:relevance for smallest squishy feature
+        // with new higher score to ensure it returns before larger squishy features
+        if (options.proximity) {
+            feat.properties['carmen:scoredist'] = proximity.scoredist(
+                Math.min(feat.properties['carmen:score'] + squishy, geocoder.maxScore),
+                geocoder.minScore,
+                geocoder.maxScore,
+                feat.properties['carmen:distance'],
+                feat.properties['carmen:zoom']
+            );
+            feat.properties['carmen:relevance'] = proximity.relevanceScore(
+                feat.properties['carmen:spatialmatch'].relev,
+                feat.properties['carmen:scoredist'],
+                feat.properties['carmen:address'],
+                feat.geometry && feat.geometry.omitted ? 1 : 0,
+                feat.properties['carmen:score'] < 0 ? 1 : 0
+            );
+        } else {
+            feat.properties['carmen:scoredist'] += squishy;
+        }
+    }
 
     relevance = relevance > 0 ? relevance : 0;
 
--- a/lib/geocoder/verifymatch.js
+++ b/lib/geocoder/verifymatch.js
@@ -199,10 +199,10 @@ function verifyFeatures(query, geocoder, spatialmatches, loaded, options) {
     // Set a score + distance combined heuristic.
     for (let k = 0; k < result.length; k++) {
         const feat = result[k];
-        // ghost features don't participate
-        if (options.proximity && feat.properties['carmen:score'] >= 0) {
+        if (options.proximity) {
             feat.properties['carmen:scoredist'] = proximity.scoredist(
                 feat.properties['carmen:score'],
+                geocoder.minScore,
                 geocoder.maxScore,
                 feat.properties['carmen:distance'],
                 feat.properties['carmen:zoom']
@@ -211,7 +211,8 @@ function verifyFeatures(query, geocoder, spatialmatches, loaded, options) {
                 feat.properties['carmen:spatialmatch'].relev,
                 feat.properties['carmen:scoredist'],
                 feat.properties['carmen:address'],
-                feat.geometry && feat.geometry.omitted ? 1 : 0
+                feat.geometry && feat.geometry.omitted ? 1 : 0,
+                feat.properties['carmen:score'] < 0 ? 1 : 0
             );
         } else {
             feat.properties['carmen:scoredist'] = feat.properties['carmen:score'];
--- a/lib/geocoder/verifymatch.js
+++ b/lib/geocoder/verifymatch.js
@@ -209,7 +209,9 @@ function verifyFeatures(query, geocoder, spatialmatches, loaded, options) {
             );
             feat.properties['carmen:relevance'] = proximity.relevanceScore(
                 feat.properties['carmen:spatialmatch'].relev,
-                feat.properties['carmen:scoredist']
+                feat.properties['carmen:scoredist'],
+                feat.properties['carmen:address'],
+                feat.geometry && feat.geometry.omitted ? 1 : 0
             );
         } else {
             feat.properties['carmen:scoredist'] = feat.properties['carmen:score'];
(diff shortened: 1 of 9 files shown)


lizziegooding:
I was looking at a downstream test that failed when searching for `Beirut` with proximity set to Berlin, Germany. Before https://github.com/mapbox/carmen/pull/777/commits/0a2f2e88ea723fa01c362114c4888a870cc36f6e, the results like:
```
1. Beirut-Restaurant, Gleimstr. 25, Berlin, 10437, Germany
2. Beirut, Berlin, 12055, Germany
3. Beirut Night, Münzgasse 7, Leipzig, Sachsen 04107, Germany
4. Beirut Snack, Brückenstraße 1, Chemnitz, Sachsen 09111, Germany
5. Beirut, Lebanon
--- (below the fold)
6. بيروت, Beirut, Lebanon
```

Where the first 4 results were local pois (👍 expected with proximity enabled) but the 5th result is Beirut the region and the 6th is Beirut the city. According to our [squishy functionality in carmen](https://github.com/mapbox/carmen/blob/master/lib/geocoder/verifymatch.js#L316), the city of Beirut should return before the region even though the region has a higher score. However, this wasn't happening on my new branch and when I looked at master, I saw that we returned the following results:
```
1. بيروت, Beirut, Lebanon        
2. Beirut-Restaurant, Gleimstr. 25, Berlin, 10437, Germany   
3. Beirut, Berlin, 12055, Germany             
4. Beirut Night, Münzgasse 7, Leipzig, Sachsen 04107, Germany         
5.  Beirut Snack, Brückenstraße 1, Chemnitz, Sachsen 09111, Germany
--- (below the fold)
6. Beirut, Lebanon
```
where Beirut city is first, then local pois, then the region of Beirut. This should have been failing on master given our new proximity implementation https://github.com/mapbox/carmen/pull/768 (the local pois should have been returning first). It turns out, if features are squishy, we bump the higher indexed lower score feature (Beirut city) above the lower indexed higher score feature (Beirut region) by adding the lower indexed feature's `carmen:score` to the higher indexed feature's `carmen:scoredist`. 
https://github.com/mapbox/carmen/blob/0ab559df0d6e764b85bcb1fe5280f7118deb1185/lib/geocoder/verifymatch.js#L388
This works fine without proximity (by now `carmen:score` === `carmen:scoredist`) but dramatically skews results with proximity enabled and means that feature will return before all other features. This is why Beirut region remained where it was in results while Beirut city moved to the top of results.

Changing to using a composite score in this PR meant that we were no longer sorting on `carmen:scoredist` directly, but now the squishy penalty now wasn't being applied at all :(. To fix this, I kept the score bump for the higher indexed feature, but with proximity enabled recalculated `carmen:scoredist` and `carmen:relevance`.

Here are the final results: local pois, then Beirut city, and finally Beirut region.
```
1. Beirut-Restaurant, Gleimstr. 25, Berlin, 10437, Germany
2. Beirut, Berlin, 12055, Germany
3. Beirut Night, Münzgasse 7, Leipzig, Sachsen 04107, Germany
4. Beirut Snack, Brückenstraße 1, Chemnitz, Sachsen 09111, Germany
5. بيروت, Beirut, Lebanon
--- (below the fold)
6. Beirut, Lebanon
```

Calculating `carmen:scoredist` and `carmen:relevance` is not an ideal solution to this issue; we should instead only calculate them once when we have all the information we need to apply penalties. Ticket landing soon, but here's a very rough outline from chat of what a refactor of verifyMatch could look like:
- pull contexts for all 20 features
- penalize, check squishy, check for duplicates, etc using full feature metadata
- calculate `carmen:scoredist` and `carmen:relevance`
- sort (once!)
- return the result to the user

lib/geocoder/verifymatch.js
```diff
@@ -384,8 +391,29 @@ function verifyContext(context, peers, strict, loose, indexes) {
...
     // EXCEPT in the case of an eligible higher-level feature that shares an
     // exact carmen:text value with something in its context (e.g. New York, New York)
-    if (squishy > 0)
-        context[0].properties['carmen:scoredist'] += squishy;
+    if (squishy > 0) {
+        const feat = context[0];
+        // Recalculate carmen:scoredist and carmen:relevance for smallest squishy feature
+        // with new higher score to ensure it returns before larger squishy features
+        if (options.proximity) {
+            feat.properties['carmen:scoredist'] = proximity.scoredist(
```
lizziegooding: Per https://github.com/mapbox/carmen/pull/777#issuecomment-437180870, we should refactor this to not calculate `scoredist`/`relevanceScore` twice

lib/util/proximity.js
```diff
@@ -180,3 +182,28 @@ function variance(scale, decay) {
...
+ * the street number if the user's query matches an address number and point in
+ * the address cluster.
+ * @param {Boolean} omitted True if the street's nearest endpoints match.
+ * @param {String} ghost True if the feature's carmen:score property is < 0.
+ * @return {Number} composite scoredist and relevance score between 0 and 1
+ */
+function relevanceScore(relev, scoredist, address, omitted, ghost) {
+    // add a slight penalty to features with carmen:address === null
+    // omitted geometries, or scores of < 0
+    if (address === null) relev = Math.max(0, relev - 0.005);
```
lizziegooding: I changed this from `0.01` to `0.005` after running downstream passrates. `0.01` meant that poi features 40 miles away were being returned before streets without address numbers 0.1 miles away. `0.001` meant that streets without address numbers were being returned before medium cities 90 miles away. The `0.005` penalty preserves the ideal behavior in both cases.

lib/util/proximity.js
```diff
@@ -180,3 +182,28 @@ function variance(scale, decay) {
...
+    if (omitted) relev = Math.max(0, relev - 0.01);
+    if (ghost) relev = Math.max(0, relev - 0.01);
+    // scale scoredist to a value between 0 and 1, give a weight of 0.4
+    const scoreDistWeight = ((scoredist - 1) / (121 - 1)) * 0.4;
```
lizziegooding: Do we want to store these constants in variables? Or are the comments/JS docs enough documentation to leave as is?

test/unit/util/proximity.scoredist.test.js
```diff
@@ -44,13 +42,13 @@ test('scoredist', (t) => {
...
-            { properties: { 'carmen:text': 'New York,NY', 'carmen:distance': 2426.866703400975, 'carmen:score': 79161, 'carmen:zoom': ZOOM_LEVELS.region, 'carmen:scoredist': 3.064512 } },
-            { properties: { 'carmen:text': 'New York,NY,NYC,New York City', 'carmen:distance': 2567.3550038898834, 'carmen:score': 31104, 'carmen:zoom': ZOOM_LEVELS.place, 'carmen:scoredist': 1.190303 } }
+            { properties: { 'carmen:text': 'New York Frankfurter Co.', 'carmen:distance': 0.4914344651849769, 'carmen:score': 1, 'carmen:zoom': ZOOM_LEVELS.poi, 'carmen:scoredist': 10.999837 } },
+            { properties: { 'carmen:text': 'New Yorker Buffalo Wings', 'carmen:distance': 0.6450163846417221, 'carmen:score': 3, 'carmen:zoom': ZOOM_LEVELS.poi, 'carmen:scoredist': 10.999757 } },
+            { properties: { 'carmen:text': 'New York,NY', 'carmen:distance': 2426.866703400975, 'carmen:score': 79161, 'carmen:zoom': ZOOM_LEVELS.region, 'carmen:scoredist': 3.064524 } },
+            { properties: { 'carmen:text': 'New York,NY,NYC,New York City', 'carmen:distance': 2567.3550038898834, 'carmen:score': 31104, 'carmen:zoom': ZOOM_LEVELS.place, 'carmen:scoredist': 1.190309 } }
```
lizziegooding: I recalculated fixture `scoredist`s after changing the minimum possible score from 0 to -1 to account for ghost features.

lizziegooding:
## New changes
Lots of things changed in this PR! While I didn't originally intend for scope creep to happen, creating a composite score using relevance and scoredist meant also accounting for [other feature properties we sort on](https://github.com/mapbox/carmen/blob/master/lib/geocoder/verifymatch.js#L398). 

### `null` carmen addresses
This happens when the user's query matches an address cluster, but not an address number/point. This may happen if that address number doesn't exist on the street, but more often happens when the user makes a query just for the street name. On master, these features are sorted to the bottom of results with equal relevance. I ended up penalizing these features slightly so streets with a matching number/point return first, and prominent features within the proximity bias return first, but these features surface before low prominence, med-high distance features. For example:
 
**master**
```
index  name                                           dist        score  scoredist  relev  address
-----  ---------------------------------------------  ----------  -----  ---------  -----  -------
1.     Waupaca                                        79.387345   30     10.788564  1
2.     Waupaca High School,primary school, secondary  87.778804   1      4.869482   1
3.     Waupaca Camping Park, LLC                      88.291561   0      4.826643   1
4.     Waupaca County Fairgrounds                     88.485261   0      4.810530   1
5.     Waupaca Bowl                                   88.587416   0      4.802046   1
```

**new weighting**
```
index  name                                           dist        score  scoredist  relev 
-----  ---------------------------------------------  ----------  -----  ---------  ----- 
1.     Waupaca                                        79.387345   30     10.788564  1     
2.     Waupaca Court                                  4.395998    0      10.976215  1    
3.     Waupaca Street                                 75.497285   0      5.954101   1     
4.     Waupaca High School,primary school, secondary  87.778804   1      4.869482   1      
5.     Waupaca Camping Park, LLC                      88.291561   0      4.826643   1      
```

Waupaca the city still gets surfaced first, but Waupaca Court (~ 4 miles away) goes from never returning in results to returning second. 

This change addresses https://github.com/mapbox/carmen/issues/778.

### omitted geometries

Similar to above, true for features where the beginning and the end of the address cluster's endpoints match. On master, these features are sorted to the bottom of results with equal relevance. I couldn't find an example in the wild of this happening, but figured I'd apply a similar penalty.

### ghost features

Ghost features are polygon features that overlap other polygon features in the same layer (usually `place` in the US) and represent alternative or local names. These features have a `carmen:score` of -1 and on master, are sorted to the bottom of results with equal relevance.

https://github.com/mapbox/carmen/blob/7be3c3687b5696fc17f7cdae626356cfd55d804d/lib/geocoder/verifymatch.js#L202

Ghost features are now penalized but still get to participate in `scoredist` calculations/ranking. I read up on why we originally implemented ghost features (https://github.com/mapbox/mapbox-places/issues/4, https://github.com/mapbox/hey/issues/1825), and while I agree that we should not be surfacing these features in reverse geocodes or in the context of forward geocodes (we should use the official feature name instead) I don't think they should be quite so hard to surface. It's reasonable to assume that if a user is looking for a place with an alternative name, that place should surface after non-ghost features with the same name, but otherwise be readily discoverable. As one example, this was causing a query for `Sunset Dis` near `Sunset District` (👻 ) in SF to return `Sunset Disco` in the Dominican Republic first. Ghost features get an additional slight penalty in that they have the lowest possible `carmen:score`, -1, which affects the final calculated `scoredist`. 

### squishy features
See https://github.com/mapbox/carmen/pull/777#issuecomment-437180870.

## Punted/stretch goals
- Adding test cases for typo and word frequency penalties. I didn't run into these issues while testing, and while the existing behavior should improve results with proximity enabled, we should definitely go back and add specific tests cases.
-  Use `carmen:relevance` as the final relevance displayed to the user in the response when proximity is enabled. ⚠️ Even though results will now be sorted by a different value, we will still be displaying the  same spatialmatch relevance property to the user. This means that it's possible for results to be in non-sequential order based on the `relevance` property in the response. cc @alianthes @karenell, will follow up before merging as well. ⚠️ I debated exposing `carmen:relevance` in the response, but the distribution of values is very different than spatialmatch relevance, this would be unintuitive if the user were comparing proximity vs non-proximity results, and we know some users rely on certain relevance ranges when assessing result quality.

lib/util/proximity.js
```diff
@@ -180,3 +182,28 @@ function variance(scale, decay) {
...
+    // give relevance score a weight of 0.6
+    const relevWeight = relev * 0.6;
```
aarthykc: Maybe I am missing documentation somewhere, could you explain the decision behind giving relevance score a weight of 0.6? You mention in the original post that these thresholds we picked arbitrarily. Do you have examples to prove that this weighting is better?

lib/util/proximity.js
aarthykc: I think JS docs for this is enough

lib/util/proximity.js
```diff
@@ -180,3 +182,28 @@ function variance(scale, decay) {
...
+ * the address cluster.
+ * @param {Boolean} omitted True if the street's nearest endpoints match.
+ * @param {String} ghost True if the feature's carmen:score property is < 0.
+ * @return {Number} composite scoredist and relevance score between 0 and 1
+ */
+function relevanceScore(relev, scoredist, address, omitted, ghost) {
+    // add a slight penalty to features with carmen:address === null
+    // omitted geometries, or scores of < 0
+    if (address === null) relev = Math.max(0, relev - 0.005);
+    if (omitted) relev = Math.max(0, relev - 0.01);
```
aarthykc: Could you also add comments (maybe on the the PR?) that explain why a penalty of 0.01 like you've explained the case above? We're there reasons why didn't chose a penalty lower or higher?

lib/geocoder/verifymatch.js
```diff
@@ -199,14 +199,21 @@ function verifyFeatures(query, geocoder, spatialmatches, loaded, options) {
     // Set a score + distance combined heuristic.
     for (let k = 0; k < result.length; k++) {
         const feat = result[k];
-        // ghost features don't participate
-        if (options.proximity && feat.properties['carmen:score'] >= 0) {
+        if (options.proximity) {
```
aarthykc: Right, this is the same as the comment that we shouldn't calculate the `RelevanceScore` and the `scoredist` twice. Do you think it would have any performance impact?

test/unit/util/proximity.relevanceScore.test.js
```diff
@@ -0,0 +1,90 @@
+'use strict';
+const test = require('tape');
+const proximity = require('../../../lib/util/proximity');
+
```
aarthykc: Great job on the tests 👏 ! 😍 Made it super easy to navigate + understand penalties etc.

test/unit/util/proximity.scoredist.test.js
aarthykc: Makes sense 👍

lib/geocoder/verifymatch.js
aarthykc: Could you test queries that trigger squishy behaviour and see if they take much longer to return results?

lib/util/proximity.js
lizziegooding: Yes!! I chose a penalty of `0.01` because it resulted in the desired behavior for the test cases in https://github.com/mapbox/carmen/blob/72ee650330c64c5c7baf9c58916d67bf84d705f4/test/unit/util/proximity.relevanceScore.test.js. These test cases were drawn from real failures I was seeing in downstream passrates. I started with https://github.com/mapbox/carmen/blob/72ee650330c64c5c7baf9c58916d67bf84d705f4/test/unit/util/proximity.relevanceScore.test.js#L41-L57. A penalty of `0.01` meant that proximate features were returned first, even with the language penalty applied, but high score, distant features without the language penalty were close behind. We should definitely consider changing this threshold in the future; I found that it fixed this test case and improved passrates for language-enabled tests downstream, but this definitely was not an exhaustive process to arrive at this number.

lizziegooding:
To understand the affect of calculating scoredist and relevanceScore twice, I ran the query for Beirut which initially surfaced this issue with each of the proximity points we use in downstream passrates. The difference in time varies between master and this branch and I'd argue is too small to be of consequence.

### Proximity point in Berlin
**Request:** `curl "http://localhost:8889/geocoding/v5/mapbox.places/Beirut.json?access_token=pk.eyJ1IjoiYXBpa2V5dXNlciIsImEiOiJhYmNkZWZnIn0.ENonA568sn1Xp32NR6CvxA&proximity=13.4085,52.5172&_=1541629302241"`
**Time:** + 4 ms

`master`

```
real	0m0.138s
user	0m0.005s
sys	0m0.000s
```

`bump-language-penalty`

```
real    0m0.142s
user    0m0.006s
sys     0m0.000s
```

### Proximity point in Washington, DC
**Request:** `curl "http://localhost:8889/geocoding/v5/mapbox.places/Beirut.json?access_token=pk.eyJ1IjoiYXBpa2V5dXNlciIsImEiOiJhYmNkZWZnIn0.ENonA568sn1Xp32NR6CvxA&proximity=-77.0405,38.9094&_=1541629302241"`
**Time:** - 7 ms

`master`

```
real	0m0.085s
user	0m0.005s
sys	0m0.000s
```

`bump-language-penalty`

```
real	0m0.078s
user	0m0.000s
sys	0m0.005s
```

### Proximity point in New York
**Request:** `curl "http://localhost:8889/geocoding/v5/mapbox.places/Beirut.json?access_token=pk.eyJ1IjoiYXBpa2V5dXNlciIsImEiOiJhYmNkZWZnIn0.ENonA568sn1Xp32NR6CvxA&proximity=-73.9908,40.7361&_=1541629302241"`
**Time:** + 2 ms

`master`

```
real	0m0.081s
user	0m0.006s
sys	0m0.000s
```

`bump-language-penalty`

```
real	0m0.083s
user	0m0.005s
sys	0m0.000s
```

### Proximity point in San Francisco
**Request:** `curl "http://localhost:8889/geocoding/v5/mapbox.places/Beirut.json?access_token=pk.eyJ1IjoiYXBpa2V5dXNlciIsImEiOiJhYmNkZWZnIn0.ENonA568sn1Xp32NR6CvxA&proximity=-122.4405,37.7641&_=1541629302241"`
**Time:** - 1 ms

`master`

```
real	0m0.083s
user	0m0.006s
sys	0m0.000s
```

`bump-language-penalty`

```
real	0m0.082s
user	0m0.000s
sys	0m0.006s
```

Issue #787, opened 2018-11-14
Pass proximity bias radiuses as a config property


lizziegooding:
### Context
In https://github.com/mapbox/carmen/pull/768 we refactored how we bias results with proximity enabled. The radius we use to bias results is currently hard-coded based on the zoom level of the index the feature belongs to:

https://github.com/mapbox/carmen/blob/6cdf5279422f0eecab8bc04991b198b8bc2ddfb1/lib/util/proximity.js#L124-L142

We should allow the user to pass the desired proximity radius for an index as a value in the tilejson config. This will allow us to set more granular radiuses for different feature types.

### Next Steps
- [ ] refactor carmen to accept proximity bias radius option, use current radiuses as default/fallback
- [ ] update downstream tilejson configs with this new property
- [ ] rerun tests/tweak radiuses based on test results

cc @miccolis @aarthykc @apendleton @boblannon

Nmargolis:
Support for a proximity radius specified on the index level was added in #836. The initial approach did pass that radius along to the methods for calculating scoredist in verifymatch, but produced suboptimal results, so the decision was to only pass the proximity radius to coalesce.

Leaving this ticket open because we may decide in the future to revisit the approach of passing a custom proximity radius to scoredist calculations in Carmen.

Issue #782, opened 2018-11-13
Low-scoring features can't be meaningfully compared in proximity queries


apendleton:
### Context
[Recent work](https://github.com/mapbox/carmen/issues/748) has significantly changed how we weight distance and score in proximity queries. This has largely resulted in a major improvement in how proximity queries perform, but in some circumstances now weights distance so strongly that even very small differences in distance can overwhelm significant differences in score if both scores are relatively small.

Two concrete examples:
* Searching for "temple univ" from a point ~10 miles away returns the medical school before the main university POI because the point for the medical school is slightly closer, even though it has less than half the score:

Winning feature:

    Temple University School of Medicine
    "carmen:distance": 9.579823207460825,
    "carmen:score": 10,
    "carmen:scoredist": 10.888214812849403,

One we wanted (in position 4):

    Temple University
    "carmen:distance": 9.90006014601079
    "carmen:score": 24
    "carmen:scoredist": 10.881548576951698,

* "Industry city" finds a brewery first instead of the gallery it seems to be named after, for the same reason -- the feature we wanted has an 8x higher score, bit is .08 miles further away:

Winning feature:

    Industry City Distillery
    "carmen:distance": 6.787741592196162,
    "carmen:score": 1,
    "carmen:scoredist": 10.943453289245141,

vs. the one we wanted

    Industry City
    "carmen:distance": 6.866850499966609,
    "carmen:score": 8,
    "carmen:scoredist": 10.942598443311798,

It seems like we need to figure out some way to preprocess score values to exaggerate the relative differences in small-ish scores like these, without significantly perturbing the relationship between small scores and large scores. It sounds like @lizziegooding explored taking the log, but that was too disruptive, so maybe some other curve is best.

@mapbox/search

PR #790, opened 2018-11-16
[proximity] Sort features with similar distance by score

lizziegooding:
### Context
Ref https://github.com/mapbox/carmen/issues/782, follow up to https://github.com/mapbox/carmen/pull/768. Right now, distance is weighted more heavily than score, and small differences in distance overcome differences in score. 

I'm exploring the following implementations:
- change the weight given to score and distance when calculating `scoredist`. Right now, these are weighted evenly. We could change this to weigh score more heavily (e.g. 70/30)
- change the proximity radius for pois. As far as I can tell, this is issue is only surfacing for pois, which tend to have very small distance and score differences. I played with this a bit, and given the current gauss curve, this alone doesn't seem like it will be an effective way to weigh score more heavily, though is something we should do anyway https://github.com/mapbox/carmen/issues/787.
- normalize scores within feature types. This branch does this in proximity.score.test.js. I want to better understand the distribution of scores within feature types and how this would affect ranking between feature types before I would feel confident shipping this approach.
- create distance 'bands' to score within. This idea is pretty nascent, but it could be worth looking at the differences in feature distance or grouping them by distance/type and then sorting by score.

### Summary of Changes
- [x] add `Temple University` test case
- [x] normalize score by feature type

### Next Steps
- [ ] apply the normalize score by feature type method to more test fixtures
- [ ] try more of the above approaches

cc @miccolis @aarthykc @apendleton @boblannon

### DIFF
--- a/index.js
+++ b/index.js
@@ -278,8 +278,9 @@ function Geocoder(indexes, options) {
         // Find the min and max score of all features in all indexes
         this.minScore = this.byidx.reduce((min, source) => Math.min(min, source.minScore), 0) || 0;
         this.maxScore = {};
-        for (const type of Object.keys(this.byname)) {
-            const maxScore = this.byname[type].reduce((max, source) => Math.max(max, source.maxscore), 0) || 1;
+        for (const type of Object.keys(this.bytype)) {
+            const byType = this.byidx.filter((source) => source.type === type);
+            const maxScore = byType.reduce((max, source) => Math.max(max, source.maxscore), 0) || 1;
             this.maxScore[type] = maxScore;
         }
         this._error = err;
--- a/index.js
+++ b/index.js
@@ -277,8 +277,11 @@ function Geocoder(indexes, options) {
         }
         // Find the min and max score of all features in all indexes
         this.minScore = this.byidx.reduce((min, source) => Math.min(min, source.minScore), 0) || 0;
-        this.maxScore = this.byidx.reduce((max, source) => Math.max(max, source.maxscore), 0) || 1;
-
+        this.maxScore = {};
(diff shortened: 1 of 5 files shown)


index.js
```diff
@@ -273,8 +273,12 @@ function Geocoder(indexes, options) {
...
         // Find the min and max score of all features in all indexes
         this.minScore = this.byidx.reduce((min, source) => Math.min(min, source.minScore), 0) || 0;
-        this.maxScore = this.byidx.reduce((max, source) => Math.max(max, source.maxscore), 0) || 1;
-
+        this.maxScore = {};
+        for (const type of Object.keys(this.bytype)) {
+            const byType = this.byidx.filter((source) => source.type === type);
+            const maxScore = byType.reduce((max, source) => Math.max(max, source.maxscore), 0) || 1;
+            this.maxScore[type] = maxScore;
+        }
```
lizziegooding: I originally tried using the `bytype` index configuration, but our country and region indexes have multiple types specified in the `geocoder_types` property of their configs. This meant that country and region indexes were being added to `bytype.place` and overwhelming the real `maxScore` for place-only indexes. I also tried using the `byname` index, but `geocoder_name` properties [do not include pois in our configuration](https://github.com/mapbox/carmen/blob/master/docs/index-structure.md#geocoder_name-geocoder_type-and-combining-indexes) ( when `geocoder_type` === `poi`, `geocoder_name` === `address` ), so I couldn't use `byname`. Instead, I grabbed an array of unique types from `bytype` and filtered directly on `byidx` using each source's `type` property.

lizziegooding:
Stale, closing in favor of https://github.com/mapbox/carmen/pull/805

Issue #785, opened 2018-11-13
Make it possible to opt some feature classes out of the backy penalty


apendleton:
### Context
We currently apply a penalty called the `backy` penalty, where we demote candidate matches where the order of matched features in the query aren't strictly ascending (e.g. address, city, state, country, as is the US norm) or strictly descending (e.g., country, state, city, address, as is the norm in Japan). One problem: postcodes often don't seem to follow the rules as to where they go in the address as compared to their position in the hierarchy. They're typically written after the city even though they're more granular.

We should have a way to flag some indexes (postcode indexes, in this case) as being exempt from this penalty (should be a new tilejson configuration flag), and then we should modify the calculation for the penalty in https://github.com/mapbox/carmen/blob/96df087/lib/geocoder/verifymatch.js#L336-L367 to skip over matches from those indexes when determining whether a match is "backy" or not.

A particular candidate query is `Quincy MA 02169`, which prefers address results that don't suffer from this penalty (interpreting the number as a postpositional house number) over the city/state/zip result that's pretty clearly intended.

@mapbox/search

PR #791, opened 2018-11-19
Allow individual indexes to be exempt from backy penalty

boblannon:
### Context

Fixes https://github.com/mapbox/carmen/issues/785

### Summary of Changes
- [x] add failing test

### Next Steps

- [x] interpret `geocoder_ignore_order` property in a tilejson
- [x] pass new test

cc @mapbox/search


boblannon:
capturing some conversation with @lizziegooding:

Lizzie pointed out that with the solution as-is, we'd still apply backy penalties for queries like `123 Main St Quincy, Massachusetts 02169`, which have the same order problem with US postcodes, but the first feature in the context array doesn't have the `geocoder_backy_exempt` property.  

So in https://github.com/mapbox/carmen/pull/791/commits/9adcdb81c34a562d9d17340701295856f85d3f80 I tried applying the exemption for each comparison separately, but it didn't fix the `Quincy MA 02169` case. This was ...unexpected.

As it turns out, though, the reason that the postcode result for `Quincy MA 02169` is getting docked isn't because the postcode is out of order.  It's actually getting docked because the place and region features appear to be out of order.

We decide the value of `direction` by looking at the context’s first two matching features.  so when `Quincy MA 02169` returns a postcode, the query is understood as `<place> <region> <postcode>`.   When checking through the context, we see that the first feature is a postcode, and the second feature is a region.  That's when we decide that the user has submitted a query that's ordered in the opposite direction as our context array.  Specifically, we look at the first mask `001` and the second mask `100` and decide that we're moving backwards in the query string as we move upwards in the context.  So we set `direction=descending`. Then, when we see that the region layer matches with mask `010`, we decide that's out of order, and apply the backy penalty.

I've added some `console.log` statements to make this more clear. here's what that looks like for the case in question:

```
0 - 02169 mapbox.places-us-postcode-both, exempt=true
direction is undefined
0 - not applying backy
1 - Quincy mapbox.places-namopen-place-both, exempt=false
1 - 02169 mapbox.places-us-postcode-both, exempt=true
direction is undefined
direction set to descending
1 - not applying backy
2 - Massachusetts mapbox.places-aa-region-both, exempt=false
2 - Quincy mapbox.places-namopen-place-both, exempt=false
direction is descending
2 - applying backy
```

None of this even gets considered in when we match `02169 Quincy Lane, Linesville, PA`, though, because this logic only applies to context features that correspond to parts of the query string.  Since `Quincy MA 02169` matches only the address feature (`MA` matches `Lane -> La` through token replacement and fuzzy-matching), the backy penalty isn't ever considered.

....But if the query is the full postal address, like `2169 Quincy Lane, Linesville, PA 16424` it would apply backy, because practically all of the features in the context correspond to parts of the query string.

boblannon:
I think that f321007 should fix the problem.  basically it avoids making decisions about `direction` if one of the components is exempt from `backy`.  this may be overloading the new flag but i think it's pretty well motivated: we exempt layers from backy because we think that they are likely to appear at different positions in query strings.  that makes them a bad source of information when deciding direction of the query.

....it might argue for naming the flag differently, though.  like `geocoder_variable_order` or something?

apendleton:
@boblannon I *think* this is approximately you've ended up doing, but just to confirm: what I was suggesting in #785 was to totally ignore layers that had this flag set when contemplating anything about backy or directions. Like, treat the stack or query as if that element just wasn't there.

So:
> When checking through the context, we see that the first feature is a postcode

what I was suggesting was that we wouldn't see the first feature as being the postcode, we would treat that stack as if it just consisted of a place and region. So, direction would be `ascending` and we wouldn't penalize, but we'd still compare other layers to each other as before, and probably compare the things on either side of the skipped element to each other (again, just pretending like that exempt thing was absent and we went straight from the preceding to the following element). I think this last thing isn't happening in your version, but the benefits, if any, are probably marginal.

> it might argue for naming the flag differently, though

Yeah, I wouldn't have mentioned if you hadn't (feels silly to bikeshed about names), but in general either way I'd vote against naming configuration flags for particular opaque Carmen jargon that will be unfamiliar to newcomers as opposed to things that explain what the feature does, so I'd probably vote against `backy` in the name, in favor of your name or `ignore_query_order` or whatever (I guess that implies the opposite boolean state from yours though).

boblannon:
thanks, yeah, that's what i was going for, here.  since it looks unlikely that this will ship today i'll go ahead an rename the flag.  I agree that the jargon is a form of coupling that shouldn't be there.

PR #796, opened 2018-11-29
Don't penalize matches in indexes with an index-wide universal text flag

apendleton:
### Context
#592 added a flag for marking a whole index as containing all universal (not-language-specific text). We thought that meant matches in those indexes were not penalized for cross-language mismatch, but that turns out not to be what that flag currently does; instead, it affects whether or not such matches can be included in results for queries made with the `languageMade: strict` flag even if they're out-of-language. This PR makes it also do the thing we already thought it did.

### Summary of Changes
- [x] if a phrasematch result is from an index with `geocoder_universal_text` set to true, don't set the `language` property on the phrasematch object that gets passed into coalesce (this will be interpreted by carmen-cache as an all-language match)
- [x] test to make sure a penalty is not applied

### Next Steps
- [x] wait for downstream passrates to ensure that this has the desired effect

cc @mapbox/search

### DIFF
--- a/lib/geocoder/phrasematch.js
+++ b/lib/geocoder/phrasematch.js
@@ -127,25 +127,20 @@ module.exports = function phrasematch(source, query, options, callback) {
     }
 
     let languages;
-    if (options.language) {
-        languages = [options.language[0]];
+    if (source.geocoder_universal_text) {
+        // if all text in the index is unversal, we won't apply any language
+        // penalties at all
+        languages = null;
     } else {
-        // this may not be what we want to do here
-        // if effectively treats 'default' as its own language, and a lack of
-        // specified language flag as a search over this language
-        // we could alternatively search all languages here by setting this field
-        // to undefined, or emply some other strategy
-        languages = ['default'];
-    }
-    languages = languages.map((l) => {
+        const languageName = options.language ? options.language[0] : 'default';
         // use the built-in language if we have it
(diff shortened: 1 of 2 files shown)


lib/geocoder/phrasematch.js
```diff
@@ -220,12 +215,16 @@ function Phrasematch(subquery, weight, mask, phrase, scorefactor, idx, cache, zo
     this.phrase = phrase;
     this.scorefactor = scorefactor;
     this.prefix = prefix;
-    this.languages = languages;
     this.idx = idx;
     this.cache = cache;
     this.zoom = zoom;
     this.editMultiplier = editMultiplier || 1;
     this.proxMatch = proxMatch || false;
+    if (languages) {
```
apendleton: This is kind of a clunky construction, done partly out of laziness to avoid having to change carmen-cache. Currently, carmen-cache allows the language property to either not be set, or be set to an array. If it's set to an array, it interprets that as a list of valid language IDs. If it's not set, it interprets that as an all-language query. Any other value that's passed it throws a type error, and an empty array is interpreted as "no languages" (that is, apply a penalty to every match). Ideally it would also accept null, or an empty array would mean all languages, or something. But in the meantime, if we want to do an all-languages scan, we need to not pass this property in at all, rather than setting it to null or whatnot.

PR #800, opened 2018-12-05
Fix default behavior for --fuzzyMatch and --autocomplete in the CLI tool

boblannon:
### Context

When `autocomplete` and `fuzzyMatch` are not included as options, their values should be `undefined`, so that `geocode` can apply the appropriate default value here https://github.com/mapbox/carmen/blob/17a391aad6cdfa07c7256a3444a740997247dd62/lib/geocoder/geocode.js#L51-L52

### Summary of Changes
- [x] add failing test
- [x] using the `minimist` config, set the defaults for these params to `null`
- [x] convert `null` to `undefined` before submitting to `geocode`
- [x] pass test

cc @mapbox/search

### DIFF
--- a/bin/carmen.js
+++ b/bin/carmen.js
@@ -111,6 +111,10 @@ if (argv.reverseMode) {
 
 if (argv.routing) argv.routing = (argv.routing || false);
 
+// this has to be done because minimist only allows null as a default value,
+// but carmen checks whether these values `=== undefined`
+if (argv.autocomplete === null) argv.autocomplete = undefined;
+if (argv.fuzzyMatch === null) argv.fuzzyMatch = undefined;
 
 let load = +new Date();
 
--- a/bin/carmen.js
+++ b/bin/carmen.js
@@ -11,8 +11,11 @@ const Carmen = require('..');
 const settings = require('../package.json');
 const argv = require('minimist')(process.argv, {
     string: ['config', 'proximity', 'query', 'debug', 'types', 'tokens'],
-    boolean: ['geojson', 'stats', 'help', 'version', 'autocomplete', 'fuzzyMatch']
+    boolean: ['geojson', 'stats', 'help', 'version', 'autocomplete', 'fuzzyMatch'],
+    default: {'autocomplete': null, 'fuzzyMatch': null}
 });
+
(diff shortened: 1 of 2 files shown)


bin/carmen.js
```diff
@@ -11,9 +11,12 @@ const Carmen = require('..');
 const settings = require('../package.json');
 const argv = require('minimist')(process.argv, {
     string: ['config', 'proximity', 'query', 'debug', 'types', 'tokens'],
-    boolean: ['geojson', 'stats', 'help', 'version', 'autocomplete', 'fuzzyMatch']
+    boolean: ['geojson', 'stats', 'help', 'version', 'autocomplete', 'fuzzyMatch'],
+    default: { 'autocomplete': null, 'fuzzyMatch': null }
 });
 
+console.dir(argv);
```
mattficke: did you mean to include this?

bin/carmen.js
```diff
@@ -11,9 +11,12 @@ const Carmen = require('..');
 const settings = require('../package.json');
 const argv = require('minimist')(process.argv, {
     string: ['config', 'proximity', 'query', 'debug', 'types', 'tokens'],
-    boolean: ['geojson', 'stats', 'help', 'version', 'autocomplete', 'fuzzyMatch']
+    boolean: ['geojson', 'stats', 'help', 'version', 'autocomplete', 'fuzzyMatch'],
```
mattficke: do you need to include autocomplete and fuzzyMatch in here at all, or can you just leave them undefined (like `bbox`, `language`, `limit`, etc)?

bin/carmen.js
boblannon: heh, nope.  thanks

boblannon:
@mattficke i went ahead and moved them out of the config.  lmk what you think

bin/carmen.js
mattficke: ohh, didn't realize it'd need special handling like this. what about keeping it in `boolean` and setting the defaults to `true`?

boblannon: uh why yes, that works perfectly.  i can not figure out why i didn't try that in the first place........

PR #802, opened 2018-12-10
Category Matching flag

### LINKED ISSUE #801: Relevance is sometimes greater than 1
### Context
https://github.com/mapbox/carmen/pull/758 added a relevance boost to category terms to make them more likely to clear the spatialmatch limit. This is useful at this phase of the query pipeline, so we should keep this boost, but it introduces the possibility that in final results, total query relevances can exceed 1, which violates API expectations. Sometime late in the pipeline (perhaps after the last sort in verifymatch) we should clamp all relevances to make sure they're at most 1.

### Next Steps
- [ ] clamp the relevance someplace
- [ ] add or adjust a test to make sure categories' relevances aren't too high (there's an existing test that looks at relevances to see if the boost worked; it currently checks for a `1.01` relevance, so we may want to repurpose that test to look at a given feature's position in the results instead of its relevance number to see if it worked)

@mapbox/search

miccolis:
### Context
 
PR is a attempt to fix #801 by setting a flag on phrasematch objects  instead of boosting relevence.

### Next Steps

-  [x] needs down stream testing.

cc @mapbox/search

### DIFF
--- a/lib/geocoder/phrasematch.js
+++ b/lib/geocoder/phrasematch.js
@@ -178,18 +178,18 @@ module.exports = function phrasematch(source, query, options, callback) {
         // If the query matches an index with geocoder_categories set
         // and the query is one of the listed categories, give this
         // possible result a small bump
+        let catMatch  = false;
         if (source.categories) {
-            if (source.categories.has(phrase)) {
-                editMultiplier *= 1.01;
-            }
+            catMatch = source.categories.has(phrase);
         }
+
         // Set prefix to determine how carmen cache will handle autocomplete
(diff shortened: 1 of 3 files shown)


lib/geocoder/phrasematch.js
```diff
@@ -178,18 +178,18 @@ module.exports = function phrasematch(source, query, options, callback) {
         // If the query matches an index with geocoder_categories set
         // and the query is one of the listed categories, give this
         // possible result a small bump
+        let catMatch  = false;
```
mattficke: https://www.cats.org.uk/catmatch/

PR #795, opened 2018-11-28
[stage 1] Token replacement + autocomplete

lizziegooding:
### Context
Right now, we replace tokens in the user’s query, then autocomplete on the token-replaced phrases. This leads to unexpected behavior when the replaced token is at the end of the user’s query where we consider possible matches where the things we’ve replaced is the *beginning* of another word. For example: :

| User’s query      | Token replacement | Autocomplete match | Result returned to the user |
| ----------------- | ----------------- | ------------------ | --------------------------- |
| "sunset district” | "sunset”, “dt”    | "dt" → "dtown”     | "dtown party bus”           |
| "the white house" | "the", "white", "ho" | "ho" → "horse" | "the white horse" |
| "saint street" | "st", "st" | "st" → "stephen" |  "saint stephen" |
| "court house" | "ct", "ho" | "ho" → "hotel" | "court hotel" |
| "the house" | "the", "ho" | "ho" → "hound" | "The Hound Gentlemen's Clothiers" |
| "chase" | "ch" | "ch" → "chicago" | Chicago, IL |
|  "delaware county" | "delaware", "co" | "co" → "college" | "university of delaware" |

### Summary of Changes
- [x] adds acceptance test with `district` test case

### Next Steps
- [ ]  Track whether the last phrase in the query was token-replaced in carmen’s replaceToken function

cc @miccolis @aarthykc @apendleton @boblannon

PR #803, opened 2018-12-12
Incorporate new token replacement logic into carmen

### LINKED ISSUE #792: Write acceptance tests around interactions between token replacement and fuzzy matching
We need at least three categories of unit tests:

- [x] at least one test where at present, we replace and then fuzzy-match (so this would be “Rhode Island” → “Rhode Is” → “Rhode it”, or the 30th-street-station issue where "station" becomes "stn" which gets fuzzy-matched to "st")
- [x] at least one test where the pre-replacement word is misspelled (so like, user types “Main Strete,” but we can’t fuzzy-match it because the thing in the index is “Main St”) — these may not actually fail right now because of double-indexing
- [x] at least one test where potentially we’d want to not replace on autocomplete even though what the user has typed so far is replaceable. This is like: index “Fortenberry,” user types “Fort,” autocomplete fails because “Fort” → “Ft”.

### LINKED ISSUE #793: Wire new fuzzy-ified token replacement logic into carmen
The tickets mapbox/fuzzy-phrase#53, mapbox/fuzzy-phrase#54, and mapbox/node-fuzzy-phrase#15 propose moving some token replacement logic into the fuzzy-phrase library and out of Carmen. Here are the carmen-side things that would be needed to support that:

- [x] Write a utility function that can look at a token replacement and figure it if it swaps a whole word for another whole word, or does something else (changes part of a word, e.g., with a regex rule, or with `skipWordBoundaries`, or changes cardinality). This should have unit tests.
- [x] Modify the carmen constructor so that at instantiation time, we use the above function to separate token replacements into simple and complex buckets
- [x] Modify the index-time operations on tokes (getIndexableWhatever) to, for now, apply things from both buckets as we generate phrases
- [x] Modify the query-time logic to only apply the complex ones
- [x] Modify the indexing logic so that when it creates the FuzzyPhraseBuilder, it calls the now-exposed token-replacement-loader function and passes in all the simple replacements

### LINKED ISSUE #794: Index fewer token-replaced phrase variants
To be done after #793 lands, in conjunction with mapbox/fuzzy-phrase#57

Once we can handle detecting prefixes of token replacement candidates, we should index fewer phrase variants. Using the simple/complex distinction introduced in #793, we should at index time:
- [x] apply the complex replacements and index both variants (pre- and post-replacement)
- [x] apply the simple replacements but only index the post-replacement versions, rather than both; also consider removing the “custom inverse replacement” mechanism altogether (would require downstream changes)

---

## Update:
Okay turns out I ended up tackling mapbox/fuzzy-phrase#57 along with the first phase of fuzzy work, so we'll want to go ahead and do all of this now. For background, I would suggest reading https://github.com/mapbox/carmen/pull/630 and https://github.com/mapbox/carmen/pull/652 , but essentially those PRs introduced mechanisms for indexing more than one variation of a phrase -- when we do token replacements, we keep both the pre-replacement and post-replacement version. We also allow this to be bidirectional, so we generate an extra token replacer that has both forward replacements (e.g., "road" => "rd") and the same replacement backwards ("rd" => "road"), and allows for specifying custom behavior if the reverse direction is ambiguous (so, if "st" could be "street" or "saint," consumers can provide a custom function which guesses which was intended).

Pretty much all of this should go away. For simple token replacements (that is, one word replaced by another word), we should just apply the replacements unidirectionally, and not keep both versions of the phrase. That is, for the string "main street" we should *only* index the string "main st". We can also probably do away with the custom reverse replacer notion altogether.

We'll still want to do double indexing for complex replacement situations, so we do probably need to keep the inverse replacer mechanism around, but it will need to be significantly reduced in scope. All of this means we may need different token replacers for simple and complex replacements, so that one can just be a applied and the other can have its possibilities enumerated. An example of this kind of complex cast would be umlaut replacement in German, where we will want to continue to index both variants since fuzzy-phrase isn't smart enough to handle these sorts of more sophisticated swaps.

Another good place to start might be https://github.com/mapbox/carmen/blob/master/test/unit/text-processing/termops.getIndexableText.test.js -- we'll want almost all of these getIndexableText calls to produce a single output, with the umlaut/ß cases being the exception.

apendleton:
### Context
This branch incorporates the latest work in fuzzy-phrase (mapbox/fuzzy-phrase#58 , mapbox/fuzzy-phrase#62 , mapbox/fuzzy-phrase#64 , mapbox/fuzzy-phrase#66) into carmen. The goal is to separate out "simple" token replacements (replacements that swap a single word for a single other word) from "complex" ones, and change how they're applied. At index time, both simple and complex tokens are now applied, but simple tokens are applied in a unidirectional manner (no more double indexing of "st" and "street" for autocomplete reasons, as this autocomplete case can now be handled at query time by fuzzy-phrase). At query time, only complex replacements are applied, as simple replacements are now handled inside fuzzy-phrase.

#### Business logic
- [x] extend @aarthykc and @lizziegooding 's work for separating out simple and complex tokens; tokens that include word boundaries are now in the complex class (the presence of such replacements is problematic, but we're holding off on banning them for now)
- [x] explicitly separate out simple and complex tokens into buckets
- [x] separate the `createReplacer` function into a `createComplexReplacer` function, which mostly behaves like the current one except always assumes the "complex" behavior where the old one varied from simple to complex, and `createSimpleReplacer` which is a radically simplified, much faster replacement function based on @yhahn's work in #664.
    - [x] update -- so I wrote this function and then ended up changing the index-side getIndexablePhrases code to not even use regex replacement at all, and instead just look each word up in a map. But I'd already written it and tested it, and it's better than what was there before, so... I guess I'll leave it?
- [x] modify getIndexablePhrases to take both replacers, and only `enumerate` on the complex ones while just applying the simple ones
- [x] at query time, only apply the complex replacements but not the simple ones
- [x] update all query-time fuzzy-phrase calls to use the new format for specifying prefix types both for input and output
- [x] use the prefix-related output from fuzzy-phrase (including word-boundary information) as input into carmen-cache, replacing the JS source of this information introduced in #795 .
- [x] update all the tests around the new expected outputs and to use the new function names and signatures
    - [x] note: at present, none of the token replacement functions can produce the `fromLastWord` outputs introduced in #795 -- the complex replacements never did, and the simple replacements are now only used at index time where that's not relevant, so the new version of the simple replacers can't either -- we instead derive this information from fuzzy-phrase. I've left the underlying mechanism in place since I suspect we might subdivide the complex replacement cases and bring it back under some circumstances, but in the meantime I've eliminated some tests of this specific functionality for the moment since they don't work anymore. We can bring these tests back if/when we reintroduce this functionality on the JS side; cc @lizziegooding 

fixes #792, fixes #793, fixes #794

### Next Steps
- [x] fix the benchmarks
- [x] make some more efficiency improvements to the one token replacements are applied
- [x] write unit tests around, at a minimum, the token categorizer function
- [x] write acceptance tests around the new functionality related to the interaction of fuzzy search and token replacement; refs #792

cc @mapbox/search


lib/text-processing/termops.js
```diff
@@ -268,8 +268,8 @@ function getIndexableText(replacer, globalTokens, doc, defaultLanguages, categor
 
     // set up the replacer
     let combinedReplacer = [];
-    if (globalTokens && globalTokens.length) combinedReplacer = combinedReplacer.concat(globalTokens);
-    if (replacer && replacer.length) combinedReplacer = combinedReplacer.concat(replacer);
+    if (globalReplacer && globalReplacer.length) combinedReplacer = combinedReplacer.concat(globalReplacer);
+    if (complexReplacer && complexReplacer.length) combinedReplacer = combinedReplacer.concat(complexReplacer);
```
boblannon: this variable naming threw me when reading further down. I forgot that "combined" meant global + complex, and thought it meant global + complex + simple.  I know the variable name is not new in this PR, but maybe we could change it to be more specific, like `globalAndComplexReplacer` or something hideous like that

lib/geocoder/phrasematch.js
```diff
@@ -184,10 +192,11 @@ module.exports = function phrasematch(source, query, options, callback) {
...
         // Set prefix to determine how carmen cache will handle autocomplete
-        let prefix = PREFIX_SCAN.disabled;
-        if (subquery.ender && !(source.geocoder_address && termops.isAddressNumber(subquery))) {
-            if (replaced.lastWord) prefix = PREFIX_SCAN.word_boundary;
-            else prefix = PREFIX_SCAN.enabled;
+        let prefix;
+        if (source.geocoder_address && termops.isAddressNumber(subquery)) {
+            prefix = PREFIX_SCAN.disabled;
+        } else {
+            prefix = subquery.ending_type;
```
boblannon: it'd be nice to have `carmen:prefix` changed to `carmen:ending_type` but i'm pretty sure that would require a carmen-cache change, so doesn't need to be in-scope here

lib/geocoder/spatialmatch.js
```diff
@@ -153,14 +154,32 @@ function collapseToArchetypes(phrasematchResults) {
...
+        for (const group of collapsed) {
+            if (
+                group.exemplars[0].subquery.length === 1 &&
+                group.exemplars[0].subquery.edit_distance === 0 &&
+                group.prefix !== PREFIX_SCAN.disabled &&
+                group.exemplars.length > 2
+            ) {
+                group.editMultiplier *= .99;
+            }
+        }
```
boblannon: it'd be nice to have some idea of what the impact of this is.  I agree that it's a good heuristic, but worry that there may be some less-than obvious class of features that would get unfairly docked as matches here.  Not requesting a  change, but maybe doing some comparison runs could help.

lib/text-processing/termops.js
```diff
@@ -315,7 +315,9 @@ function getIndexableText(replacer, globalTokens, doc, defaultLanguages, categor
         variants = variants.slice(0, 8);
 
         for (const variant of variants) {
-            const tokens = tokenize(variant);
+            const encoded = encodableText(variant);
+            // do simple token replacements without regexes -- just look them up in a dict
+            const tokens = tokenize(encoded).map((token) => simpleReplacer.tokens.get(token) || token);
```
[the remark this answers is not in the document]
apendleton: So... hm. They're actually not the same. WORD_BOUNDARY includes periods and apostrophes, but in tokenize, periods and apostrophes are replaced with *nothing*, and everything else is used as a boundary. I actually hadn't realized this, so it's a good flag.

I *think* we're still safe here, because anything with apostrophes or periods in it on either the from or to side will get binned into the complex bin and shouldn't affect simple stuff one way or the other, but I'm not 100% sure this is the case. I'm not sure what the best way to test it is though. If you have thoughts I'd be interested to see them.

lib/text-processing/termops.js
apendleton: No objection to changing it.

lib/geocoder/phrasematch.js
apendleton: Yes, it would indeed require a carmen-cache change, so that's why I didn't change it. Worth putting a pin it for later though.

lib/geocoder/spatialmatch.js
apendleton: > hat there may be some less-than obvious class of features that would get unfairly docked as matches here

I don't think so. Before token replacement was introduced, there was no way to get back more than one result from the same index in the same position with an edit distance of 0. There could, definitionally, be at most one word that was spelled exactly as the word in the query was spelled. Now there can be more than one because all token replacement possibilities mirror the edit distance of the thing they're replacing, and in the prefix/terminal-token case specifically we can produce more than one, because we can emit a full word at edit distance 0 for every replacement candidate that starts with the given prefix. There's no other way to produce more than one, though, so that's why I'm checking that the edit distance is 0 *and* we're at the end (we have a prefix match type) *and* there are more than 2 of them *and* that it's a single word.

The alternative would be to explicitly return metadata from fuzzy-phrase about which words had been token replaced, but this would be a much more invasive change, and I don't think it's necessary.

lib/text-processing/token.js
```diff
@@ -23,36 +23,39 @@ const WORD_BOUNDARY = require('../constants.js').WORD_BOUNDARY;
  * Create a per-token replacer
  *
  * @access public
- * @name createReplacer
+ * @name createComplexReplacer
  *
  * @param {object} tokens - tokens
```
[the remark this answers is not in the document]
apendleton: Fine by me.

lib/text-processing/termops.js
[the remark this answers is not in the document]
apendleton: Thought about this a bit more -- I'm actually now pretty confident that the fact that it *doesn't* use the `simpleReplacer.to` function is actually necessary precisely because of this difference -- at query time, replacement *also* happens on a word by word basis after tokenization (in fuzzy-phrase), so doing it this way here ensures parity whereas using the `WORD_BOUNDARY` regex wouldn't -- it would allow for words that had commas or apostrophes in them to be replaced at query time, after those characters are stripped, but not at index time. So it turns out the thing I was thinking was mostly motivated by performance turns out to also probably be better behavior.

It's still pretty funky, though. We probably just shouldn't have the simple replacer thing produce a regex-style replacer at all, and just produce a dictionary, so that we don't maintain the fiction that it works either way. There are still a few places (mostly tests, but also the frequency analysis code I think) that still use it the other way though, so that would be a pretty invasive change that could maybe wait for a follow-up.

PR #807, opened 2019-01-03
Switch back to Travis

miccolis:
Like the title says

cc @mapbox/search


apendleton:
Are we meant to be doing Travis-the-app or Travis-the-... whatever the old thing was? We're inconsistent across repos about this at the moment, with fuzzy-phrase using the new flavor and carmen-cache using the old one (and node-fuzzy-phrase using *both* for some reason, because of some configuration snafu nobody has bothered to fix).

PR #808, opened 2019-01-16
Intersection search

aarthykc:
### Context
Adding unit tests for intersection search and ability to  index and then search for an intersection. 

### Explanation 
The problem that we currently have with carmen especially with addresses is that we index a lot of features that have the same relevance and score and because of the limits we have in place for verifymatch and spatialmatch, a specific feature we'd like to surface, just by virtue of being lower down in the sort id filtered out. 
While thinking of how to solve this problem I started looking at how we index house numbers. We don’t create a new feature for every house number: we have one feature and an `addressnumber` array with all the house numbers. Essentially, for intersection search we'll want to take a similar approach, i.e. we want fewer features which more densely populated. 

The initial data spec that @ingalls worked on, looked something like this: 

    {
        id: 3,
        properties: {
            'carmen:text': '9th st nw and 15th st nw,15th st nw and 9th st nw',
            'carmen:center': [0,0],
        },
        geometry: {
            type: 'Point',
            coordinates: [0,0]
        }
    };

The problem with this approach however was that because of the aforementioned limits we have in place, we only ever surface intersections and only when the user completely types out the street name like `9th Street Northwest` do we surface the actual street. The underlying problem being that we don’t have a good way to know whether the user is searching for an intersection or just the street. This PR aims to trigger intersection search only after the user starts typing `and`. 

What new approach are we using:

### Index Time:
1. At indexing time, we index intersections in the `carmen:intersections` in the following manner: 
`'carmen:intersections': [['highway number 2']]` → `['+intersection highway number 2 ,']`
We use `+intersection` as a special tag for intersections and we use the `,` as a delimiter between the intersection and the actual street name. 

2. Since, all features have the same score and relevance for addresses it is likely that intersections would get filtered out even when the user is searching for an intersection, to get around this we increase the weights for intersections to make sure that they are surfaced as results when someone is looking for an intersection. 
```
    if (weights[i] === -2) {
        weights[i] = 0.4;
    }
    else {
        weights[i] = Math.max(weights[i] / weightsum) * 0.6;
    }
```
tokens that have a `+intersection` are given a constant weight of 0.4 and the rest of the tokens are scaled between 0-0.6.

3. We standarize Multipoints to GeometryCollections to make it easier to generate intersection data and also reduce the number of features indexed/ size of the index. Therefore, now intersections don’t have to be a separate address cluster.

### Query Time:
4. At query time, we look for the keyword  ‘and’ which we use as a trigger for intersection search. For example, `'9th street northwest and f street northwest'` gets tokenized to `['+intersection', '9th', 'street', 'northwest', ',', 'f', 'street', 'northwest']`
5. We check whether the search is an intersection search by checking whether the cover text has the special tag `+intersection` if it does then we look up the corresponding geometry in the `GeometryCollection`. 
6. Once we have the geometry and the intersection name we stitch the place_name together which return to the user.

### Data Spec:
1. For the data design, let’s assume: 
- 9th Street Northwest which is also known as US Road, Highway Number 6
- F Street Northwest which is also known as Highway Number 4

#### Description: 

- 9th Street Northwest intersects with F Street Northwest at [0,2] and Highway Number 2 at [0,1]
- The aliases for 9th Street Northwest go in carmen:text separated by a comma. While the streets 9th Street Northwest intersects with goes in the carmen:intersections property.
- The coordinates in the geometry correspond to the point of intersection of the street in carmen:intersections and the feature in carmen:text.
    
```    
    Example data spec for intersections:
    {
        const address = {
            id:2,
            properties: {
                // Synonyms of the feature go in 'carmen:text'
                'carmen:text': 'US Road,Highway Number 6,9th Street Northwest',
                'carmen:center': [0,0],
                // intersections with the feature go here
                'carmen:intersections': ['F Street Northwest', 'Highway Number 2']
            },
            geometry: {
                type: 'MultiPoint',
                // coordinates correspond to the street in 'carmen:intersection'
                // for example, F Street Northwest and 9th Street Northwest intersect at [0,2]
                coordinates: [[0,2], [0,1]]
            }
        };
        queueFeature(conf.address, address, t.end);
    });
```

- Intersections can also represented as a GeometryCollection by adding  `carmen:intersections`: `[null,['1st Avenue', '2nd Avenue', '3rd Avenue'], null]` property
- And the corresponding coordinates in the geomteries property
-  The position in the geometries array corresponding to the non null array value in `'carmen:intersections'`, similar to addressnumbers
- We standardise a multipoint intersection feature as a GeometryCollection to allow intersections to be part of the same address cluster and avoid having to index a feature and it’s intersection separately

```
    {
        properties: {
            'carmen:text': 'Main Street Northwest',
            'carmen:addressnumber': [[1, 2, 3], null, null],
            'carmen:intersections': [null,['1st Avenue', '2nd Avenue', '3rd Avenue'], null],
            'carmen:rangetype': 'tiger',
            'carmen:parityl': [[], [], []],
            'carmen:parityr': [[], [], []],
            'carmen:lfromhn': [[], [], []],
            'carmen:rfromhn': [[], [], []],
            'carmen:ltohn': [[], [], []],
            'carmen:rtohn': [[], [], []]
        },
        geometry: {
            type: 'GeometryCollection',
            geometries: [{
                type: 'MultiPoint',
                coordinates: [[1,1], [2,2], [3,3]] <--- addressnumber MultiPoint array is 1st, since addressnumber: [[1, 2, 3], null, null]
            },{
                type: 'MultiPoint',
                coordinates: [[2,2], [3,2], [3,3]] <--- intersections MultiPoint array is 2nd, since intersections: [null,['1st Avenue', '2nd Avenue', '3rd Avenue'], null]
            },{
                type: 'MultiLineString',
                coordinates: [[[1,1], [2,2], [3,3]]]
            }]
        }
    };
```

- If there is more than one name for F Street Northwest and it intersects with 9th Street Northwest, the alias is added to carmen:intersections and the corresponding coordinates in the coordinates property of the geometry.

### Summary of Changes
- [ ] Adds unit tests for intersection search 
   - Does this new approach surface streets before intersections 
   - Are we able to find intersections?
   - Can we find intersections for abbreviated versions of 9th Street Northwest and F Street Northwest like 9th st nw and F st nw
- [ ] Adds special indexing intersections
- [ ] Adds geomtery lookup for intersections in the geometry collection
- [ ] Support to return the name of the street and the intersection
- [ ] New weighting for when the tokens have `+intersection`
- [ ] Don’t hard code the replacement `and`, have a way to specify this as a part of the index
- [ ] Trigger an intersection search only after the user types 'and'

Next Actions:
- [ ] Check with real index data generated by pt2itp cc @ingalls 
- [ ] Can this approach of tagged text we used beyond `and` for other languages or other query patterns such as `between X and Y`
- [ ] Global token support for allowing users to query for Street A & Street B

cc @mapbox/search

### DIFF
--- a/lib/geocoder/ops.js
+++ b/lib/geocoder/ops.js
@@ -47,9 +47,8 @@ function getPlaceName(context, formatString, language, languageMode, matched) {
     // Use geocoder_format to format output if applicable
     if (!formatString || formatString === true || formatString === 1) {
         place_name = (
-            (feat.properties['carmen:address'] ?
-            feat.intersections ?
-            feat.properties['carmen:address'] + ' and ' : feat.properties['carmen:address'] + ' ' : '') +
+            (feat.properties['carmen:intersection'] ? feat.properties['carmen:intersection'] + ' and ' : '') +
+            (feat.properties['carmen:address'] ? feat.properties['carmen:address'] + ' ' : '') +
             context
                 .filter((f) => { return matched || featureMatchesLanguage(f, { language: [language], languageMode: languageMode }); })
                 .map((f) => { return (matched ? f.matching_text : undefined) || closestLang.getText(language, f.properties).text; })
--- a/lib/geocoder/ops.js
+++ b/lib/geocoder/ops.js
@@ -82,7 +82,8 @@ function getPlaceName(context, formatString, language, languageMode, matched, ge
                     if (templateSubprop === '_name') {
                         if (f.properties['carmen:intersection']) {
                             const intersectionPrefix = feat.properties['carmen:intersection'] + ' and ';
-                            val = intersectionPrefix + getIntersectionName(geocoder, f.properties);
+                            const intersectionSuffix = getIntersectionStreetName(geocoder, f.properties);
+                            if (intersectionSuffix) val = intersectionPrefix + intersectionSuffix;
                         }
                         // `name` is a hardcoded subproperty that maps to carmen:text
                         else if (matched && f.matching_text) val = f.matching_text;
@@ -342,7 +343,7 @@ function getMatchingText(item, geocoder, requestedLanguage) {
 * @param {object} properties A feature properties object
 * @return String returns the correct carmen:text synonym based on the input query
 */
-function getIntersectionName(geocoder, properties) {
+function getIntersectionStreetName(geocoder, properties) {
     if (properties['carmen:intersection'] && properties['carmen:query_text'] !== undefined) {
         let featureSynonymMatch;
         const source = geocoder.byidx[properties['carmen:idx']];
--- a/lib/geocoder/ops.js
+++ b/lib/geocoder/ops.js
@@ -41,20 +41,19 @@ function getFormatString(context, format, language) {
  * @param {Boolean} matched If set, use matched text in place name
  * @return {String} formatted place_name string
  */
-function getPlaceName(context, formatString, language, languageMode, matched) {
+function getPlaceName(context, formatString, language, languageMode, matched, geocoder) {
     const feat = context[0];
     let place_name;
     // Use geocoder_format to format output if applicable
     if (!formatString || formatString === true || formatString === 1) {
         let prefix;
-        if (feat.properties['carmen:intersection']) prefix = feat.properties['carmen:intersection'] + ' and ';
-        else if (feat.properties['carmen:address']) prefix = feat.properties['carmen:address'] + ' ';
+        if (feat.properties['carmen:address']) prefix = feat.properties['carmen:address'] + ' ';
         else prefix = '';
         place_name = (
             prefix +
             context
                 .filter((f) => { return matched || featureMatchesLanguage(f, { language: [language], languageMode: languageMode }); })
-                .map((f) => { return (matched ? f.matching_text : undefined) || (f.properties['carmen:intersection'] ? getIntersectionName(f.properties) : undefined) || closestLang.getText(language, f.properties).text; })
+                .map((f) => { return (matched ? f.matching_text : undefined) || closestLang.getText(language, f.properties).text; })
                 .join(', ')
         ).trim();
     } else {
@@ -81,8 +80,13 @@ function getPlaceName(context, formatString, language, languageMode, matched) {
                 // If we found a match between the template property and one within the returned context
                 if (templateProp === carmenProp) {
                     if (templateSubprop === '_name') {
+                        if (f.properties['carmen:intersection']) {
+                            const intersectionPrefix = feat.properties['carmen:intersection'] + ' and ';
+                            val = (intersectionPrefix + context.map((f) => { return (f.properties['carmen:intersection'] ? getIntersectionName(geocoder, f.properties) : undefined); }));
+                        }
                         // `name` is a hardcoded subproperty that maps to carmen:text
-                        val = matched && f.matching_text ? f.matching_text : closestLang.getText(language, f.properties).text;
+                        else if (matched && f.matching_text) val = f.matching_text;
+                        else val = closestLang.getText(language, f.properties).text;
                         formatString = formatString.replace('{' + templateProp + '._name}',val);
                     } else if (templateSubprop === '_number') {
                         // `address` is a hardcored subproperty that maps to carmen:address
@@ -108,6 +112,7 @@ function getPlaceName(context, formatString, language, languageMode, matched) {
         formatString = formatString.replace(/\{.+?\}/g, '').replace(/ , /g,', ').replace(/ {2}/g,' ').replace(/, -/,',').replace(/, ,/g,'').replace(/^,/,'').replace(/,,/,',').trim().replace(/,$/,'');
         place_name = formatString;
     }
+
     return place_name;
 }
 
@@ -148,8 +153,10 @@ function toFeature(context, format, languages, languageMode, debug, geocoder, cl
         const formatString = getFormatString(context, format, language);
         memo[`text${suffix}`] = text.text;
         if (text.language) memo[`language${suffix}`] = text.language.replace('_', '-');
-        memo[`place_name${suffix}`] = getPlaceName(context, formatString, language, languageMode);
-
+        if (feat.properties['carmen:intersection']) {
+            if (geocoder) memo[`place_name${suffix}`] = getPlaceName(context, formatString, language, languageMode, false, geocoder);
+        }
+        else memo[`place_name${suffix}`] = getPlaceName(context, formatString, language, languageMode);
         if (i === 0) {
             memo.text = memo[`text${suffix}`];
             if (text.language) memo.language = memo[`language${suffix}`];
@@ -163,7 +170,7 @@ function toFeature(context, format, languages, languageMode, debug, geocoder, cl
                         context[k].matching_text = matched.matching_text;
                         context[k].matching_language = matched.matching_language;
                         if (k === 0) feature.matching_text = matched.matching_text;
-                        feature.matching_place_name = getPlaceName(context, formatString, language, languageMode, !!matched);
+                        feature.matching_place_name = getPlaceName(context, formatString, language, languageMode, !!matched, geocoder);
                         break;
                     }
                 }
@@ -329,12 +336,22 @@ function getMatchingText(item, geocoder, requestedLanguage) {
     }
 }
 
-function getIntersectionName(properties) {
+/*
+* getIntersectionName - helper function to get the street to add to the cross street
+* @param {Object} geocoder Geocoder object
+* @param {object} properties A feature properties object
+*/
+function getIntersectionName(geocoder, properties) {
     if (properties['carmen:intersection'] && properties['carmen:query_text'] !== undefined) {
-        const featureSynonymMatch = properties['carmen:text'].split(',').find((f) => {
-            // find the feature or the synonym of the feature in carmen:text
-            // return the text to be added to the place_name
-            return f.toLowerCase().indexOf(properties['carmen:query_text'].toLowerCase()) >= 0;
+        let featureSynonymMatch;
+        const source = geocoder.byidx[properties['carmen:idx']];
+        properties['carmen:text'].split(',').forEach((sourceText) => {
+            const texts = termops.getIndexableText(source.simple_replacer, source.complex_query_replacer, geocoder.replacer,
+                { properties: { 'carmen:text': sourceText } }
+            );
+            if (texts[0].tokens.join(' ').indexOf(properties['carmen:query_text']) >= 0) {
+                featureSynonymMatch = sourceText;
+            }
         });
         return featureSynonymMatch.trim();
     }
--- a/lib/geocoder/ops.js
+++ b/lib/geocoder/ops.js
@@ -339,6 +339,8 @@ function getMatchingText(item, geocoder, requestedLanguage) {
 
 /*
 * getIntersectionName - helper function to get the street to add to the cross street
+* this is for Street B in Street A and Street B
+* the carmen:text field has a number of synonyms, we need to pick out the one similar to what the user typed
 * @param {Object} geocoder Geocoder object
 * @param {object} properties A feature properties object
 * @return String returns the correct carmen:text synonym based on the input query
@@ -347,12 +349,10 @@ function getIntersectionStreetName(geocoder, properties) {
     if (properties['carmen:intersection'] && properties['carmen:query_text'] !== undefined) {
         let featureSynonymMatch;
         const source = geocoder.byidx[properties['carmen:idx']];
-        properties['carmen:text'].split(',').forEach((sourceText) => {
-            const texts = termops.getIndexableText(source.simple_replacer, source.complex_query_replacer, geocoder.replacer,
-                { properties: { 'carmen:text': sourceText } }
-            );
-            if (texts[0].tokens.join(' ').indexOf(properties['carmen:query_text']) >= 0) {
-                featureSynonymMatch = sourceText;
+        properties['carmen:text'].split(',').forEach((synonymText) => {
+            const tokenizedIntersection = termops.tokenize(synonymText).map((token) => source.simple_replacer.tokens.get(token) || token);
+            if (tokenizedIntersection.join(' ').indexOf(properties['carmen:query_text']) >= 0) {
+                featureSynonymMatch = synonymText;
             }
         });
         return featureSynonymMatch.trim();
--- a/lib/geocoder/ops.js
+++ b/lib/geocoder/ops.js
@@ -82,7 +82,7 @@ function getPlaceName(context, formatString, language, languageMode, matched, ge
                     if (templateSubprop === '_name') {
                         if (f.properties['carmen:intersection']) {
                             const intersectionPrefix = feat.properties['carmen:intersection'] + ' and ';
-                            val = (intersectionPrefix + context.map((f) => { return (f.properties['carmen:intersection'] ? getIntersectionName(geocoder, f.properties) : undefined); }));
+                            val = intersectionPrefix + getIntersectionName(geocoder, f.properties);
                         }
                         // `name` is a hardcoded subproperty that maps to carmen:text
                         else if (matched && f.matching_text) val = f.matching_text;
@@ -340,6 +340,7 @@ function getMatchingText(item, geocoder, requestedLanguage) {
 * getIntersectionName - helper function to get the street to add to the cross street
 * @param {Object} geocoder Geocoder object
 * @param {object} properties A feature properties object
+* @return String returns the correct carmen:text synonym based on the input query
 */
 function getIntersectionName(geocoder, properties) {
     if (properties['carmen:intersection'] && properties['carmen:query_text'] !== undefined) {
--- a/lib/geocoder/ops.js
+++ b/lib/geocoder/ops.js
@@ -46,9 +46,12 @@ function getPlaceName(context, formatString, language, languageMode, matched) {
     let place_name;
     // Use geocoder_format to format output if applicable
     if (!formatString || formatString === true || formatString === 1) {
+        let prefix;
+        if (feat.properties['carmen:intersection']) prefix = feat.properties['carmen:intersection'] + ' and ';
+        else if (feat.properties['carmen:address']) prefix = feat.properties['carmen:address'] + ' ';
+        else prefix = '';
         place_name = (
-            (feat.properties['carmen:intersection'] ? feat.properties['carmen:intersection'] + ' and ' : '') +
-            (feat.properties['carmen:address'] ? feat.properties['carmen:address'] + ' ' : '') +
+            prefix +
             context
                 .filter((f) => { return matched || featureMatchesLanguage(f, { language: [language], languageMode: languageMode }); })
                 .map((f) => { return (matched ? f.matching_text : undefined) || closestLang.getText(language, f.properties).text; })
--- a/lib/geocoder/ops.js
+++ b/lib/geocoder/ops.js
@@ -46,7 +46,10 @@ function getPlaceName(context, formatString, language, languageMode, matched) {
     let place_name;
     // Use geocoder_format to format output if applicable
     if (!formatString || formatString === true || formatString === 1) {
-        place_name = ((feat.properties['carmen:address'] ? feat.properties['carmen:address'] + ' ' : '') +
+        place_name = (
+            (feat.properties['carmen:address'] ?
+            feat.intersections ?
+            feat.properties['carmen:address'] + ' and ' : feat.properties['carmen:address'] + ' ' : '') +
             context
                 .filter((f) => { return matched || featureMatchesLanguage(f, { language: [language], languageMode: languageMode }); })
                 .map((f) => { return (matched ? f.matching_text : undefined) || closestLang.getText(language, f.properties).text; })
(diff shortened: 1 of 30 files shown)


test/unit/text-processing/token.replaceToken.test.js
```diff
@@ -314,3 +314,12 @@ test('replace complex global tokens', (t) => {
     t.deepEqual(token.replaceToken(replacer, 'talstr 3-5'), { query: 'tal str 3-5', lastWord: false }, 'talstr 3-5 => tal str 3-5');
     t.end();
 });
+
+test('replace complex global tokens', (t) => {
+    const replacer = token.createGlobalReplacer({
+        '&': 'and'
```
aarthykc: Adding this global token replacement causes `AT&T` to change to `ATandT`

lib/text-processing/termops.js
```diff
@@ -573,6 +578,17 @@ function numTokenize(text, version) {
 
             numTokenized.push(replaced);
         }
+        if (replaced[i] === 'and') {
+            // f st +++ 9th st
+            replaced[i] = replaced[i].replace('and', '+++');
```
miccolis: I think we can just drop this line, immediately below we remove the element from the array.

lib/text-processing/termops.js
```diff
@@ -573,6 +578,17 @@ function numTokenize(text, version) {
 
             numTokenized.push(replaced);
         }
+        if (replaced[i] === 'and') {
```
miccolis: One thing we should try and avoid is hardcoding any assumptions about english always being the language of our data. Perhaps we could consider using something that is explicitly a placeholder here.

lib/geocoder/verifymatch.js
```diff
@@ -188,7 +192,7 @@ function verifyFeatures(query, geocoder, spatialmatches, loaded, options) {
             // and fell back to a street result. This penalty used to live in phrasematch
             // before being moved here to allow street level fallback mode
             if (feat.properties['carmen:address'] === false) cover.relev *= 0.99;
-
+            if (feat.intersections) cover.relev *= 0.99;
```
miccolis: Did you end up finding that we still need this after moving to a prefix string based approach? ...or is this just a remnant of the earlier attempt?

lib/geocoder/verifymatch.js
```diff
@@ -119,7 +119,10 @@ function verifyFeatures(query, geocoder, spatialmatches, loaded, options) {
 
         // Calculate to see if there is room for an address in the query based on bitmask
         // This recognizes 100 in the query list - `['1', 'fake', 'street', '100']` and returns the position of the addressnumber (3) and the addressnumber (100)
-        const address = termops.maskAddress(query, cover.text, cover.mask);
+        let address = termops.maskAddress(query, cover.text, cover.mask);
+        if (feats[0].intersections) address = {
+            addr: feats[0].properties['carmen:addressnumber'].slice(-1)[0]
+        };
```
miccolis: It's not clear to me the purpose  of this address object editing is. What am I missing?

lib/text-processing/termops.js
aarthykc: Not done this yet, @miccolis, will add this at the end!

ingalls:
I've included a bunch of new tests and some samples of the classes of intersection features

Ref: https://github.com/mapbox/carmen/pull/818

Basically it can have two variants
- Orphaned intersection - MultiPoint Feature => Geometry Collection
- Orphaned intersection Single Geometry Geometry Collection => No transformation needed
- Combined Intersection - GeometryCollection w/ Address => No transformation needed
- Combined Intersection - GeometryCollection w/ Network => No transformation needed
- Combined Intersection - GeometryCollection w/ Network & Address => No transformation needed

lib/geocoder/verifymatch.js
```diff
@@ -120,9 +120,25 @@ function verifyFeatures(query, geocoder, spatialmatches, loaded, options) {
         // Calculate to see if there is room for an address in the query based on bitmask
         // This recognizes 100 in the query list - `['1', 'fake', 'street', '100']` and returns the position of the addressnumber (3) and the addressnumber (100)
         const address = termops.maskAddress(query, cover.text, cover.mask);
+        let intersection;
+        // cover.text for intersections will look like - +intersection f street northwest , 9th street
+        // i.e they will be indexed with +intersection
+        if (cover.text.indexOf('+intersection') > -1) {
```
miccolis: Could this be more specific and test that `cover.text.indexOf('+intersection') === 0`? ...or are the circumstances where it wouldn't be the first bit of text?

lib/geocoder/verifymatch.js
```diff
@@ -120,9 +120,25 @@ function verifyFeatures(query, geocoder, spatialmatches, loaded, options) {
...
+            intersection = cover.text.split(',')[0].replace('+intersection', '').trim();
+        }
 
         if (source.geocoder_address) {
-            if (address && (feats[0].properties['carmen:addressnumber'] || feats[0].properties['carmen:rangetype'])) {
+            if (intersection) {
+                const exactMatchIntersection = feats[0].properties['carmen:intersections'].find((f) => { return intersection === f.toLowerCase(); });
+                // Get the index of the intersection in the ['carmen:intersections'] property
+                // and then look up the coordinates in geometry coordinates of the feature
+                const intersectionCoordinates = feats[0].properties['carmen:intersections'].findIndex((f) => { return intersection === f.toLowerCase(); });
```
miccolis: Could we avoid iterating over the `feats[0].properties['carmen:intersections']` array twice?

lib/geocoder/ops.js
```diff
@@ -46,7 +46,9 @@ function getPlaceName(context, formatString, language, languageMode, matched) {
     let place_name;
     // Use geocoder_format to format output if applicable
     if (!formatString || formatString === true || formatString === 1) {
-        place_name = ((feat.properties['carmen:address'] ? feat.properties['carmen:address'] + ' ' : '') +
+        place_name = (
+            (feat.properties['carmen:intersection'] ? feat.properties['carmen:intersection'] + ' and ' : '') +
+            (feat.properties['carmen:address'] ? feat.properties['carmen:address'] + ' ' : '') +
```
miccolis: This makes it seem like we'd support `carmen:intersection` and `carmen:address` on a single feature. Is that the case?

test/acceptance/geocode-unit.address-intersections.test.js
```diff
@@ -0,0 +1,272 @@
...
+        c.geocode('F Street Northwest and 9th Street Northwest', {}, (err, res) => {
+            t.ifError(err);
+            t.deepEquals(res.features[0].place_name, 'F Street Northwest and 9th Street Northwest', 'F Street Northwest and 9th Street Northwest');
+            t.deepEquals(res.features[0].geometry, { type: 'Point', coordinates: [0,2] }, 'Returns the correct geometry for F Street Northwest and 9th Street Northwest');
+            t.end();
+        });
+    });
+
+    tape('Searching for the intersection - F street northwest and 9th street northwest', (t) => {
+        c.geocode('F Street Northwest and', {}, (err, res) => {
```
miccolis: Just for completeness it would be nice to see a few more tests around the boundary, things like; `f st nw and`, `f st nw and 9`, `f st nw and 9th st`, `f street nw and 9th stre`

lib/geocoder/verifymatch.js
```diff
@@ -308,7 +323,13 @@ function verifyContexts(contexts, sets, indexes, options, geocoder) {
                 cover = verify[ctx.properties['carmen:tmpid']];
                 ctx.properties['carmen:matches_language'] = cover.matches_language;
                 ctx.properties['carmen:prefix'] = cover.prefix;
-                ctx.properties['carmen:query_text'] = cover.text;
+                // for intersections we want the feature 9th street where the cover.text = +intersection f street northwest , 9th street
+                if (ctx.properties['carmen:intersection']) {
+                    // only if the name of the street has been typed out
+                    if (cover.text.indexOf(',') > -1) {
+                        ctx.properties['carmen:query_text'] = cover.text.split(',')[1].trim();
```
miccolis: Something about this makes me a little nervous. Are there any cases where we can't assume that comma is going to be in the place we think it is? ...if it was ever missing we'd crash the process because element 1 would be undefined.

lib/text-processing/closest-lang.js
```diff
@@ -319,6 +319,15 @@ function getLanguageCode(str) {
  */
 function getText(language, properties) {
     if (!properties['carmen:text']) throw new Error('Feature has no carmen:text');
+    // query_text for intersections use case is undefined when the user has not typed out the street after the and
+    if (properties['carmen:intersection'] && properties['carmen:query_text'] !== undefined) {
```
miccolis: To play devil's advocate; is this really the right place for code that tests against the user's query? The rest of the method is narrowly focused on the language and this addition seems to change the scope of the method.

ingalls:
Working on supporting the new spec here: https://github.com/ingalls/pt2itp/pull/394

aarthykc:
Added a better explanation about the approach in the OP.

lib/geocoder/ops.js
```diff
@@ -76,8 +80,13 @@ function getPlaceName(context, formatString, language, languageMode, matched) {
                 // If we found a match between the template property and one within the returned context
                 if (templateProp === carmenProp) {
                     if (templateSubprop === '_name') {
+                        if (f.properties['carmen:intersection']) {
+                            const intersectionPrefix = feat.properties['carmen:intersection'] + ' and ';
+                            val = (intersectionPrefix + context.map((f) => { return (f.properties['carmen:intersection'] ? getIntersectionName(geocoder, f.properties) : undefined); }));
```
miccolis: If I follow this correctly it's rely on the Array to string case to turn this into a comma separated list... but won't any undefined elements end up as ",," in the resultant string?

lib/geocoder/ops.js
```diff
@@ -41,12 +41,16 @@ function getFormatString(context, format, language) {
  * @param {Boolean} matched If set, use matched text in place name
  * @return {String} formatted place_name string
  */
-function getPlaceName(context, formatString, language, languageMode, matched) {
+function getPlaceName(context, formatString, language, languageMode, matched, geocoder) {
```
miccolis: The js doc isn't updated, but if I follow the 'geocoder' param is the geocoder itself. I'd expect that to be the first param.

lib/geocoder/ops.js
```diff
@@ -321,3 +335,24 @@ function getMatchingText(item, geocoder, requestedLanguage) {
         return matched;
     }
 }
+
+/*
+* getIntersectionName - helper function to get the street to add to the cross street
+* @param {Object} geocoder Geocoder object
+* @param {object} properties A feature properties object
```
miccolis: Can we get a '@return' doc here. thx.

lib/geocoder/verifymatch.js
```diff
@@ -121,8 +121,41 @@ function verifyFeatures(query, geocoder, spatialmatches, loaded, options) {
...
+
         if (source.geocoder_address) {
-            if (address && (feats[0].properties['carmen:addressnumber'] || feats[0].properties['carmen:rangetype'])) {
+            if (intersection) {
+                let exactMatchIntersection, intersectionCoordinates, geometryPos;
+                // Get the index of the intersection in the ['carmen:intersections'] property
+                // and then look up the coordinates in geometry coordinates of the feature
+                const intersectionsArray = feats[0].properties['carmen:intersections'];
+                for (let i = 0; i < intersectionsArray.length; i++) {
+                    if (!intersectionsArray[i]) continue;
```
miccolis: When would we have an empty element in this array?

lib/geocoder/verifymatch.js
```diff
@@ -121,8 +121,41 @@ function verifyFeatures(query, geocoder, spatialmatches, loaded, options) {
...
+                    // geometryPos is position of the intersection multipoint array in the GeometryCollection
+                    geometryPos = intersectionsArray.indexOf(intersectionsArray[i]);
+                    for (let j = 0; j < intersectionsArray[i].length; j++) {
+                        const indexedIntersectionText = termops.getIndexableText(source.simple_replacer, source.complex_query_replacer, [],
```
miccolis: nit pick, but can we change this variable name, it makes me think we're about do index something.

Also, I'm not sure of a better way, but getIndexableText is a pretty heavy function. Are they any alternatives?

lib/text-processing/termops.js
```diff
@@ -369,13 +390,20 @@ function getMinimalIndexableText(replacer, globalTokens, doc) {
...
         if (!tokens.length) continue;
 
-        if (housenumRange) {
+        if (intersections) {
+            let l = intersections.length;
+            while (l--) {
+                const withIntersections = [intersections[l]].concat(tokens);
+                add(withIntersections);
+            }
+        } if (housenumRange) {
```
miccolis: This `if` statement should be on a new line.

lib/text-processing/termops.js
```diff
@@ -536,16 +571,28 @@ function getWeights(tokens, freq) {
...
                 weights[i] = 0.2;
-            } else {
+            }
+            else {
                 weights[i] = Math.max(weights[i] / weightsum) * 0.8;
             }
         }
-    } else {
+    }
+    // if the token has a +intersection give a constant weight of 0.4
```
miccolis: I'm curious, how did you pick 0.4?

lib/geocoder/verifymatch.js
aarthykc: I changed this to `tokenizedIntersectionText`, does that work better? I've added comments about why we need to do this. 

I agree, I see that portion of the getIndexableText where we tokenize a string based on the simple token replacement being helpful in other places as well. Are you saying we should move that to a separate function?

lib/text-processing/termops.js
aarthykc: Done!

lib/geocoder/verifymatch.js
aarthykc: I've mentioned this in the data spec, checkout -- https://github.com/mapbox/carmen/pull/808/files#diff-1dacf7eebaabb0846d2e11bcdb3a6be2R40

lib/geocoder/ops.js
aarthykc: Yess, totally missed this which ended up ruining the formatting, I've fixed this in the next commit.

aarthykc: `$(yarn bin)/carmen --config dev/mapbox.places.js --query "O Street Northwest and Naylor Court Northwest" --types address`

tokens
------
o, street, northwest, and, naylor, court, northwest

Features
--------
- 1.00 O Street Northwest and Naylor Court Northwest, Washington, District of Columbia 20001, United States (address.4561921048832130)

lib/geocoder/ops.js
aarthykc: done!

lib/text-processing/termops.js
aarthykc: This was a little bit of a trial, I tried with 0.2 but it was not high enough to bump intersections and so I ended up going with 0.4

lib/text-processing/termops.js
```diff
@@ -620,3 +686,32 @@ function decode3BitLogScale(num, max) {
...
+            ranges.push(numToken);
+        }
+    }
+    if (doc.properties['carmen:intersections']) {
+        // carmen:intersections will look something like [null, [intersections array], null]
+        const keys = doc.properties['carmen:intersections'].filter(Boolean);
+        for (let i = 0; i < keys.length; i++) {
+            for (let j = 0; j < keys[i].length; j++) {
+                // indexes F Street Northwest as 'intersectionsearch <key>'
+                add('intersectionsearch ' + keys[i][j]);
```
aarthykc: Since, the + in +intersection gets stripped anyway, chose to go with `intersectionsearch` and covert tokens to '+intersection' <streetname> ',' finally in tokenize()

test/acceptance/geocode-unit.address-intersections.test.js
```diff
@@ -0,0 +1,400 @@
...
+
+    tape('Searching for the intersection - F st nw and 9th st', (t) => {
+        c.geocode('F st nw and 9th st', {}, (err, res) => {
+            t.deepEquals(res.features[0].place_name, 'F Street Northwest and 9th Street Northwest', 'F st nw and 9th st');
+            t.deepEquals(res.features[0].geometry, { type: 'Point', coordinates: [0,2] }, 'Returns the correct geometry for F Street Northwest and 9th Street Northwest');
+            t.end();
+        });
+    });
+
+    tape('Searching for the intersection - F st nw and 9th (should favour returning the street over the intersection)', (t) => {
```
aarthykc: This is because of the penalty I added in for intersections when we don't have enough information about the cross street, I feel fine returning an intersection only when we have more information instead of guessing.

lib/text-processing/termops.js
aarthykc: done

lib/geocoder/ops.js
miccolis: Ping on this again. I really think this function signature needs to be changed, or maybe it's too overloaded and needs to be refactored? I'm not sure.

lib/geocoder/ops.js
```diff
@@ -321,3 +336,25 @@ function getMatchingText(item, geocoder, requestedLanguage) {
...
+* @return String returns the correct carmen:text synonym based on the input query
+*/
+function getIntersectionStreetName(geocoder, properties) {
```
miccolis: A few things about this method kinda bother me. At a high level I think I just can't see all the implications of the aggressive string manipulations here. Should we have unit tests of this method?

lib/geocoder/verifymatch.js
```diff
@@ -121,8 +121,53 @@ function verifyFeatures(query, geocoder, spatialmatches, loaded, options) {
...
+                    if (!intersectionsArray[i]) continue;
+                    // geometryPos is position of the intersection multipoint array in the GeometryCollection
+                    geometryPos = intersectionsArray.indexOf(intersectionsArray[i]);
+                    for (let j = 0; j < intersectionsArray[i].length; j++) {
+                        // while indexing we tokenize the intersection array
+                        // to find the exact match intersection we have to tokenize it first before we attempt to compare them
+                        const tokenizedIntersectionText = termops.getIndexableText(source.simple_replacer, source.complex_query_replacer, [],
+                            { properties: { 'carmen:text': intersectionsArray[i][j] } }
+                        );
+                        if (intersection === tokenizedIntersectionText[0].tokens.join(' ')) {
```
miccolis: I'm not sure I follow the intention here. Won't getIndexableText generate a lot of options? Why would we do that and only look at the first?

lib/text-processing/termops.js
```diff
@@ -126,6 +128,13 @@ function tokenize(query, lonlat) {
         }
     }
 
+    // we check if the tokens have 'intersectionsearch' if so we know that it is intersection data
+    // add +intersection to the front of the array
+    // add a comma to the end of the array
+    if (tokens.includes('intersectionsearch')) {
```
miccolis: What I liked about `+intersection` was that is used a reserved character that a user couldn't enter. Any way we can  prevent the "+" from getting stripped away?

More importantly, I don't follow why we'd want to have re-ordering happen in tokenize. Seems like that changes the nature of the method. Is there a reason we can't handle it all elsewhere?

lib/text-processing/termops.js
miccolis: Yea, this make me sad. Do you know where it's getting stripped?

lib/text-processing/termops.js
```diff
@@ -560,7 +618,7 @@ function isAddressNumber(tokens) {
 }
 
 module.exports.numTokenize = numTokenize;
-function numTokenize(text, version) {
+function numTokenize(text, version, intersectionToken) {
```
miccolis: Seems like we should change the name of the method if we're going to change it's scope here - AddressTokenize? ...even then this isn't about tokenization. variants?

test/unit/util/feature.addrTransform.test.js
```diff
@@ -0,0 +1,388 @@
...
+});
+
+tape('Combined Features', (t) => {
+    t.deepEquals(transform({
+        type: 'Feature',
+        properties: {
+            'carmen:text': 'Main Street Northwest',
+            'carmen:addressnumber': [null, [1, 2, 3]],
+            'carmen:rangetype': 'tiger',
+            'carmen:parityl': [[], null],
```
miccolis: I'm curious why are these sometimes null and sometimes the empty array? Is there a meaningful difference? Should we prefer one form? ...personally I like empty array's better

lib/text-processing/termops.js
aarthykc: This is getting stripped in tokenize, which is why i tag the `+intersection` and add the comma there given that both get stripped earlier on in the tokenize function

lib/geocoder/ops.js
aarthykc: sure, I can add unit tests

lib/text-processing/termops.js
```diff
@@ -573,6 +633,14 @@ function numTokenize(text, version) {
 
             numTokenized.push(replaced);
         }
+        if (!!intersectionToken && replaced[i] === intersectionToken) {
```
miccolis: @aarthykc as per chat - what I'm wondering is if we should make sure that the token we're examining isn't the last one. A possible check could be `... && i  < text.length - 1`. that way we'd avoid looking for cross streets when we don't have any info about one yet.

aarthykc:
Joint forces with @miccolis post tokenization refactor to create a new PR - #838

PR #810, opened 2019-01-16
Cluster Properties Spec

ingalls:
### Context

We need the ability for properties to be assigned to individual addresses in an address cluster. At this time you can assign properties just like you would any GeoJSON feature, however they will apply to all
point geometries in the cluster

This PR introduces an override system designed to allow individual addresses on the cluster to assign differing properties.

A sample of this override behavior can be viewed below where the feature is given the default property of `accuracy: point` however element 1 (`[1,1]`) of the points array overrides this property with a value of `accuracy: entrance`

```JSON
{
  "type": "Feature",
  "properties": {
    "accuracy": "point",
    "carmen:addressprops": {
     "accuracy": {
        "1": "entrance"
      }
    }
  },
  "geometry": {
    "type": "MultiPoint",
    "coordinates": [[0,0], [1,1],[2,2]]
  }
}
```

cc @mapbox/search


apendleton:
So I think it's a little funky that address numbers are associated with points by position (0th goes with 0th, 1st goes with 1st, etc.) whereas these properties are associated by casting the position to a string and storing it as an object key. If the idea is to use this feature to store accuracy or admin data or whatever, in many (most?) cases we'll have such metadata for every house number. Seems like this encoding could end up super bulky (and actually, even if we did it positionally, we'd still have to represent the *name* of the property -- "accuracy" for example -- once per house number). Makes me a bit concerned about size.

ingalls:
Per chat, agree that the current spec is pretty funky and we should revisit in the future for a a wider look at updating.

For now this seems to be the most efficient way forward short of storing all the properties individually in a parallel array (like `carmen:addressnumber` is now)

ingalls:
## TODO

Stopping for tonight, this is technically feature complete but I would like to clean up the interface to reduce some code existing complexity by outputting an array of Features from `addresscluster#forward` instead of the current strange half features. This will match the behavior of `addresscluter#reverse`. It will also reduce the number of `Object.keys` calls required.

ingalls:
The beginnings of a working implementation of this spec can be found here: https://github.com/ingalls/pt2itp/pull/387

lib/geocoder/addresscluster.js
```diff
@@ -4,21 +4,22 @@ module.exports.forwardPrefix = forwardPrefix;
...
+ *
+ * @param {Object} feat GeoJSON Feature to derive individual address from
+ * @param {String|number} address User's queried address number
+ * @param {number} num Max number of address features to potentially return (default 10)
+ *
+ * @return {Array|false} Array of potential address feats or false
  */
-function forward(feature, address) {
-    if (!feature.geometry || feature.geometry.type !== 'GeometryCollection') return false;
+function forward(feat, address, num = 10) {
```
apendleton: To make sure I'm following: the num=10 isn't a behavior change, right, because before we were truncating to 10 in verifymatch?

lib/geocoder/addresscluster.js
```diff
@@ -31,24 +32,40 @@ function forward(feature, address) {
...
+                return feat_clone;
             });
         }
     }
 
     return false;
 }
 
-/**
+/*
```
apendleton: Why? Is this not valid jsdoc?

lib/geocoder/addresscluster.js
```diff
@@ -94,6 +111,14 @@ function forwardPrefix(feature, address) {
     return [];
```
apendleton: Any new functionality that's added to forward should be added to forwardPrefix as well.

lib/geocoder/verifymatch.js
```diff
@@ -168,15 +168,11 @@ function verifyFeatures(query, geocoder, spatialmatches, loaded, options) {
...
-                            addressPoints = [{
-                                type:'Point',
-                                coordinates: [
-                                    Math.round(firstMiddleLast[0].geometry.coordinates[0] * 1e6) / 1e6,
-                                    Math.round(firstMiddleLast[0].geometry.coordinates[1] * 1e6) / 1e6
-                                ]
-                            }];
-                            feats[0].properties['carmen:address'] = firstMiddleLast[0].number;
-                            feats[0].properties['carmen:distance'] = firstMiddleLast[0].distance;
+                            addressPoints = addressCluster.forward(feats[0], firstMiddleLast[0].number).map((pt) => {
```
apendleton: This doesn't make sense to me. Why would we search through the feature once to find things that match the prefix, then search through it again to get that same number? Why not just return what we need from forwardPrefix in the first place?

apendleton: To expand on that a bit: in the event that the same number is duplicated in the address cluster more than once, you're not guaranteed to get back the same result this time as last time. We picked a specific instance of that number because it's the closest one to the proximity point, but this might give us back a different one, in which case we would no longer be returning the closest one, and the distance might be incorrect.

ingalls: @apendleton features with a high degree of duplicative data can cause extremely high memory usage due to the cloning of the feature JSON. I moved the `num=10` into the code to combat the unnecessary use of feature clones.

This was a cheap way to do the same but I agree with your analysis. Will rework and likely move a bunch of the selection logic from verify match into it's own function

PR #813, opened 2019-01-23
Numerical autocomplete

apendleton:
### Context
This is PR adds a new mode of attempting to prefix-match in forward searches, specifically geared towards the circumstance in which the user is still in the process of typing in the numerical portion of an address (and focusing at least for now on US-ordered addresses, where the number comes first). At present we assume when we process numbers that they're complete, and don't have a mechanism for searching for, for example, "any address near me that starts with 45" (potentially including "4512"). We'll only be enabling this behavior where proximity is enabled, when matching against address indexes that contain the proximity point.

### Summary of Changes
- [x] added ew code in phrasematch to generate additional waffle combinations for prefix-matching of addresses, and flagging them as partial numbers, in the event that we're querying an address number and it contains the prox point
    - [x] conversely, we also now explicitly *disable* all-number matches for address indexes when this isn't the case, at the phrasematch stage; we didn't do this before, we'd just generate phrasematches that had their autocomplete flags turned off so they wouldn't ever actually match anything, but still generated stackable stacks
- [x] added a tiny bit of new code in spatialmatch to pass the partial-number-ness through to verifymatch, and also to generate a bounding box for partial-number matches to control the scale of possibilities we consider (this adds a new dependency on cheap-ruler, which can calculate such bboxes very quickly)
- [x] added functionality in verifymatch and related utilities (address-cluster, especially) to select from within a given address cluster which address should be surfaced
    - [x] the strategy here is, once we have an address cluster, to surface at most one result from that cluster; we ignore interpolation altogether and consider only exact numbers, and select which number to display by filtering the address number list for numbers that start with the requested prefix, sorting that filtered list, selecting the first, middle, and last numbers, and choosing the one that's spatially closest to the proximity point (note that this avoids calculating the distance to every address, but does mean that if the proximity point is literally on top of one of the addresses in the cluster, we may still not surface that address if it's not the first/middle/last of the matching cluster)
- [x] added support both at query and index time for index bounds that cross the antimeridian, to make it possible for affected indexes to have smaller bounds such that they don't accidentally suck in lots of actually-far-away proximity points and do expensive address-number scans that won't actually yield anything; this involved writing new antimeridian-aware versions of several bounding box utilities, and lays the groundwork for supporting bbox filters from users that cross the antimeridian as well (though that's out of scope for this PR)
- [x] moved to a new carmen-cache version (mapbox/carmen-cache#142) that has a bunch of performance improvements that make the above not take a million years to complete
- [x] added a major scoredist boost for address matches that are a result of this procedure; the typically-low scores of addresses mean they aren't surfaceable otherwise
- [x] added tests

### Next Steps
- [ ] release a new carmen-cache and switch to it
- [ ] confirm that quality looks good downstream

cc @mapbox/search


lib/geocoder/addresscluster.js
```diff
@@ -39,6 +40,43 @@ function forward(feature, address) {
...
+            const element = typeof e === 'string' ? e : ('' + e);
+            return element.startsWith(a1) ? a.concat(i) : a;
+        }, []);
+
+        // Check if cluster is pt geom
+        if (a_index.length && feature.geometry.geometries[c_it].type === 'MultiPoint') {
+            return a_index.map((idx) => {
+                return {
+                    number: cluster[c_it][idx],
+                    numberAsInt: parseInt(cluster[c_it][idx]),
```
miccolis: Should probably add a radix.

lib/geocoder/phrasematch.js
```diff
@@ -98,11 +103,32 @@ module.exports = function phrasematch(source, query, options, callback) {
...
+        } else if (tokenized.length === 1 && termops.parseSemiNumber(tokenized[0])) {
+            if (proxMatch) {
+                // if the query is just one word and it's a number, and we're
+                // at a prox point inside this index, do some special stuff to
+                // consider what different things that number might be short for
+                const numTokenized = termops.numTokenizePrefix(tokenized, source.version);
+                for (let i = 0; i < numTokenized.length; i++) {
+                    allSubqueries = allSubqueries.concat(termops.permutations(numTokenized[i]));
+                }
+                partialNumber = true;
```
miccolis: I'm trying to understand what would happen in the case where the query is a single token w/ number and alpha chars. Like a poi type search for "123company". If I'm following things properly we'd strip that down do "123" in `numTokenizePrefix` and search for just that. Is that right? ...or am I missing something.

lib/geocoder/addresscluster.js
apendleton: Added

lib/text-processing/termops.js
```diff
@@ -577,6 +577,34 @@ function numTokenize(text, version) {
     return numTokenized;
 }
 
+module.exports.numTokenizePrefix = numTokenizePrefix;
+function numTokenizePrefix(text, version) {
```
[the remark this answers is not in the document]
apendleton: Added

lib/geocoder/phrasematch.js
apendleton: This block now only applies strictly to tokens comprised entirely of digits.

lib/geocoder/addresscluster.js
```diff
@@ -39,6 +40,43 @@ function forward(feature, address) {
     return false;
 }
 
+function forwardPrefix(feature, address) {
```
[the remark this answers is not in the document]
apendleton: Added

lib/text-processing/termops.js
```diff
@@ -389,6 +389,11 @@ function getMinimalIndexableText(replacer, globalTokens, doc) {
     return indexableText;
 }
 
+/**
+ * Detect and extract a number from a mixed string
+ * @param {string}
+ * @return {null|string}
```
miccolis: My bad here, the return is `{null|number}`

Issue #515, opened 2016-10-19
Runtime token replacement last stretch...


yhahn:
Debugging some IRL scenarios with https://github.com/mapbox/carmen/pull/476 the last stragglers seem to be cases where we can deconstruct what's happening like this:

tokens:

```
north => n
```

Becomes

```
(\\W|^)north(\\W|$) => n
(\\W|^)n(\\W|$) => north
```

_(regex word boundaries)_

These falsely split on high range unicode characters creating totally bogus permutations for DAWG lookup like:

```
Original query:
  General Pirán Argentina

One of the token-replaced permutations:
  General Piránorth Argentina
```

:grimacing: sometimes when this happens the resulting permutation actually matches something :confounded: and honestly we should have never even considered the permutation legit in the first place.
### Possible solutions
- Change how the token replacement function works (https://github.com/mapbox/carmen/blob/master/lib/util/token.js#L48-L58). For example we could
  - use the tokenizer which _does_ have a more sophisticated concept of word boundary and then replace individual tokens post-tokenization. We would remove the weird `(\\W)` word boundary portions of our regex.
- Use a more sophisticated word boundary concept around our computed token regexes.
- Use xregexp or other library that is more unicode aware.
- Other?

_Any/all of these approaches could have a serious impact on performance. Token replacement can get very expensive so we should profile._
### Next actions
- [x] Branch off of https://github.com/mapbox/carmen/pull/476
- [x] Add some unit tests for this word boundary vs unicode characters fail case
- [x] Implement token replacement in a way where these false permutations are not considered

ansis:
I started looking into this and it seemed like the bidirectional token replacement in #476 only works for basic string tokens. Regex tokens don't work because they can't just be swapped. Source tokens that use regex are pretty rare though. From what I could find, the only mapbox sources that use token replacement with regex are the `kr` ones:
```json
"geocoder_tokens": {
            "(?<root>\\w+)-dong": "${root}dong",
            "(?<root>\\w+)-si": "${root}si",
            "(?<root>\\w+)-gu": "${root}gu",
            "(?<root>\\w+)-gun": "${root}gun"
        }
```
We either need to:
- provide the regex for the reverse replacement
- approximate the replacement without regex  `"-dong": "dong"` and completely drop regex support for source token replacement

Am I missing some more uses of regex for source token replacement?

----

[...]

PR #816, opened 2019-01-29
Remove xregexp module

miccolis:
PR removes the `xregexp` module and named regex support. We haven't used it for anything in quite a while.

Refs #515

cc @mapbox/search


miccolis:
Spent some time digging into the node 10 test failure only to discover that named capture groups are supported on Node 10. Not too much detail on the web, but http://2ality.com/2017/05/regexp-named-capture-groups.html is a great read on the details. Should we need it, we still have this tool at our disposal.

Issue #815, opened 2019-01-29
Sort out word boundary & word separator definitions


miccolis:
The WORD_BOUNDARY constant and the termops token splitter are very similar, but not exactly the same. This isn't inherently a problem, but they usage also seems pretty unaware of each other may cause some issues and certainly causes confusion.

The word boundary constant looks like:

https://github.com/mapbox/carmen/blob/69639e02369d272f62297ac5a0b46b5f97deae3c/lib/constants.js#L26

Termops token splitter is;

https://github.com/mapbox/carmen/blob/69639e02369d272f62297ac5a0b46b5f97deae3c/lib/text-processing/termops.js#L96

It's probably not be possible unify these, they have different jobs to do. For example, we don't treat `-` as a word boundary for tokenization because we support some address formats where the house number has hyphens in it. We should make those reasons explicit and the code comments should indicate why one is different than the other. The current link to stackoverflow does not inspire confidence.

### Next Steps

- Code comments, even if nothing changes we need to document the current behavior in code
- Reconsider where the word_boundary is defined. As a constant it appears to be some thing spans contexts, this may not be true. It seems to have a specific job.
- Review and develop an opinion and document about the job of the word splitting regex and replacement boundary regexes and the character they test for. Should anything change? What decisions are implied here?

@mapbox/search

miccolis:
Gonna close this;

* After #871 the regular expressions we use are much better documented. Like https://github.com/mapbox/carmen/blob/5545c90f500e2ee396eecf78b56c8c4a61ed365f/lib/text-processing/termops.js#L75-L98
* Also in #871 I've moved the word_boundary to a local context, as it is only used in one spot.
* While, not merged yea, I'm moving on changing how tokinzation work in https://github.com/mapbox/carmen/pull/826 and removing the word boundary duplication in https://github.com/mapbox/carmen/pull/831

PR #817, opened 2019-01-30
Define boundaries

miccolis:
Goal of this PR is to more explicitly capture when we split user input into tokens and why. We do this in a few spots and they've been pretty dense regular expressions and code point ranges. To improve that this PR does the following things;

1. Calls out the regexp that is used to split a query into token as the `WORD_SEPARATOR` in `termops.js` and documents all the Unicode ranges and characters that it contains.
2. Moves the `WORD_BOUNDARY` definition out of the constants file and into `token.js` which is the only place it is used. It also documents all the  Unicode ranges and characters that it contains.
3. Adds a code comment to the CJK specific token splitting and uses hex ranges to make the intent more clear. 
4. ...and I couldn't help it, I consolidated the two post-splitting token iterations. Tests pass and perf looks better, but it could use close review.
5. Can't stop, Also removes the treatment of simple replacements via regexp. It was only used for frequency calculation and category normalization and was overly complex for that use.

Perf stats from my laptop, posting b/c CJK seems much faster on this branch, not quite sure what to make of that.

On Master:
```
# tokenize
tokenize x 360,489 ops/sec ±0.57% (97 runs sampled)
tokenize - cjk x 372,653 ops/sec ±0.73% (92 runs sampled)
```

This branch
```
# tokenize
tokenize x 389,059 ops/sec ±0.72% (95 runs sampled)
tokenize - cjk x 428,671 ops/sec ±0.83% (93 runs sampled)
```

Refs #815 

cc @mapbox/search


index.js
```diff
@@ -179,18 +179,17 @@ function Geocoder(indexes, options) {
...
                 for (let category of info.geocoder_categories) {
-                    category = termops.tokenize(category, true);
+                    category = termops.tokenize(category);
 
-                    source.categories.add(category.join(' '), true);
+                    source.categories.add(category.join(' '));
 
                     category = category.map((cat) => {
-                        return token.replaceToken(catReplacer, cat).query.toLowerCase();
+                        return token.replaceToken(source.complex_query_replacer, source.simple_replacer.tokens.get(cat) || cat).query;
```
apendleton: We discussed this earlier and I probably don't care, but still curious if it's possible that a global/cardinality-changing token replacement could affect a category name, and if so, if applying replacements to category names a word at a time would keep them from working.

lib/text-processing/termops.js
```diff
@@ -88,44 +111,39 @@ function tokenize(query, lonlat) {
...
         .toLowerCase()
-        .replace(/[\^]+/g, '')
-        // collapse apostraphes, periods
-        .replace(/[\u2018\u2019\u02BC\u02BB\uFF07'\.]/g, '')
-        // all other ascii and unicode punctuation except '-' per
-        // http://stackoverflow.com/questions/4328500 split terms
-        .replace(/[\u2000-\u206F\u2E00-\u2E7F\\'!"#\$%&\(\)\*\+,\.\/:;<=>\?@\[\]\^_`\{\|\}~]/gi, ' ')
-        .split(/[\s+]+/gi);
+        // collapse apostraphes, periods, caret
+        .replace(/[\u2018\u2019\u02BC\u02BB\uFF07'\.\^]/g, '')
```
apendleton: What changed here? The caret, right?

lib/text-processing/termops.js
```diff
@@ -88,44 +111,39 @@ function tokenize(query, lonlat) {
...
+        // split based on other ascii and unicode punctuation except '-' which
+        // we handle separately below.
+        .split(new RegExp(`[${WORD_SEPARATOR}+]`, 'g'));
```
apendleton: Probably doesn't matter, but out of curiosity, think there's any benefit to just creating this regex the ones and reusing it rather than making a new one every time?

lib/text-processing/token.js
```diff
@@ -1,6 +1,33 @@
...
+    // :;<=>?@
+    '\u003A-\u0040',
+    // [\]^_`
+    '\u005B-\u0060',
+    // {|}~
+    '\u007B-\u007E',
+
+    ']'
+].join('');
+module.exports.WORD_BOUNDARY = WORD_BOUNDARY;
```
apendleton: For some reason I thought we had concluded earlier that the distinction between the two kinds of separators wasn't important anymore because it turns out that the simple token replacements now use the tokenized string, so the separator used for token replacement and the separator used for tokenization are now effectively the same. What are the circumstances under which we still need to to distinguish between one kind and the other?

lib/text-processing/termops.js
miccolis: Yea, like 3 lines up was `.replace(/[\^]+/g, '')`

index.js
miccolis: The intent here was to preserve the existing behavior, which looks to have been; tokenize, do simple replacements, then complex ones. So yes, any categories that could be affected by complex multi-token replacements won't be here. However I did a quick scan of some production data and can't find a situation where it would happen. Categories like 'apartment building' are more common and would get the proper treatment.

lib/text-processing/token.js
miccolis: I was getting ahead of myself, we still need both. The only difference is `-` which we have special behavior for in tokenization. I think we can unwind that special behavior and only do this sort of splitting once. But it'll require a firm contract between tokenization and replacements and we're just not there yet.

Issue #821, opened 2019-02-05
Preference non-interpolated address results


lizziegooding:
### Context
Right now, interpolated and non-interpolated results with the same relevance are returned in whatever order they came back from spatialmatch in. This results in a negative user experience where the desired feature is not only duplicated, but often the poorer quality interpolated result is returned first.

![screen_shot_2019-02-05_at_3_17_08_pm](https://user-images.githubusercontent.com/18095166/52301815-5700a900-2959-11e9-99ff-80686e0c20ce.png)
_Example querying `1013 N Bonnie Brae St, Los Angeles, California 90026, United States`_

### Next Steps
An easy fix would be to add a sort to https://github.com/mapbox/carmen/blob/c328db6117fc09067cb53467ab43c4040797b2b7/lib/geocoder/verifymatch.js#L583-L592
that would sort non-interpolated results before interpolated results. This would improve the order of results but do nothing to deduplicate results. 

To not return multiple results, we could either add some kind of deduping logic at query time if the address features seem similar but one is interpolated and one is not or we could dive deeper into how we're constructing address clusters. My preference would be the latter.

@mapbox/search

lizziegooding:
Done in https://github.com/mapbox/carmen/pull/822

PR #822, opened 2019-02-07
Preference non-interpolated address results

lizziegooding:
### Context
Ref https://github.com/mapbox/carmen/issues/821

### Summary of Changes
- [x] add a interpolation sort criteria in verifyMatch for address results

### Next Steps
This is a simple change but revealed some other behavior around how we handle interpolation and deduplication I'd like to dig into.

cc @mapbox/search

### DIFF
--- a/lib/geocoder/verifymatch.js
+++ b/lib/geocoder/verifymatch.js
@@ -654,6 +654,10 @@ function sortContext(a, b) {
         // for address results, prefer those from point clusters
         if (a[0].properties['carmen:addressnumber'] && !b[0].properties['carmen:addressnumber']) return -1;
         if (b[0].properties['carmen:addressnumber'] && !a[0].properties['carmen:addressnumber']) return 1;
+
+        // prefer non-interpolated address results
+        if ((a[0].geometry && !a[0].geometry.interpolated) && (b[0].geometry && b[0].geometry.interpolated)) return -1;
+        if ((b[0].geometry && !b[0].geometry.interpolated) && (a[0].geometry && a[0].geometry.interpolated)) return 1;
     }
 
     // omitted difference
--- a/test/acceptance/geocode-unit.address-sort-interpolated.test.js
+++ b/test/acceptance/geocode-unit.address-sort-interpolated.test.js
@@ -1,0 +1,118 @@
+'use strict';
+const tape = require('tape');
+const Carmen = require('../..');
+const context = require('../../lib/geocoder/context');
+const mem = require('../../lib/sources/api-mem');
+const queue = require('d3-queue').queue;
+const { queueFeature, buildQueued } = require('../../lib/indexer/addfeature');
+
+// Test that non-interpolated results are returned before interpolated results
+// with the same relevance. Note that features must have different
+// `place_name`s to avoid being deduped in dedupe.js.
+(() => {
+    const conf = {
+        address: new mem({ maxzoom: 6,  geocoder_address:1, geocoder_tokens: { 'Street': 'St' } }, () => {}),
+    };
+    const c = new Carmen(conf);
+    tape('index address cluster to interpolate', (t) => {
+        const address = {
+            id: 1,
+            properties: {
+                'carmen:text': 'Main St',
+                'carmen:center': [-97.2, 37.3],
+                'carmen:rangetype': 'tiger',
+                'carmen:lfromhn': [['100'], []],
+                'carmen:ltohn': [['200'], []],
+                'carmen:rfromhn': [['101'], []],
+                'carmen:rtohn': [['199'], []],
+                'carmen:parityl': [['E'], []],
+                'carmen:parityr': [['O'], []],
+                'carmen:addressnumber': [null, ['100', '200']]
+            },
+            geometry: {
+                type: 'GeometryCollection',
+                geometries: [
+                    {
+                        type: 'MultiLineString',
+                        coordinates: [
+                            [
+                                [-97.2, 37.2],
+                                [-97.2, 37.4]
+                            ]
+                        ]
+                    },
+                    {
+                        type: 'MultiPoint',
+                        coordinates: [[-97.2, 37.2],[-97.2, 37.4]]
+                    }
+                ]
+            }
+        };
+        queueFeature(conf.address, address, t.end);
+    });
+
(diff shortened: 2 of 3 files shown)


test/acceptance/geocode-unit.address-sort-interpolated.test.js
```diff
@@ -0,0 +1,118 @@
+'use strict';
+const tape = require('tape');
+const Carmen = require('../..');
+const context = require('../../lib/geocoder/context');
+const mem = require('../../lib/sources/api-mem');
+const queue = require('d3-queue').queue;
+const { queueFeature, buildQueued } = require('../../lib/indexer/addfeature');
+const addrTransform = require('../../lib/util/feature').addrTransform;
```
lizziegooding: We are not currently validating address features we use in our acceptance tests; we should add this transform to `queueFeature`/`buildQueued`

test/acceptance/geocode-unit.address-sort-interpolated.test.js
```diff
@@ -0,0 +1,118 @@
...
+            });
+        });
+        q.awaitAll(t.end);
+    });
+
+
+    tape('Return non-interpolated address before interpolated address', (t) => {
+        c.geocode('150 Main Street', { limit_verify: 2 }, (err, res) => {
+            t.ifError(err);
+            t.deepEquals(res.features.length, 2);
```
lizziegooding: This fails because we only return a single, non-interpolated result in this case. The interpolated result is deduped in https://github.com/mapbox/carmen/blob/master/lib/geocoder/dedupe.js because both features have the exact same name and the interpolated features also has `geometry.omitted` set to `true` (more on that later). If you change one of the features from `Main Street` to `Main St`, they fail the `placename` check and tests pass. I've left this failing on purpose for the moment because I think we should revisit this behavior.

test/unit/geocoder/address.test.js
```diff
@@ -667,6 +696,33 @@ test('parity: odd + even', (t) => {
     t.end();
 });
 
+test('bug(?) address.forward', (t) => {
+    t.deepEqual({
+        type: 'Point',
+        coordinates: [-97.2, 37.4],
+        interpolated: true
```
lizziegooding: `address.forward` sets this feature's `geometry.omitted` to `true` when it is possible to interpolate an address from this feature that is not a line endpoint -- this results in the deduping that happens later.

test/unit/geocoder/verifymatch.test.js
```diff
@@ -20,122 +20,69 @@ tape('verifymatch.sortFeature', (t) => {
     t.end();
 });
 
-tape('verifymatch.sortContext (no distance)', (t) => {
+tape('verifymatch.sortContext', (t) => {
```
lizziegooding: These tests relied on a `_distance` property which we no longer use in `sortContext`. I removed all but one of the tests and added more test features to demonstrate every property we test against in the sort.

test/acceptance/geocode-unit.address-sort-interpolated.test.js
lizziegooding: Changed to `Main St` vs `Main Street` to avoid deduplication and demonstrate we can return both results and we will preference the non-interpolated one.

lizziegooding: See https://github.com/mapbox/carmen/issues/823 for more thoughts on current deduping functionality

test/acceptance/geocode-unit.address-sort-interpolated.test.js
lizziegooding: See https://github.com/mapbox/carmen/issues/824

test/unit/geocoder/address.test.js
```diff
@@ -271,6 +271,35 @@ test('address.standardize', (t) => {
...
+        geometry: {
+            type: 'GeometryCollection',
+            geometries: [{
+                type:'MultiLineString',
+                coordinates: [[[-97.2, 37.2],[-97.2, 37.4]]]
+            }]
+        }
+    }, 0), [{
+        i: 0,
+        lf: 100,
```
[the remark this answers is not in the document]
lizziegooding: This was happening because the input geojson returned by https://github.com/mapbox/carmen/blob/9eb7e7b368d5b1efce64d3e2e8f522f3a15b0b79/lib/util/feature.js#L14 was invalid. For example, given an 'carmen:lfromhn' of `[['101'], []]`, this should correctly return a `lf` of `101`. See https://github.com/mapbox/carmen/issues/824.

lib/geocoder/verifymatch.js
```diff
@@ -654,6 +654,10 @@ function sortContext(a, b) {
         // for address results, prefer those from point clusters
         if (a[0].properties['carmen:addressnumber'] && !b[0].properties['carmen:addressnumber']) return -1;
         if (b[0].properties['carmen:addressnumber'] && !a[0].properties['carmen:addressnumber']) return 1;
+
+        // prefer non-interpolated address results
+        if (!a[0].geometry.interpolated && b[0].geometry.interpolated) return -1;
+        if (!b[0].geometry.interpolated && a[0].geometry.interpolated) return 1;
```
miccolis: Is it safe to assume the geometry is always going to be here? Just a couple lines down we do an explicit check to see if the geometry is present.

lib/geocoder/verifymatch.js
```diff
@@ -654,6 +654,10 @@ function sortContext(a, b) {
         // for address results, prefer those from point clusters
         if (a[0].properties['carmen:addressnumber'] && !b[0].properties['carmen:addressnumber']) return -1;
         if (b[0].properties['carmen:addressnumber'] && !a[0].properties['carmen:addressnumber']) return 1;
+
+        // prefer non-interpolated address results
+        if ((a[0].geometry && !a[0].geometry.interpolated) && (b[0].geometry && b[0].geometry.interpolated)) return -1;
+        if ((b[0].geometry && !b[0].geometry.interpolated) && (a[0].geometry && a[0].geometry.interpolated)) return 1;
```
lizziegooding: @miccolis know we're using several different styles in this sorting func ☝️ -- let me know if you'd prefer to see this differently/remove extra parenthesis. I thought it helped with readability but can remove.

PR #826, opened 2019-02-08
Tokenizer plus

miccolis:
### Context

This branch ~attempts to set the stage for unifying~ unites query time tokenization by capturing more information about how a query is split into tokens and eliminating the differences between a token boundary and a replacement boundary.

In addition to restructuring the tokenization itself this also allow us to successfully parse complex addresses like `34 1/2 s vermont avenue #1`, keep number like `21/2-4` in a single token and do a little better job with CJK text that beings with numbers like `123中津川` .

This PR introduces a breaking change where text will only be indexed after global replacements. I think this ok because the scale where they are require is drastically reduced.

There is an upstream change that is likely required in geocoder-tokens. Specifically, regexps need to be compatible with the `u` flag. Of course actually using the cardiality changing replacement support is likely going to mean changes there anyhow.

### Next Steps

- [x] More tests for enumerateTokenReplacements
- [x] Allow replacements to span tokens
- [x] Determine what it will take to invert global/complex replacement with tokenization in all querytime circumstances.
- [x] Tear out the WORD_BOUNDARY separator
- [x] ~See if there is a stopping point that wouldn't require reindexing.~ nope.
- [x] More tests around cardinality change w/ emoji
- [x] Something is going on with the handling of 'ae' ligatures. Probably small, but needs a look.
- [x] I've marked v2 address format support as deprecated, should probably just remove
- [x] Ordering of replacement enumeration & getIndexableText is not as constant as it was previously. We need to decide if that matters. 
- [x] are 8 variants really enough in getIndexable text?
- [x] ~Could we actually remove getMinimalIndexableText?~ Doesn't really get us anything here. punt.
- [x] We need to revisit our stance on handling periods. Running complex replacements before tokenization was more flexible here, we need to develop an opinion now. For examle `test/unit/text-processing/token.replaceToken.test.js` has a replacemnt "S." -> "S" which seems safe to remove. But what if replacement ends in period... what do we do for that?
- [x] ~There is some whitespace trimming that seems excessive or is covering up some bad code. Presently I'm copying forward, probably worth a closer look~ gonna need to live with it.
- [x] extract reverse geocode detection from tokenize
- [x] ~remove mask generation in `permutations`, it's bogus now~ Not exactly true. Added docs.
- [x] numTokenize jsdoc
- [x] rewrite/remove tokenValidator
- [x] some global replacement tests need to become complex replacement tests
- [x] ~should getIndexablePhrases accept just a string array?~ It's fine.
- [x] More tests around cardinality change in addresses
- [x] Add better tests around mask handling in phrasematch
- [x] ~"all permutations" will generate non-contiguous regions which we don't really support. May be time to remove it.~ We want these at index time.
- [x] Non-regex replacements won't have char escaped & we may need to escape them.
- [x] Probably also need to review how we currently identify regexp replacements categorized as complex.
- [x] It's likely that we should strip periods from supplied tokens when we create replacers.
- [ ] There are some docs to write.

cc @mapbox/search


lib/text-processing/termops.js
```diff
@@ -594,7 +638,7 @@ function isAddressNumber(tokens) {
 
 module.exports.numTokenize = numTokenize;
 function numTokenize(text, version) {
-    if (typeof text === 'string') text = tokenize(text);
+    if (typeof text === 'string') text = tokenize(text).tokens;
```
apendleton: As long as we're changing how this works anyway, I wonder if this is the time to get opinionated here and just enforce what type we want. Let's maybe not tolerate a string, and require it to already be tokenized -- and in general, maybe that `{ tokens, separators }` object should be the shape of the thing that gets passed around to all these functions, rather than either a string or an array

miccolis: Agreed on both points. Current state of this PR is just a reasonable stopping point where I could get tests passing.

miccolis:
@apendleton fair points. I commit to adding solid docs to all the phrasematch methods (in fact I think I've done this already in the `no-warning` branch.  And yea, we should harden the tokenized query format in the future.

Thanks of spending the time working through this. Gonna merge!

PR #832, opened 2019-02-27
Boundaries plus regexes

apendleton:
@miccolis these are the changes to your branch that were necessary to get carmen-indexer tests to pass. Will just comment inline with what each one does.

### DIFF
--- a/lib/text-processing/closest-lang.js
+++ b/lib/text-processing/closest-lang.js
@@ -80,9 +80,9 @@ const hasLanguage = function(str) {
     return (language && (Array.isArray(language) ? language[0].type === 'language' : language.type === 'language'));
 };
 
-/* from http://stackoverflow.com/questions/3446170/escape-string-for-use-in-javascript-regex */
+/* from https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions */
 const escapeRegExp = function(str) {
-    return str.replace(/[\-\[\]\/\{\}\(\)\*\+\?\.\\\^\$\|]/g, '\\$&');
+    return str.replace(/[.*+?^${}()|[\]\\]/g, '\\$&'); // $& means the whole matched string
 };
 
 const getScriptComponent = function(mLang) {
@@ -339,4 +339,5 @@ module.exports.closestLangLabel = closestLangLabel;
 module.exports.getText = getText;
 module.exports.fallbackRanking = fallbackRanking;
 module.exports.fallbackMatrix = fallbackMatrix;
+module.exports.escapeRegExp = escapeRegExp;
 
--- a/lib/text-processing/token.js
+++ b/lib/text-processing/token.js
@@ -423,7 +423,7 @@ module.exports.categorizeTokenReplacements = function(geocoder_tokens) {
 
             let replacementOpts = {};
             let to = orig_to;
-            if (typeof orig_to == 'object' && orig_to.text) {
+            if (typeof orig_to == 'object' && typeof orig_to.text == 'string') {
                 replacementOpts = orig_to;
                 to = orig_to.text;
             }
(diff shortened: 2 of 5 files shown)


lib/text-processing/closest-lang.js
```diff
@@ -80,9 +80,9 @@ const hasLanguage = function(str) {
     return (language && (Array.isArray(language) ? language[0].type === 'language' : language.type === 'language'));
 };
 
-/* from http://stackoverflow.com/questions/3446170/escape-string-for-use-in-javascript-regex */
+/* from https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions */
 const escapeRegExp = function(str) {
-    return str.replace(/[\-\[\]\/\{\}\(\)\*\+\?\.\\\^\$\|]/g, '\\$&');
+    return str.replace(/[.*+?^${}()|[\]\\]/g, '\\$&'); // $& means the whole matched string
```
apendleton: The old one sometimes generated regexes that the strict checking flag didn't like, and now I'm using this function in a place where that matters (to escape replacements that aren't regex=True), so it needed to produce ones pass master. Still copy and paste, but hopefully-better copy and paste.

Sidenote: this could move out of this file now that it's also used elsewhere, but there wasn't an obvious place to put it, so I left it alone for the moment. Can revisit.

lib/text-processing/token.js
```diff
@@ -103,9 +104,9 @@ module.exports.createComplexReplacer = function(tokens, inverseOpts) {
 
             const entry = {};
             if (replacementOpts.skipBoundaries) {
-                entry.from = new RegExp(from, 'giu');
+                entry.from = new RegExp(replacementOpts.regex ? from : escapeRegExp(from), 'giu');
```
apendleton: This is the explicit escaping of periods, etc., in replacements that aren't marked regex=True

lib/text-processing/token.js
```diff
@@ -423,7 +424,7 @@ module.exports.categorizeTokenReplacements = function(geocoder_tokens) {
 
             let replacementOpts = {};
             let to = orig_to;
-            if (typeof orig_to == 'object' && orig_to.text) {
+            if (typeof orig_to == 'object' && typeof orig_to.text == 'string') {
```
apendleton: In replacements of the replace-something-with-nothing flavor (apartment numbers, etc.), orig_to.text will be an empty string, which is falsy. That did weird things here.

lib/text-processing/token.js
```diff
@@ -454,7 +456,7 @@ module.exports.categorizeTokenReplacements = function(geocoder_tokens) {
                         .toLowerCase()
                 );
                 const simple_to = removeDiacritics(
-                    orig_to
+                    to
```
apendleton: I *think* this is what's wanted here? Wasn't entirely sure, but as it was, sometimes we would end up calling `.replace` on an object.

test/acceptance/geocode-unit.tokens.test.js
```diff
@@ -83,7 +83,7 @@ const addFeature = require('../../lib/indexer/addfeature'),
     const conf = {
         address: new mem({
             maxzoom: 6,
-            geocoder_tokens: { 'q([a-z])([a-z])([a-z])': '$3$2$1' }
+            geocoder_tokens: { 'q([a-z])([a-z])([a-z])': { text: '$3$2$1', regex: true } }
```
apendleton: Now things that aren't explicitly marked as regexes are escaped, so I went through and marked them all. Alternatives would be to treat things as regexes if they're explicitly marked *or* if they contain parens (bot not periods, because of `N.T.`), or to assume not-escaping and make escaping opt-in rather than opt-out (so `N.T.` could be explicitly exempted). I think this flavor is probably cleanest, but it introduced some diff noise.

Issue #827, opened 2019-02-12
Let address indexes express expected address order


apendleton:
### Context
Different countries format addresses differently. In some places the house number typically comes before the street name and in other places the number comes after. Carmen supports both formats and standardizes them to the same internal representation. It currently doesn't know, though, what the expected position of the house number is in any given country, and so doesn't make use of whether or not the house number in an input query is where we'd expect it to be. This poses a problem in cases where an address contains both a house number before the street name and a unit number after; Carmen will consider two possible parses of such a query, one in which the number before is treated as the housenumber and one after, and has no signal for deciding which is the more likely parse.

We should create a mechanism for communicating which position is expected in a given address index, and penalizing candidate parses which have the address number in a position other than the expected one.

The attempt at rearranging a given subquery is here: https://github.com/mapbox/carmen/blob/master/lib/text-processing/termops.js#L466-L469 and we return a thing that has properties associated with it, so there probably needs to be new input to this function that communicates expected address order, and then we need to check if there's an address number in either the front or the back (currently we only check if there's one at the back) and to compare to that input, and if so, attach an additional property to the permutation that dings the `editMultiplier` slightly if it gets tripped up.

We'll also want to handle the case where the index doesn't express an opinion one way or the other about housenumber position (which will be the typical case) -- in these situations, we won't apply a penalty either way, so this will be opt-in on an index-by-index basis: indexes might have an expected housenumber position of start or end, or no expected position at all.

For an example of adding a new index configuration option, see https://github.com/mapbox/carmen/pull/775 .

refs #809 

### Next Steps

@mapbox/search

PR #833, opened 2019-03-08
🔪 termops.maskAddress

apendleton:
### Context
This branch tweaks boundaries-plus to delete the maskAddress function and corresponding tests, and replace them with a property passthrough mechanism. In current master, each phrasematch for an address index potentially represents a hypothesis for which token in the query represents an address number, and we perform a geocode based on that hypothesis, but we don't actually store it as we proceed, so by the time we get to verifymatch, we don't remember what it is anymore, and have to guess, and sometimes we guess wrong. Some changes around token replacement in boundaries-plus and other repos seem to make the incorrect guessing somewhat more likely in some circumstances, especially in Japanese when using a newly-possible non-global approach to Japanese number standardization, which has forced our hand in dealing with this now.

The approach is basically to stop guessing, and instead remember what we did in the first place (as proposed in #648 and elsewhere). Steps:

* numTokenize now stores what the original number and position were of the number it's replacing
* permutations checks to see as it generates each permutation if that permutation contains stored address metadata supplied by numTokenize
* phrasematch copies the address info off of the permutation and onto the fuzzy match response, and resets the position to be the true position (using the `owner` map) if necessary
* it also now copies the address metadata on gap expansion, if necessary
* in spatialmatch, after stacking, for each stack before it's coalesced, any address metadata associated with any element in the stack is stored, keyed by index
* after coalesce, the indexes in the response elements in the stack are used to re-retrieve the address metadata and reattach it to the mask
* in verifymatch, rather than calling maskAddress, we just use the address data passed in on the spatialmatch object

Incidentally this also lays some groundwork that I think will make #827 easier.
 
### Summary of Changes
- [x] add logic to store and pass through number and position
- [x] delete maskAddress and corresponding unit tests
- [x] update other tests with new additional expected metadata
- [x] add new Japanese address test that fails on boundaries-plus because maskAddress guessed wrong

### Next Steps
nope

cc @mapbox/search


lib/geocoder/spatialmatch.js
```diff
@@ -253,10 +260,10 @@ function allowed(stacks, options) {
     return filtered;
 }
 
-function stackByIdx(stack) {
-    const byIdx = {};
+function getStackByIdx(stack) {
```
apendleton: stackByIdx referred to a function in some parts of the file and the results you got from calling the function in others, which was confusing. Made the data always be stackByIdx and the function by getStackByIdx. (Also made it return a map because yay maps, and to make it more consistent with newer address stuff.)

apendleton:
This is failing codecov because of a single line that I made a change to that wasn't covered in the original branch either, and I have no idea what it does (but I had to fix it because I changed the type of an affected variable). Since this is not a PR into master, I'm inclined not to worry about it.

lib/geocoder/verifymatch.js
```diff
@@ -3,7 +3,6 @@ const addressItp = require('./addressitp');
 const addressCluster = require('./addresscluster');
 const queue = require('d3-queue').queue;
 const context = require('./context');
-const termops = require('../text-processing/termops');
```
miccolis: This make me unreasonably happy.

Issue #834, opened 2019-03-11
Eliminate JSDoc warnings


miccolis:
Currently we get 249 JSDoc warnings when we lint master. The warnings aren't helping us and make any bury real lint issues in a wall of text. We should resolve all the errors. This ticket takes over from #506 

I've been picking away at this in the [no-warning](https://github.com/mapbox/carmen/tree/no-warning) branch. I'm down to 123 warnings spread across the following files:

- [ ] carmen/lib/geocoder/addressitp.js
- [ ] carmen/lib/geocoder/context.js
- [ ] carmen/lib/geocoder/filter-sources.js
- [ ] carmen/lib/geocoder/geocode.js
- [ ] carmen/lib/geocoder/phrasematch.js
- [ ] carmen/lib/geocoder/spatialmatch.js
- [ ] carmen/lib/geocoder/verifymatch.js
- [ ] carmen/lib/indexer/addfeature.js
- [ ] carmen/lib/indexer/index.js
- [ ] carmen/lib/sources/api-mem.js
- [ ] carmen/lib/text-processing/closest-lang.js
- [ ] carmen/lib/util/feature.js

Lets fix all of them.

This is not a PR because if it we have mergeable docs we should merge them and keep moving, lets open PRs at reasonable times. We can close this ticket when all warnings are fixed and JSDoc problems cause automated test runs to fail.

Two important points:

* This is a call for help! If you want to help with a file add your name to it and push commits
* Making JSDoc problems fail tests doesn't mean we have to do everything 100% ...in fact I've already just disabled that rule for any test files that don't currently pass. If it makes more sense to not document something thats fine, rare, but fine.

@mapbox/search

ingalls:
I'm taking on 
- lib/geocoder/addresscluster.js
- lib/geocoder/addressitp.js

Ref: https://github.com/mapbox/carmen/pull/810

miccolis:
Quick update; as of v28.1.0 we're down to 89 warnings.

PR #835, opened 2019-03-11
Less warnings

miccolis:
### Context

First pass at #834 

### Next Steps

- [ ] ~See if I can't get it down to zero, or merge before then next set of big changes.~

Looks like things are going to start moving quickly in camen again. Would be nice to get these docs in before then. There are also a handful of code changes here. Most importantly, I've removed some a `uniq` method that can be replaced with usage of `Set`

cc @mapbox/search


lib/geocoder/geocode.js
```diff
@@ -116,7 +116,7 @@ module.exports = function(geocoder, query, options, callback) {
         });
         if (invalidLanguages.length) return callback(errcode('\'' + invalidLanguages.join(',') + '\' is not a valid language code', 'EINVALID'));
         if (options.language.length > 20) return callback(errcode('options.language should be a list of no more than 20 languages', 'EINVALID'));
-        if (uniq(options.language).length !== options.language.length) return callback(errcode('options.language should be a list of unique language codes', 'EINVALID'));
+        if ((new Set(options.language)).size !== options.language.length) return callback(errcode('options.language should be a list of unique language codes', 'EINVALID'));
```
miccolis: This is a functional change.

lib/indexer/cxxcache.js
```diff
@@ -58,7 +57,10 @@ exports = module.exports = cache;
             }
         }
 
-        result = type === 'grid' ? result : uniq(result);
+        if (type !== 'grid') {
+            result = Array.from(new Set(result));
+            result.sort((a, b) => a - b);
+        }
```
miccolis: This is a functional change.

lib/util/analyze.js
```diff
@@ -38,7 +37,7 @@ module.exports = function analyze(source, callback) {
         const rels = grids.length;
 
         // Verify that relations are unique.
-        if (rels !== uniq(grids).length) {
+        if (rels !== (new Set(grids)).size) {
```
miccolis: This is a functional change.

miccolis:
I removed the `getall` method, looks like it really was dead code.

PR #836, opened 2019-03-12
Configure proximity radius on the index level

Nmargolis:
### Context
Issues: #787

### Summary of Changes
- [x] Passes a proximity radius specified on an index through to coalesce
- [x] Slightly increases the penalty for features where an address number isn't found (e.g. street fallbacks), since more of these make it into final results when address indexes have a higher proximity radius
- [x] Use [carmen-cache version 0/27.0](https://github.com/mapbox/carmen-cache/pull/145) that specifies a distance floor

cc @mapbox/search


lib/geocoder/spatialmatch.js
```diff
@@ -81,12 +81,16 @@ function spatialmatch(query, phrasematchResults, options, callback) {
...
+                while (l--) {
+                    maxZoom = Math.max(maxZoom, stack[l].zoom);
+                    maxRadius = Math.max(maxRadius, stack[l].radius || 0);
+                }
                 coalesceOpts.centerzxy = proximity.center2zxy(
                     options.proximity,
                     maxZoom
                 );
-                coalesceOpts.radius = constants.COALESCE_PROXIMITY_RADIUS;
+                coalesceOpts.radius = maxRadius;
```
apendleton: Hm, do we want this to be the max in the stack, or just the radius of the index of the most specific feature? In practice for the moment we're only doing this with addresses, but hypothetically if we weren't, it seems like this could introduce the possibility that we'd be calculating the distance to the most specific feature using the radius for some other feature in the stack, which seems a little surprising.

lib/util/proximity.js
```diff
@@ -112,29 +113,33 @@ function scoreWeight(score, minScore, maxScore) {
...
+ * @param {String} radius The proximity radius defined on the index the feature is a part of.
  * @return {Number} scaled radius in miles
  */
-function scaleRadius(zoom) {
-    let radius;
-    if (zoom <= 6) {
+function scaleRadius(zoom, radius) {
+    if (radius) {
+        // The proximity radius is configured on the index level
+        return radius;
```
apendleton: It's surprising to me that if a value is ever passed into this function for `radius`, we just don't do anything and return it again. Like, the purpose of passing the radius property to this function is to communicate that we didn't actually want to use this function? Seems like the callsite should maybe be `(radius ? radius : scaleRadius(zoom))` and this function should stay the same.

test/acceptance/geocode-unit.address-proximity-radius.test.js
```diff
@@ -0,0 +1,111 @@
...
+                buildQueued(conf[c], cb);
+            });
+        });
+        q.awaitAll(t.end);
+    });
+
+    tape('Forward search for address', (t) => {
+        c.geocode('fake', { limit: 10}, (err, res) => {
+            t.ifError(err);
+            console.warn('\x1b[35m\x1b[7m\x1b[1m%s\x1b[0m', '\n   res:  ','\x1b[1m', res, '\x1b[0m');
```
apendleton: Assuming these will go away before this is done...

lib/geocoder/spatialmatch.js
Nmargolis: most specific feature! this is updated

lib/util/proximity.js
Nmargolis: Good point, but this is no longer relevant given that the custom radius will only be passed to coalesce and not used here

lib/geocoder/phrasematch.js
```diff
@@ -476,7 +477,8 @@ function PhrasematchResult(phrasematches, source) {
 * Attributes used by carmen-cache, all phrasematches from the same source have the same values for idx, cache, zoom
 **/
 module.exports.Phrasematch = Phrasematch;
-function Phrasematch(subquery, weight, mask, phrase, scorefactor, idx, cache, zoom, prefix, languages, editMultiplier, proxMatch, catMatch, partialNumber, extendedScan, address) {
+
+function Phrasematch(subquery, weight, mask, phrase, scorefactor, idx, cache, zoom, radius, prefix, languages, editMultiplier, proxMatch, catMatch, partialNumber, extendedScan, address) {
```
miccolis: @Nmargolis while you're here, could you add JSDoc to this? ...even if half the descriptions are just TODO

lib/geocoder/phrasematch.js
```diff
@@ -472,11 +473,28 @@ function PhrasematchResult(phrasematches, source) {
...
+ * @param {Object} cache carmen-cache grid
+ * @param {Number} zoom zoom of the source
+ * @param {Number} radius proximity radius of the source to use in coalesce
+ * @param {Number} prefix the ending type of the match - nonPrefix: 0, anyPrefix: 1, wordBoundaryPrefix: 2
+ * @param {Array} languages langagues matched
+ * @param {Number} editMultiplier an approximation of a levenstein ratio
+ * @param {boolean} proxMatch whether the proximity point is inside the source bounds, or false if no proximity was specified
+ * @param {boolean} catMatch whether the phrase matches any categories specified on the source index
+ * @param {boolean} partialNumber whether the phrase is a number-only query
+ * @param {boolean} extendedScan whether an extendedScan was done
```
Nmargolis: TODO: whether or not carmen-cache should do an extended scan

PR #837, opened 2019-03-13
Support preferred address order for address indexes

### LINKED ISSUE #827: Let address indexes express expected address order

apendleton:
### Context
#827 proposes introducing a mechanism to let indexes specify whether in the affected country it's expected than a house number come before or after the street name, and applies a penalty to candidate results which assume a parse in which the house number comes in the position other than the expected one. It builds to some extent on #833, which captured metadata about query components being interpreted as house numbers, and adds additional metadata about whether any given interpretation of the query assumes the number comes first or last (or, in some sense, both -- marked with a `null` value in the code -- if the subquery is only a single numerical token long). This can then be compared to the index's expected number order, if any, and a penalty applied if they don't match.

Also, tests.

fixes #827

### Summary of Changes
- [x] capture whether the number is being interpreted as first or last (or both)
- [x] compare that to the index's expected order, if any
- [x] tests

### Next Steps
- [ ] check passrates
- [ ] incorporate into index generation pipeline

cc @mapbox/search


test/acceptance/geocode-unit.address-preferred-order.test.js
```diff
@@ -0,0 +1,164 @@
+'use strict';
+// Test to make sure identified house numbers in Japan are properly passed along
```
miccolis: Is this comment a copy forward?

PR #838, opened 2019-03-14
Intersection search

aarthykc:
### Context
Adding unit tests for intersection search and ability to  index and then search for an intersection. 

### Explanation 
The problem that we currently have with carmen especially with addresses is that we index a lot of features that have the same relevance and score and because of the limits we have in place for verifymatch and spatialmatch, a specific feature we'd like to surface, just by virtue of being lower down in the sort id filtered out. 
While thinking of how to solve this problem I started looking at how we index house numbers. We don’t create a new feature for every house number: we have one feature and an `addressnumber` array with all the house numbers. Essentially, for intersection search we'll want to take a similar approach, i.e. we want fewer features which more densely populated. 

The initial data spec that @ingalls worked on, looked something like this: 

    {
        id: 3,
        properties: {
            'carmen:text': '9th st nw and 15th st nw,15th st nw and 9th st nw',
            'carmen:center': [0,0],
        },
        geometry: {
            type: 'Point',
            coordinates: [0,0]
        }
    };

The problem with this approach however was that because of the aforementioned limits we have in place, we only ever surface intersections and only when the user completely types out the street name like `9th Street Northwest` do we surface the actual street. The underlying problem being that we don’t have a good way to know whether the user is searching for an intersection or just the street. This PR aims to trigger intersection search only after the user starts typing `and`. 

What new approach are we using:

### Index Time:
1. At indexing time, we index intersections in the `carmen:intersections` in the following manner: 
`'carmen:intersections': [['highway number 2']]` → `['+intersection highway number 2 ,']`
We use `+intersection` as a special tag for intersections and we use the `,` as a delimiter between the intersection and the actual street name. 

2. We standarize Multipoints to GeometryCollections to make it easier to generate intersection data and also reduce the number of features indexed/ size of the index. Therefore, now intersections don’t have to be a separate address cluster.

### Query Time:
1. At query time, we look for the keyword  ‘and’ which we use as a trigger for intersection search. For example, `'9th street northwest and f street northwest'` gets tokenized to `['+intersection', '9th', 'street', 'northwest', ',', 'f', 'street', 'northwest']`
2. We check whether the search is an intersection search by checking whether the cover text has the special tag `+intersection` if it does then we look up the corresponding geometry in the `GeometryCollection`. 
3. Once we have the geometry and the intersection name we stitch the place_name together which return to the user.

### Data Spec:
1. For the data design, let’s assume: 
- 9th Street Northwest which is also known as US Road, Highway Number 6
- F Street Northwest which is also known as Highway Number 4

#### Description: 

- 9th Street Northwest intersects with F Street Northwest at [0,2] and Highway Number 2 at [0,1]
- The aliases for 9th Street Northwest go in carmen:text separated by a comma. While the streets 9th Street Northwest intersects with goes in the carmen:intersections property.
- The coordinates in the geometry correspond to the point of intersection of the street in carmen:intersections and the feature in carmen:text.
    
```    
    Example data spec for intersections:
    {
        const address = {
            id:2,
            properties: {
                // Synonyms of the feature go in 'carmen:text'
                'carmen:text': 'US Road,Highway Number 6,9th Street Northwest',
                'carmen:center': [0,0],
                // intersections with the feature go here
                'carmen:intersections': ['F Street Northwest', 'Highway Number 2']
            },
            geometry: {
                type: 'MultiPoint',
                // coordinates correspond to the street in 'carmen:intersection'
                // for example, F Street Northwest and 9th Street Northwest intersect at [0,2]
                coordinates: [[0,2], [0,1]]
            }
        };
        queueFeature(conf.address, address, t.end);
    });
```

- Intersections can also represented as a GeometryCollection by adding  `carmen:intersections`: `[null,['1st Avenue', '2nd Avenue', '3rd Avenue'], null]` property
- And the corresponding coordinates in the geomteries property
-  The position in the geometries array corresponding to the non null array value in `'carmen:intersections'`, similar to addressnumbers
- We standardise a multipoint intersection feature as a GeometryCollection to allow intersections to be part of the same address cluster and avoid having to index a feature and it’s intersection separately

```
    {
        properties: {
            'carmen:text': 'Main Street Northwest',
            'carmen:addressnumber': [[1, 2, 3], null, null],
            'carmen:intersections': [null,['1st Avenue', '2nd Avenue', '3rd Avenue'], null],
            'carmen:rangetype': 'tiger',
            'carmen:parityl': [[], [], []],
            'carmen:parityr': [[], [], []],
            'carmen:lfromhn': [[], [], []],
            'carmen:rfromhn': [[], [], []],
            'carmen:ltohn': [[], [], []],
            'carmen:rtohn': [[], [], []]
        },
        geometry: {
            type: 'GeometryCollection',
            geometries: [{
                type: 'MultiPoint',
                coordinates: [[1,1], [2,2], [3,3]] <--- addressnumber MultiPoint array is 1st, since addressnumber: [[1, 2, 3], null, null]
            },{
                type: 'MultiPoint',
                coordinates: [[2,2], [3,2], [3,3]] <--- intersections MultiPoint array is 2nd, since intersections: [null,['1st Avenue', '2nd Avenue', '3rd Avenue'], null]
            },{
                type: 'MultiLineString',
                coordinates: [[[1,1], [2,2], [3,3]]]
            }]
        }
    };
```

- If there is more than one name for F Street Northwest and it intersects with 9th Street Northwest, the alias is added to carmen:intersections and the corresponding coordinates in the coordinates property of the geometry.

### Summary of Changes
- [x] Adds unit tests for intersection search 
   - Does this new approach surface streets before intersections 
   - Are we able to find intersections?
   - Can we find intersections for abbreviated versions of 9th Street Northwest and F Street Northwest like 9th st nw and F st nw
- [x] Adds special indexing intersections
- [x] Adds geomtery lookup for intersections in the geometry collection
- [x] Support to return the name of the street and the intersection
- [x] Don’t hard code the replacement `and`, have a way to specify this as a part of the index
- [x] Trigger an intersection search only after the user types 'and'

Next Actions:
- [ ] Can this approach of tagged text we used beyond `and` for other languages or other query patterns such as `between X and Y`
- [ ] Global token support for allowing users to query for Street A & Street B

cc @mapbox/search

### DIFF
--- a/lib/geocoder/ops.js
+++ b/lib/geocoder/ops.js
@@ -346,6 +346,8 @@ function getMatchingText(item, geocoder, requestedLanguage) {
 
 /*
 * getIntersectionName - helper function to get the street to add to the cross street
+* this is for Street B in Street A and Street B
+* the carmen:text field has a number of synonyms, we need to pick out the one similar to what the user typed
 * @param {Object} geocoder Geocoder object
 * @param {object} properties A feature properties object
 * @return String returns the correct carmen:text synonym based on the input query
@@ -354,12 +356,10 @@ function getIntersectionStreetName(geocoder, properties) {
     if (properties['carmen:intersection'] && properties['carmen:query_text'] !== undefined) {
         let featureSynonymMatch;
         const source = geocoder.byidx[properties['carmen:idx']];
-        properties['carmen:text'].split(',').forEach((sourceText) => {
-            const texts = termops.getIndexableText(source.simple_replacer, source.complex_query_replacer, geocoder.replacer,
-                { properties: { 'carmen:text': sourceText } }
-            );
-            if (texts[0].tokens.join(' ').indexOf(properties['carmen:query_text']) >= 0) {
-                featureSynonymMatch = sourceText;
+        properties['carmen:text'].split(',').forEach((synonymText) => {
+            const tokenizedIntersection = termops.tokenize(synonymText).map((token) => source.simple_replacer.tokens.get(token) || token);
+            if (tokenizedIntersection.join(' ').indexOf(properties['carmen:query_text']) >= 0) {
+                featureSynonymMatch = synonymText;
             }
         });
         return featureSynonymMatch.trim();
--- a/lib/geocoder/ops.js
+++ b/lib/geocoder/ops.js
@@ -54,7 +54,7 @@ function getPlaceName(context, formatString, language, languageMode, matched) {
             prefix +
             context
                 .filter((f) => { return matched || featureMatchesLanguage(f, { language: [language], languageMode: languageMode }); })
-                .map((f) => { return (matched ? f.matching_text : undefined) || closestLang.getText(language, f.properties).text; })
+                .map((f) => { return (matched ? f.matching_text : undefined) || (f.properties['carmen:intersection'] ? getIntersectionName(f.properties) : undefined) || closestLang.getText(language, f.properties).text; })
                 .join(', ')
         ).trim();
     } else {
@@ -333,5 +333,16 @@ function getMatchingText(item, geocoder, requestedLanguage) {
             matched.matching_language = matching_language.replace('_', '-');
         }
         return matched;
+    }
+}
+
+function getIntersectionName(properties) {
+    if (properties['carmen:intersection'] && properties['carmen:query_text'] !== undefined) {
+        const featureSynonymMatch = properties['carmen:text'].split(',').find((f) => {
+            // find the feature or the synonym of the feature in carmen:text
+            // return the text to be added to the place_name
+            return f.toLowerCase().indexOf(properties['carmen:query_text'].toLowerCase()) >= 0;
+        });
+        return featureSynonymMatch.trim();
     }
 }
--- a/lib/geocoder/ops.js
+++ b/lib/geocoder/ops.js
@@ -47,9 +47,8 @@ function getPlaceName(context, formatString, language, languageMode, matched) {
     // Use geocoder_format to format output if applicable
     if (!formatString || formatString === true || formatString === 1) {
         place_name = (
-            (feat.properties['carmen:address'] ?
-            feat.intersections ?
-            feat.properties['carmen:address'] + ' and ' : feat.properties['carmen:address'] + ' ' : '') +
+            (feat.properties['carmen:intersection'] ? feat.properties['carmen:intersection'] + ' and ' : '') +
+            (feat.properties['carmen:address'] ? feat.properties['carmen:address'] + ' ' : '') +
             context
                 .filter((f) => { return matched || featureMatchesLanguage(f, { language: [language], languageMode: languageMode }); })
                 .map((f) => { return (matched ? f.matching_text : undefined) || closestLang.getText(language, f.properties).text; })
--- a/lib/geocoder/ops.js
+++ b/lib/geocoder/ops.js
@@ -82,7 +82,8 @@ function getPlaceName(context, formatString, language, languageMode, matched, ge
                     if (templateSubprop === '_name') {
                         if (f.properties['carmen:intersection']) {
                             const intersectionPrefix = feat.properties['carmen:intersection'] + ' and ';
-                            val = intersectionPrefix + getIntersectionName(geocoder, f.properties);
+                            const intersectionSuffix = getIntersectionStreetName(geocoder, f.properties);
+                            if (intersectionSuffix) val = intersectionPrefix + intersectionSuffix;
                         }
                         // `name` is a hardcoded subproperty that maps to carmen:text
                         else if (matched && f.matching_text) val = f.matching_text;
@@ -349,7 +350,7 @@ function getMatchingText(item, geocoder, requestedLanguage) {
 * @param {object} properties A feature properties object
 * @return String returns the correct carmen:text synonym based on the input query
 */
-function getIntersectionName(geocoder, properties) {
+function getIntersectionStreetName(geocoder, properties) {
     if (properties['carmen:intersection'] && properties['carmen:query_text'] !== undefined) {
         let featureSynonymMatch;
         const source = geocoder.byidx[properties['carmen:idx']];
--- a/lib/geocoder/ops.js
+++ b/lib/geocoder/ops.js
@@ -81,7 +81,8 @@ function getPlaceName(context, formatString, language, languageMode, matched, ge
                 if (templateProp === carmenProp) {
                     if (templateSubprop === '_name') {
                         if (f.properties['carmen:intersection']) {
-                            const intersectionPrefix = feat.properties['carmen:intersection'] + ' and ';
+                            const source = geocoder.byidx[f.properties['carmen:idx']];
+                            const intersectionPrefix = f.properties['carmen:intersection'] + ' ' + source.geocoder_intersection_token + ' ';
                             const intersectionSuffix = getIntersectionStreetName(geocoder, f.properties);
                             if (intersectionSuffix) val = intersectionPrefix + intersectionSuffix;
                         }
@@ -357,7 +358,7 @@ function getIntersectionStreetName(geocoder, properties) {
         let featureSynonymMatch;
         const source = geocoder.byidx[properties['carmen:idx']];
         properties['carmen:text'].split(',').forEach((synonymText) => {
-            const tokenizedIntersection = termops.tokenize(synonymText).tokens.map((token) => source.simple_replacer.tokens.get(token) || token);
+            const tokenizedIntersection = source.simple_replacer.replacer(termops.tokenize(synonymText).tokens);
             if (tokenizedIntersection.join(' ').indexOf(properties['carmen:query_text']) >= 0) {
                 featureSynonymMatch = synonymText;
             }
--- a/lib/geocoder/ops.js
+++ b/lib/geocoder/ops.js
@@ -82,7 +82,7 @@ function getPlaceName(context, formatString, language, languageMode, matched, ge
                     if (templateSubprop === '_name') {
                         if (f.properties['carmen:intersection']) {
                             const intersectionPrefix = feat.properties['carmen:intersection'] + ' and ';
-                            val = (intersectionPrefix + context.map((f) => { return (f.properties['carmen:intersection'] ? getIntersectionName(geocoder, f.properties) : undefined); }));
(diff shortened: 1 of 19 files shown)


lib/geocoder/ops.js
```diff
@@ -328,3 +344,25 @@ function getMatchingText(item, geocoder, requestedLanguage) {
         return matched;
     }
 }
+
+/*
+* getIntersectionName - helper function to get the street to add to the cross street
```
apendleton: That's not what this function appears to be called?

lib/geocoder/ops.js
```diff
@@ -328,3 +344,25 @@ function getMatchingText(item, geocoder, requestedLanguage) {
...
+* this is for Street B in Street A and Street B
+* the carmen:text field has a number of synonyms, we need to pick out the one similar to what the user typed
+* @param {Object} geocoder Geocoder object
+* @param {object} properties A feature properties object
+* @return String returns the correct carmen:text synonym based on the input query
+*/
+function getIntersectionStreetName(geocoder, properties) {
+    if (properties['carmen:intersection'] && properties['carmen:query_text'] !== undefined) {
+        let featureSynonymMatch;
```
apendleton: This isn't initialized with a default value...

lib/geocoder/ops.js
```diff
@@ -328,3 +344,25 @@ function getMatchingText(item, geocoder, requestedLanguage) {
...
+        const source = geocoder.byidx[properties['carmen:idx']];
+        properties['carmen:text'].split(',').forEach((synonymText) => {
+            const tokenizedIntersection = source.simple_replacer.replacer(termops.tokenize(synonymText).tokens);
+            if (tokenizedIntersection.join(' ').indexOf(properties['carmen:query_text']) >= 0) {
+                featureSynonymMatch = synonymText;
+            }
+        });
+        return featureSynonymMatch.trim();
```
apendleton: ... so we might never have set it by the time we get here, and might call `trim` on an undefined.

lib/geocoder/verifymatch.js
```diff
@@ -140,11 +141,43 @@ function verifyFeatures(query, geocoder, spatialmatches, loaded, options) {
...
 
-        // this is passed-through information from phrasematch about what and
-        // where we think the address number is in this query
         const address = spatialmatch.address;
 
         if (source.geocoder_address) {
+
+            // cover.text for intersections will look like - +intersection f street northwest , 9th street
+            // i.e they will be indexed with +intersection
+            if (cover.text.indexOf('+intersection') === 0) {
```
apendleton: Nitpick, but can this be `if (cover.text.startsWith('+intersection')) {` instead? Would prefer not to search the whole string for something that we know will be at the beginning if it's present at all.

lib/geocoder/verifymatch.js
```diff
@@ -359,7 +391,12 @@ function verifyContexts(contexts, sets, indexes, options, geocoder) {
                 cover = verify[ctx.properties['carmen:tmpid']];
                 ctx.properties['carmen:matches_language'] = cover.matches_language;
                 ctx.properties['carmen:prefix'] = cover.prefix;
-                ctx.properties['carmen:query_text'] = cover.text;
+                // for intersections we want the feature 9th street where the cover.text = +intersection f street northwest , 9th street Northwest
+                if (ctx.properties['carmen:intersection'] || (cover.text.indexOf('+intersection') === 0)) {
```
apendleton: Likewise, can be startsWith

lib/text-processing/termops.js
```diff
@@ -655,6 +694,15 @@ function addressPermutations(permutations) {
...
     const uniq = {};
     const phrases = [];
+    let obj = {};
+    if (text.tokens[0] === '+intersection') {
+        const etext = normalizeText(text.tokens.join(' '));
+        return [{
+            relev: 1,
+            text: etext,
+            phrase: etext
+        }];
```
apendleton: Just to confirm my understanding here: this is to avoid all the frequency/word-dropping stuff, right?

lib/text-processing/termops.js
```diff
@@ -712,7 +760,8 @@ function getWeights(tokens, freq) {
         if (/#/.test(tokens[i])) {
             numTokens = true;
             weights[i] = -1;
-        } else {
+        }
+        else {
```
apendleton: Why this change? Our general style in Carmen is to put `else` on the same line as the closing brace

lib/text-processing/termops.js
```diff
@@ -764,6 +812,46 @@ function numTokenize(text, version) {
...
+    if (text.length < 3) return ret;
+    // intersection token can't be at the start or the end of the query
+    // eg. "and first street" or "first street and"
+    for (let i = 1; i < text.length - 1; i++) {
+        if (text[i] === intersectionToken) {
+            // f st and 9th st
+            // for intersections we need a way to replace the and with a token that we recognise as an intersection
+            // which is +intersection
+            const replaced = text.slice(0);
+            replaced[i] = replaced[i].replace(intersectionToken, ',');
```
apendleton: Hm, why can't this just be `replaced[i] = ','`?

lib/text-processing/termops.js
```diff
@@ -764,6 +812,46 @@ function numTokenize(text, version) {
...
+ * @param {TokenizedQuery} tokens - tokens
+ * @param {string} intersectionToken - 'and' or the like.
+ * @return {Array<string>} permutations array, has a few extras;
+ *        - ender: {boolean}
+ *        - mask: {number}
+ *        - relev: {number}
+ */
+function intersectionPermutations(tokens, intersectionToken) {
+    const ret = [];
+    const text = tokens.tokens; // todo use this to calculate mask
```
apendleton: Nitpick, but elsewhere in the code we have `text` as a string, as opposed to `tokens`, which is an array of strings. Calling the list of tokens `text` is sort of surprising (as it's not text), and made later parts of the function a little tricky to understand.

lib/text-processing/termops.js
```diff
@@ -842,3 +930,25 @@ function decode3BitLogScale(num, max) {
     if (!num || !max) return 0;
     return Math.round(Math.pow(max, num / 7));
 }
+
+/**
+* getIntersectionText - Indexes intersection data differently by marking it with a special word "intersectionsearch"
+* @param {Object} doc The document to be indexed
+* @return {Object} ranges An array of indexed intersection data, false if nothing was indexed
```
apendleton: Looks like it's not always an object, it's an object or bool?

lib/geocoder/verifymatch.js
```diff
@@ -140,11 +141,43 @@ function verifyFeatures(query, geocoder, spatialmatches, loaded, options) {
...
+
+                let geometryPos;
+                let intersectionPos;
+                let exactMatchIntersection;
+                for (let i = 0; i < intersections.length; i++) {
+                    if (exactMatchIntersection) break;
+                    if (!intersections[i]) continue;
+                    for (let j = 0; j < intersections[i].length; j++) {
+                        const tokens = source.simple_replacer.replacer(termops.tokenize(intersections[i][j]).tokens);
+                        if (intersectionQuery === tokens.join(' ')) {
```
apendleton: Is it actually necessary to tokenize each of these and then join them back together? This seems like an expensive thing to have to do at query time if each feature might have lots and lots of intersections, and it's not clear why you can't just compare them directly -- they'll already have been tokenized/normalized at index time, no?

lib/text-processing/termops.js
aarthykc: Yup, thats why!

lib/geocoder/verifymatch.js
aarthykc: Per chat, we're not changing this just yet

lib/geocoder/ops.js
```diff
@@ -76,8 +80,15 @@ function getPlaceName(context, formatString, language, languageMode, matched) {
                 // If we found a match between the template property and one within the returned context
                 if (templateProp === carmenProp) {
                     if (templateSubprop === '_name') {
+                        if (f.properties['carmen:intersection']) {
+                            const source = geocoder.byidx[f.properties['carmen:idx']];
+                            const intersectionPrefix = f.properties['carmen:intersection'] + ' ' + source.geocoder_intersection_token + ' ';
+                            const intersectionSuffix = getIntersectionStreetName(geocoder, f.properties);
+                            if (intersectionSuffix) val = intersectionPrefix + intersectionSuffix;
```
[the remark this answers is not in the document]
aarthykc: yeah, you're right, i remember putting it in there because we were returning `undefined` as a result of `getIntersectionStreetName()` but that went away because now the permutations never allow us to have an empty string for StreetB in `+intersection StreetA, StreetB`.

lib/geocoder/ops.js
```diff
@@ -143,8 +155,10 @@ function toFeature(context, format, languages, languageMode, debug, geocoder, cl
         const formatString = getFormatString(context, format, language);
         memo[`text${suffix}`] = text.text;
         if (text.language) memo[`language${suffix}`] = text.language.replace('_', '-');
-        memo[`place_name${suffix}`] = getPlaceName(context, formatString, language, languageMode);
-
+        if (feat.properties['carmen:intersection']) {
+            if (geocoder) memo[`place_name${suffix}`] = getPlaceName(context, formatString, language, languageMode, false, geocoder);
+        }
```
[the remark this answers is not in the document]
aarthykc: Changed!

lib/geocoder/ops.js
aarthykc: Oops, good catch

lib/geocoder/ops.js
aarthykc: and done.. good catch!

lib/geocoder/verifymatch.js
aarthykc: done

lib/text-processing/termops.js
aarthykc: yes, absolutely, there were few others as well which I cleaned up and totally missed this

lib/text-processing/termops.js
aarthykc: changed this to tokenizedQuery, works?

apendleton:
Want to call out that this branch also included this commit: https://github.com/mapbox/carmen/pull/838/commits/d62b69878606601847201019c8c32ee93265649b which we snuck in, which fixes an issue with partial number filtering that @Nmargolis identified, since it didn't get its own PR.

Issue #839, opened 2019-03-15
Postal Overrides


ingalls:
### Context

Now that support for the `carmen:addressprops` feature has launched, we should explore the ability to allow lower features to override parts of the `context` array in the resultant feature.

Since postal polygons will always have a certain degree of error due to their conversion from routes, this will provide a method in which we can error correct using known postal information about a given address.

#### Overriding
There are two potential parts to this quest - the first is simply overriding the `context` at the end of verifymatch. It could potentially look like something like:

**Address Feature Props**
```
override:postcode: 26884
override:country: 'United States'
```

Where features with these properties would get first dibs at filling in the `geocoder_format` value, before being removed from the properties object.

Gotchas: The context has a feature id that would need to be overwritten as well.

#### Indexing

The second potential operation is to include the postcode in the indexes to allow forward searches for addresses with postal codes that do not match their corresponding postal polygon. I am less confident of the feasibility of this approach.

@mapbox/search

miccolis:
This is available in Carmen 28.5.0

PR #840, opened 2019-03-18
Override

ingalls:
Refs https://github.com/mapbox/carmen/issues/839

Adds support for `addressprops` that begin with `override` to "override" context elements of the feature. The expected usage is for information like postcodes, which are assigned to addresses individually, but are also expected to behave like geographic areas. This functionality allows a feature to specify the correct context, while allowing for fallback to generic polygons. The present functionality is unaware of languages, which limit it's suitability to postcode-like data exclusively.

This PR modifies verifyContexts; immediately after we reverse geocode potential results it checks if any result have an override for a something in it's context. We compare the `type` of the context and the `carmen:text` to detect equality. This check is simplistic; it ignores language and assumes uniqueness, but it works for postcodes. 

When overriding a context we use the value of the `override:<type>` address property as the `carmen:text` and copy forward the center. The id we set is a bit of a hack, it's the type of the override and the numeric idea of the feature that is setting the override. So a override from `address:12345` for a postcode would look like `postcode:12345`. Probably the only nice thing about this is that it's fairly easy to spot when a override has happened by looking at matching numeric ids. As we don't make any assertions about the stability or uniqueness of these ids I think it's likely safe.

There are two changes to the ranking of features here as well:

1. A relevance boost when an context is overridden and that same context is also it's own result. This allows an address with the postcode that doesn't match the postcode polygon to surface in before the zipcode itself. Without this boost both the address and postcode would have equal relevance and the higher scored postcode would get the first position.
2. sortByRelev up in spatial match sorts addresses before non-addresses as a tiebreaker within an index.

Next steps:

- [x] Review.
- [ ] Downstream testing.

cc @mapbox/search

### DIFF
--- a/lib/geocoder/spatialmatch.js
+++ b/lib/geocoder/spatialmatch.js
@@ -54,14 +54,19 @@ function spatialmatch(query, phrasematchResults, options, callback) {
 
     coalesceStacks();
 
-    // coalesce all stacks.
+    /** coalesce all stacks. */
     function coalesceStacks() {
         const q = queue();
         for (let i = 0; i < stacks.length; i++) q.defer(coalesceStack, stacks[i]);
         q.awaitAll(coalesceFinalize);
     }
 
-    // Coalesce a single stack, add debugging info.
+    /**
+     * Coalesce a single stack, add debugging info.
+     * @param {Array} stack - phrasematch results
+     * @param {Callback} callback - callback
+     * @returns {undefined}
+     */
     function coalesceStack(stack, callback) {
         // Proximity option is set.
         // Convert proximity to xy @ highest zoom level for this stack
@@ -123,7 +128,12 @@ function spatialmatch(query, phrasematchResults, options, callback) {
         });
     }
 
-    // Final feature collection and sort.
+    /**
+     * Final feature collection and sort.
+     * @param {Error} err - error
+     * @param {Array<Array<Object>>} results - array of results from carmen-cache
+     * @returns {undefined}
+     */
     function coalesceFinalize(err, results) {
         if (err) return callback(err);
         let combined = [];
@@ -266,8 +276,13 @@ function expandFromArchetypesInner(stack, maxOut, matchIdx, soFar, out) {
     }
 }
 
-// Filter an array of stacks down to only those whose maxidx is allowed
-// by a passed in allowed_idx filter.
+/**
+ * Filter an array of stacks down to only those whose maxidx is allowed
+ * by a passed in allowed_idx filter.
+ * @param {Array} stacks - array of phrasematch results
+ * @param {Object} options - query options
+ * @return {Array} filtered list of stacks
+ */
 function allowed(stacks, options) {
     if (!options.allowed_idx) return stacks;
     const filtered = [];
@@ -283,6 +298,11 @@ function allowed(stacks, options) {
     return filtered;
 }
 
+/**
+ * Generate a map of stack idx to data
+ * @param {Array<object>} stack - list of phrasematches
+ * @return {Map} matched results keyed by index.
+ */
 function getStackByIdx(stack) {
     const byIdx = new Map();
     let l = stack.length;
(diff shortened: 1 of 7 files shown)


lib/geocoder/verifymatch.js
```diff
@@ -327,41 +327,71 @@ function loadContexts(geocoder, features, sets, options, callback) {
...
+                if (feat.properties[prop] && ctx.properties['carmen:text'] !== feat.properties[prop]) {
+                    // Create a minimal new context to replace the recalled one. Proxy
+                    // the geometry and extid forward of the feature which has the override.
+                    context[j] = {
+                        properties: {
+                            'carmen:text': feat.properties[prop],
+                            'carmen:index': ctx.properties['carmen:index'],
+                            'carmen:types': [type],
+                            'carmen:center': feat.properties['carmen:center'],
+                            'carmen:extid': `postcode.${feat.properties['carmen:extid']}`
```
miccolis: This is generating ids like `postcode.address.5658655856915702` so it needs a minor fixup.

lib/geocoder/spatialmatch.js
```diff
@@ -447,7 +474,8 @@ function sortByZoomIdx(a, b) {
 function sortByRelev(a, b) {
     return (b.relev - a.relev) ||
         (b.scoredist - a.scoredist) ||
-        (a.covers[0].idx - b.covers[0].idx);
+        (a.covers[0].idx - b.covers[0].idx) ||
+        (b.address ? 1 : 0) - (a.address ? 1 : 0);
```
apendleton: Checking my own understanding: this sorts results with addresses before those without? or the other way? (I always mix up which order does what)

lib/geocoder/verifymatch.js
```diff
@@ -361,36 +361,65 @@ function loadContexts(geocoder, features, sets, options, callback) {
...
+                            'carmen:extid': `${type}.${feat.properties['carmen:extid'].split('.', 2)[1]}`
+                        }
+                    };
+
+                    // It's possible that we have a peer that is the match for
+                    // the feature we're overriding. If that's the case we need
+                    // bump the relevence of our cover
+                    for (const peer of peers.values()) {
+                        if (indexes[peer.properties['carmen:index']].type === type && peer.properties['carmen:text'] === feat.properties[prop]) {
+                            sets[feat.properties['carmen:tmpid']].relev += 0.05;
```
apendleton: .05 because that's a token's worth? Wondering if we need to do anything to ensure that the relevance can't exceed 1 here (and I know it shouldn't be able to, but uncertain because of rebalancing)

apendleton:
Oh also: what ID will the invented feature in the context have? That's not clear from the PR.

lib/geocoder/spatialmatch.js
miccolis: This will sort features with `address` set before those without... so the full treatment is like

* higher relevance before lower
* lower index position before higher 
* addresses before non-addresses (which would always be features in the same index at this point, so we're talking addresses before streets or intersections)

lib/geocoder/verifymatch.js
miccolis: Right, 0.05 because it's a minimum relev we'd assign to a index (not token, as we're post re-balancing at this stage) and therefore safe to use here because we know we don't have a perfect match. But you're right, this value assumes a few things and it's almost certain that those assumptions will fail eventually. At the very least this should actually reference the constant and it would be great to future proof this against changes in rebalacing. I'll see what I can come up with.

miccolis:
@apendleton I updated the OP with a explanation of the approach here. I'll take a look at a way to future proof the relevance tweak next.

lib/geocoder/verifymatch.js
miccolis: @apendleton I didn't come up with anything creative/future proof here. I changed the code to reference the max tokens constant and used `Math.min` to cap the value. I'm sure we'll revisit this in the future.

PR #841, opened 2019-03-20
Fix empty-phrasematch issue

apendleton:
### Context
A funny interaction occurs between CJK tokenization and emoji removal. It seems that a query that ends in a CJK character followed by an emoji can lead to a set of tokenization/normalization steps that results in a token set in which the final token (where the emoji was originally) is an empty string instead. If instead the query is a Latin-letter word followed by a space followed by an emoji, there's no extra blank token, so apparently something about opting into the alternative tokenization path for CJK makes this happen. (Also: if there are Latin letters immediately next to the emoji, the last token is just those letters, with the emoji removed).

For non-address, indexes, which use windowed search, this presents a problem for autocomplete searches, because the empty string is a prefix for every word, so we're virtually guaranteed to produce a window with a phrase consisting only of the empty string if the empty string is in the final position. If we generate that phrasematch, and also some other phrasematch in a different index elsewhere in the query such that the two can stack and together exceed the 0.5 relevance threshold, we end up creating a stack with one phrasematch with `''` as the phrase.

If *that* happens and we pass it into `coalesce`, it barfs: empty phrases aren't allowed, and we get a `TypeError: encountered invalid phrase` back.

This PR fixes it in a relatively clunky way: just skip generating Phrasematch objects if the phrase is empty, for whatever reason. There might be some reason to instead change the tokenization behavior so that CJK and non-CJK tokenization behavior is consistent, and the former doesn't generate these empty-string tokens in the first place in the presence of emoji, but I think this is probably fine for now -- it solves the immediate problem, anyway.

### Summary of Changes
- [x] modify the emoji test so it allows for stacking and tests a CJK-pattern query
- [x] fix it by adding a check that the phrase being generated is non-empty

### Next Steps
No

cc @mapbox/search

### DIFF
--- a/lib/geocoder/phrasematch.js
+++ b/lib/geocoder/phrasematch.js
@@ -247,7 +247,7 @@ module.exports = function phrasematch(source, query, options, callback) {
 
     for (const subquery of subqueries) {
         const phrase = subquery.join(' ');
-        if (!phrase) continue;
+        if (phrase.length === 0) continue;
         // Adjust weight relative to input query.
         const b = findMaskBounds(subquery.mask, MAX_QUERY_TOKENS);
         const weight = (b[1] - b[0] + 1) / query.tokens.length;
--- a/lib/geocoder/phrasematch.js
+++ b/lib/geocoder/phrasematch.js
@@ -247,6 +247,7 @@ module.exports = function phrasematch(source, query, options, callback) {
 
     for (const subquery of subqueries) {
         const phrase = subquery.join(' ');
+        if (!phrase) continue;
         // Adjust weight relative to input query.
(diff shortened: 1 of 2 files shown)


miccolis:
Eeek, that's a good bug. Thanks for digging in & sorry about that! What do you think about moving the diacritic & emoji stripping before the length check back on normalizeQuery.  Basically moving https://github.com/mapbox/carmen/blob/62eae06e419431f47bc873d58ee42a5a621032c1/lib/text-processing/termops.js#L98 up a couple lines?

apendleton:
@miccolis yeah I'm fine with that also, but didn't know the code well enough to feel confident messing with it. Sounds per voice like we can just go ahead with this change and circle back and do the simpler thing later.

PR #845, opened 2019-03-21
Fixes fatal aborts that occur because of Invalid regular expression 

aarthykc:
### Context
Seeing a few fatal aborts because of "Invalid regular expression". If this function is just for checking the occurrence of a string, we should avoid using regexes. This PR replaces the regex approach with `indexOf`.

### Summary of Changes
- [x] Changes to `feature.js`

### Next Steps
- [x] reviewed
- [ ] merged and deployed after coordinating with @ingalls 

cc @mapbox/search

### DIFF
--- a/lib/util/feature.js
+++ b/lib/util/feature.js
@@ -315,16 +315,12 @@ function matchScore(features, cover, source) {
 
 // If score isn't enought to match a cover to a single feature, look at text properties.
 function matchText(features, cover) {
-    let re;
     features = features.filter((feature) => {
         let matches = false;
         for (const key of Object.keys(feature.properties)) {
             if (!feature.properties[key] || !/^carmen:text/.exec(key)) continue;
             feature.properties[key].split(',').forEach((label) => {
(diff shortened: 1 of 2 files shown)


apendleton:
I tweaked the test so that it actually triggers the crasher behavior on master (and doesn't anymore in this branch). The test feature needed an ID that resulted in the same 20-bit masked out ID component as an existing feature, so that an ID collision would be triggered and the text disambiguation strategy would be attempted. I can verify that the change does now fix the crasher, so I'm going to approve this PR so we can roll it out.

Note, however: this doesn't actually make the function behave correctly, in the sense that in the event of an ID collision, we won't actually be able to identify the correct feature in the case of intersections, because the cover text (include `+intersection`) does not actually occur in the feature text on the retrieved geojson. Given that this is a crasher, I think we should proceed anyway -- better to return nothing than crash. But I'm going to separately ticket out a rethink of this function, which seems fundamentally not to work quite the way we need it to.

Issue #823, opened 2019-02-08
Revisit how we deduplicate features


lizziegooding:
particularly addresses

### Context
While writing tests for https://github.com/mapbox/carmen/pull/822, I looked at how we current remove duplicate features before we return them to the user https://github.com/mapbox/carmen/blob/master/lib/geocoder/dedupe.js. We remove duplicate features if they have the exact same `place_name` and only return whichever feature happens to be sorted first. If one of those exact matches has its `geometry.omitted` property set to `true`, we drop that feature instead.

One example of problems with this deduping functionality is how we handle addresses with slightly different names. For example, a query for `1013 N Bonnie Brae St` in Los Angeles returns both an interpolated and a non-interpolated result, but because their place names don't exactly match (one is `1013 N Bonnie Brae St` and the other is `1013 North Bonnie Brae Street`), we return both.

![image](https://user-images.githubusercontent.com/18095166/52503642-28323f00-2bb4-11e9-92d2-3b89dff75b00.png)

### Next Steps
We should consider doing a smarter deduplication at query time that takes into account:
- if token-replaced `place_name`s are the same
- how proximate the features are; this is unlikely but it's probably possible to have two features with the exact same `place_name` in totally different parts of the world

cc @mapbox/search

miccolis:
With #846 we're put de-duplication in place that should address the issues in the OP, the query for "1013 N Bonnie Brae St Los Angeles California 90026" now returns;

![Selection_080](https://user-images.githubusercontent.com/77205/55183662-b2cffd00-5166-11e9-915d-712ef5448e08.png)

The approach take here looks at the text that was used to find the match and the elements of a result that are assembled into it's display name. This avoids attempting to redo or undo token replacement and leverages the display format to indicate what elements of a context should be used to determine uniqueness.

Closing, but @lizziegooding if I'm missing anything please reopen.

PR #846, opened 2019-03-22
De-duplicate addresses based on how they match query

miccolis:
Refs: #823 

This PR attempts to correct for the situation where data is not uniformly normalized and may contain an address like "100 Main Street" and "100 Main St"  as different features in the same index. 

At indexing & in phrasematch & spatialmatch these feature names are normalized and become identical. However at the last stage of things, in verifymatch, we use the original feature and they fail to be de-duplicated by the existing `place_name` comparison.

To account for this situation, and for addresses exclusively, this PR adds another duplication key that comprised of the text used to match the feature in spatial match and the external id of "significant" context elements. 

~Presently significant context element are ones from any index that has a zoom of greater than the address index zoom + 2. The main target of this threshold is the neighborhood index, which shouldn't prevent deduplication... It's unclear if this approach is the best. Another alternative would be to reference the format string on the index and use it as guide to what is significant.~ As may have been expected, this didn't work out.

### Summary of Changes

1. Moves existing `place_name` deduplication into `geocode.js/toFeatures`. Deletes associated files.
1. Adds address deduplication logic based on cover text

### Next Steps

- [x] Use types from the format string (if available)
- [x] Test with real data and known failure situations.

cc @mapbox/search


test/acceptance/geocode-unit.address-sort-interpolated.test.js
```diff
@@ -111,6 +111,16 @@ const { queueFeature, buildQueued } = require('../../lib/indexer/addfeature');
         });
     });
 
+    tape('De-duplicate -interpolated address', (t) => {
```
aarthykc: Do you think it is worth having this test over in `test/acceptance/geocode-unit.duplicate-address.test.js`? We have a test that shows that we prefer non-omitted address over omitted addresses and it might be good to show that we prefer a non interpolated address over an interpolated address while de-duping.

lib/geocoder/format-features.js
```diff
@@ -239,6 +261,108 @@ function toFeature(context, format, languages, languageMode, debug, geocoder, cl
...
+
+            // Dedupe by place name, prefer feature with non-omitted, non-interpolated geom
+            const keys = [feature.place_name];
+            if (context[0].properties['carmen:spatialmatch'] !== undefined && context[0].properties['carmen:address'] !== undefined) {
+                keys.push(uniqueAddressId(geocoder, types, context));
+            }
+
+            const previous = keys.reduce((m, v) => (m || by_place_name[v] || m), false);
+            if (previous) {
+                if (features[previous.index].address && !feature.address) {
```
aarthykc: per chat, if you can find notes on why this would happen or tests that capture it, that would be 👍

miccolis: @aarthykc looking at this again the comment that is already there "Don't allow street fallbacks to replace omitted or interpolated addresses" pretty much captures this. They won't be 'interpolated' or 'omitted' so w/o this explicit check to see if they represent address results they'd look like better matches.

test/acceptance/geocode-unit.address-sort-interpolated.test.js
miccolis: I'm kinda hesitant to combine these, the data being tested against is pretty different and the interpolation stuff is odd...and I can totally imagine us adding more tests here. I'd prefer to keep 'em separate for now.

Issue #853, opened 2019-04-21
Remove frequency data from indexes


miccolis:
**tl;dr** - we should remove frequency information from our indexes & plan to bring it back in the future as part of a data structure fuzzy-phrase can use.

## How do we calculate it now? What does it do?

When indexing documents we batch them into sets of 10,000 and calculate word frequency. Words that are appear a lot [insert definition] within this set of 10k documents, are deemed “low information". Features that contain that contain these “low information” words have versions of their text indexed at lower relevance with those words omitted. Assuming a uniform type of feature would work at great identifying common words (ie ‘street’, ‘way’) and indexing versions of features with those words omitted.

Unfortunately out data isn’t uniform. Our data is such that local history & geography can have a lot of influence on when a word is common or not. “Northeast”, “Northwest”, “Southeast”, “Southwest” are probably the most common words in the DC street name data set. They literally apply to every street. However, they are completely un-used across the river in Virginia. We have no control over how we batch records into 10k sets. Maybe we’ll do enough of DC in one go to indentify these as common, maybe we won’t.

Also, we don’t pay attention to language when we generate frequency information, which is just silly.

## Why eliminating it is ok?

We don’t actually use this frequency information at query time. Removing it from out indexes will have no negative affects.

## Where we’ll want to bring it back

We should use frequency information at query time and there are two lines of thinking here. Likely they’ll both prove useful and we’d implement both sequentially;

1. Given a list of low information words we could evaluate a users’s query with those words omitted. For example if “street” & “avenue” were deemed low information (very likely) we’d have a mechanism for handling when a user enters the wrong street type. To get this functionality we’d want to expose a datastructure to fuzzy-phrase that can make a determination about relevance when evaluating whether a index contains particular text.
2. With the above in place it may be possible to offload the work of dropping low information words from feature names entirely to query time. This could significantly reduce index size.

### Next Steps

* Actually do the removal. Will likely require changes in any dependent project. 

@mapbox/search

miccolis:
Done in #854

PR #854, opened 2019-04-25
Expect max score in the index definition

miccolis:
### Context

As described in #853 we package frequency info, but don't actually use it data at query time. This PR removes it and demands that the one value we do use - `maxscore` is written to the index metadata. This will be a major breaking change.

The intent is that this is a temporary situation, we aim to make frequency available again at query time and use it. This sets the stage for that work. 

A serious downside is that until we actually bring frequency information back to the index itself Carmen doesn't write a max score to the index definition, that responsibility falls to the application/user which manages index construction.

### Next Steps

- [x] Remove some code which looped over `['grid', 'freq']` arrays during index construction.
- [x] Downstream testing.

### Draft changelog entry

```
28.0.0

* No longer generate or read the frequency.rocksdb file
* Demand that the `maxscore` is in index metadata
* Remove out-of-date `bin/tokenize.js` script
```

cc @mapbox/search


test/unit/indexer/index.test.js
```diff
@@ -105,8 +105,7 @@ test('index.update -- error', (t) => {
             }
         }], { zoom: 6 }, (err) => {
             q.ifError(err);
-            q.deepEqual(conf.to._geocoder.freq.get('__COUNT__'), [2]);
-            q.deepEqual(conf.to._geocoder.freq.get('__MAX__'), [10]);
+            q.deepEqual(conf.to._geocoder.grid.list(),[['main st', [0]]]);
```
miccolis: This doesn't look right to me and I'm not quite sure what the intent of the original test was. Need another look.

apendleton: The original intent was to verify that the `__COUNT__` and `__MAX__` values were right in the freq table -- that the number of words we've seen is 2 (I think that's what `__COUNT__` was?), and the highest score is 10. What's happening now is a list in the grid table instead -- that we have one phrase, included for language set `0` (so, default text only). That's as expected from the test, but I'm not sure it has much to do with what the test was doing before.

miccolis: Right, count it the terms, not entries. Great. So, should I just :hocho: this test?

miccolis:
@apendleton I think I addressed the test side changes you requested. Let me know.

Issue #828, opened 2019-02-12
Don't use category synonyms for matching text.


apendleton:
### Context
We currently supply `matching_text` and `matching_place_name` fields on some responses, if the thing the user typed is more similar to a synonym or translation on a feature other than the one that's the default label in the user's requested language (for example, if the user types "Deutschl" with language=en, `text` might be `Germany` while `matching_text` might be `Deutschland`). If category information for points of interest is included on features as synonyms, though, this leads to weird behavior where, for example, the `matching_text` field for all results in a search for `coffee` is just `coffee`, and the `matching_place_name` field might be `coffee, washington, dc`. This is not likely to be useful to end users, so we should skip emitting matching_text/matching_place_name data in this circumstance.

Matching text is determined in the `getMatchingText` function here: https://github.com/mapbox/carmen/blob/master/lib/geocoder/ops.js#L226 and as of #758 we now have a list of all the categories in an index attached as metadata to the index's configuration. We also have a reference to the `geocoder` object from within the `getMatchingText` function. We should have all the pieces necessary to just skip attaching matching text in the event that thing we were going to attach is in the category list, fi present.

cc @lizziegooding 

### Next Steps
- [x] add new matching text functionality
- [x] tests

@mapbox/search

Nmargolis:
Done in #855

PR #855, opened 2019-04-25
No matching text categories

Nmargolis:
### No matching text for categories
The `matching_text` and `matching_place_name` properties are intended to surface feature names that most closely match the query, and are often used as the label for a feature in an api consumer's UI. Since categories are not distinguishable from synonyms, a feature that matched replace the primary feature name with the category in `matching_text` and `matching_place_name`, leading to confusing and indistinguishable results. For example a search for `coffee, washington, DC` would return `matching_text` of `coffee` and `matching_place_name` of `coffee, washington DC` for all features in the result set that matched the `coffee` category.

With this PR, if the matching text also matches a category configured on the index, the category will not replace the primary place name of a feature or it's context.

✅ **Cases that are handled:**
- If the matching_text is the same as a category specified in the index that feature is from, don't add matching_text to the feature
- If the matching_text is a translation that happens to match a category specified on the index, do add matching_text to the feature

❌ **Cases that aren't handled:**
- If a synonym of a feature happens to match a category specified on the index, the synonym will not be added to matching_text. This is because it is impossible to distinguish synonyms from categories, since they are both stored in the `carmen:text` property.

### Add to matching_place_name for all features in context with matching text
There was also a bug with matching_place_name, where matching text for the context was not getting added to the matching_place_name if a more specific part of the context, or the feature itself, had matching_text. This was [fixed in `toFeature`](https://github.com/mapbox/carmen/pull/855/commits/e43033a902a2cd38100ffa6941d9c5033c19380c#diff-a6b7f06c105beb5fb302175ee6ca3449L198) by continuing to loop through all of the features in the context even after one matching_text was found.

Issue: #828 

## Summary of Changes
- [x] Fix bug with matching_text for contexts in matching_place_name
- [x] Don't add matching_text to a feature or contexts if it appears to be a category

cc @mapbox/search

### DIFF
--- a/lib/geocoder/format-features.js
+++ b/lib/geocoder/format-features.js
@@ -195,7 +195,6 @@ function toFeature(context, format, languages, languageMode, debug, geocoder, cl
                         context[k].matching_language = matched.matching_language;
                         if (k === 0) feature.matching_text = matched.matching_text;
                         feature.matching_place_name = getPlaceName(context, formatString, language, languageMode, !!matched, geocoder);
-                        // break;
                     }
                 }
             }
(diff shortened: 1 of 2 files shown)


Nmargolis:
Good point @apendleton! We should definitely fix that somewhere upstream, not sure where. For good measure though, I added a .trim() to the matching text just in case, and updated the test case to include in a space in the category.

PR #856, opened 2019-05-02
VTQuery

ingalls:
### Context

The PR updates our vector-tile query code to use the much more modern and well tested `vtquery` library. This PR maintains parity with the mapnik code that it replaces. 2 tests were changed to get things passing, both have explanations below as to why they were changed.

### Future Steps

The use of vtquery will allow us in the near future to generate sibling tiles and query accross a 9x9 tile grid, eliminating a class of bug (usually seen on address data) where the most proximial address was just slightly off the tile being queried (on the sibling tile)

cc @mapbox/search


test/acceptance/geocode-unit.limit.test.js
```diff
@@ -406,9 +406,9 @@ tape('teardown', (t) => {
 
     tape('Reverse ITP', (t) => {
         c.geocode('-79.37745451927184,38.83420867393712', { limit: 3, types: ['address'] }, (err, res) => {
-            t.equal(res.features[0].place_name, '13 main street, west virginia');
```
ingalls: These two unit tests changed as the geometry returned by `vtquery` is more precise than that returned by mapniks implmentation.

Screenshot from 2019-05-16 22-21-29

As can be seen in the screenshot above, the `x/y hit` from mapnik was returned at the far end of the line, while vtquery returns the closest point on the line. This behavior is more desirable.

test/acceptance/geocode-unit.language-flag.test.js
```diff
@@ -106,7 +106,7 @@ const addFeature = require('../../lib/indexer/addfeature'),
             geometry: {
                 type: 'MultiPolygon',
                 coordinates: [
-                    [[[-5.625,0],[-5.625,5.615985819155337],[0,5.615985819155337],[0,0],[-5.625,0]]]
```
ingalls: This polygon was slightly modified due to vtquery being able to return multiple results for overlapping polygons if no limit is set. 

Screenshot from 2019-05-16 22-51-06

Previously on master, mapnik would simply return the first feature it found on the tile. These are only overlapping as back in the day I converted them from tile coordinates to actual geometries. They don't have to overlap nor was that the original intent.

ingalls:
@CyanRook You merged this right? Can I close and delete this branch?

CyanRook:
@ingalls Yup

Issue #858, opened 2019-05-31
Ampersand replacement to 'and' for intersections.


miccolis:
I'm attempting to replace ampersand with 'and' to use that separator for intersection search. It's not working. The replacement I'm using is;

```
"(.+) & (.+)": { "regex": true, "spanBoundaries": 1, "text": "$1 and $2" }
```

So the query comes in for `1st & main st`

```
{
    tokens: ["1st", "main", "st"],
    separators: [" & ", " ", ""],
    owner: [0, 1, 2]
}
```

We token replace the `&` and normalize that into

```
{
    tokens: ["1st", "and", "main", "st"],
    separators: [" ", " ", " ", ""],
    owner: [0, 0, 0, 2]
}
```

In phrasematch the results we generate are a little weird because of the cardinality changing replacement. We do "Gap expansion" and end up with 2 matches that differ just in the mask. Presently we generate;

```
    mask: 1, phrase: "+intersection 1st , main st",
    mask: 3, phrase: "+intersection 1st , main st"
```

This seems wrong. I'd expect the expanded mask to be `111` not `011`. Then when we calculate a weight for the match it ends up being `0.666..` and not `1`. So then we've got a bad mask that could potentially allow incorrect matches AND a weight that would sinks the single index match we're looking for.

samely:
Question @miccolis
Do we prefer street names results over intersection results? In cases where the street name contains `&` char in its name.

miccolis:
@samely I suspect the indexing process would remove `&` from street names. As we're not actually replacing ampersand yet now is the right time to figure out these sorts of implications. Do you have an example or two of a street name that has an ampersand?

samely:
Yeah...

```
40 M&W Lane, Waynesboro, Virginia 22980, United States
130 Smith & Brogdon Rd, Chase City, Virginia 23924, United States
11720 Lewis & Clark Dr, Lolo, Montana 59847, United States
2590 Coach Surrey Ln Part, Aurora, Illinois 60506, United States
```

miccolis:
At this point #859 fixes the issue here and thankfully it's a small change that is limited to intersection code. Further details on the issue are there.

@samely to give your question a more complete answer; the replacement I'm proposing that we add would normalize ` & ` to ` and `. This would work both for street names and intersections queries that match the pattern `[space]&[space]`.

[...]

PR #859, opened 2019-05-31
Properly handle ampersand replacement

miccolis:
### Context

Refs #858 Presently just a sets of test that capture where a `&` -> `and`  works fine and where it's broken.

**Update** I think I've figured out the underlying issue - `intersectionPermutations` was generating a mask that used the `owners` of the tokens, this yielded a final mask that made sense against the actual user input. That seemed great at the time, but the problem is the surrounding code expects a local mask (one that doesn't account for cardinality changing replacements), and translates what it's given it into a final mask. This PR now generates a naive local mask and the surrounding code correctly translates it.

### Next Steps

- [x] Resolve the issue in phasematch that gets us a broken mask & weight.
- [x] Downstream testing.

cc @mapbox/search

### DIFF
--- a/test/unit/geocoder/phrasematch.test.js
+++ b/test/unit/geocoder/phrasematch.test.js
@@ -644,3 +644,49 @@ tape('fuzzyMatchMulti - masks for removed terms at the end of a query', (t) => {
         t.end();
     });
 });
+
+tape('fuzzyMatchMulti - masks for intersection queries', (t) => {
+    const c = fakeCarmen({
+        fuzzyMatchMulti: (a, b, c, d) => {
+            const results = fakeFuzzyMatches(a);
+            const expected = [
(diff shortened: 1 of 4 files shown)


test/unit/geocoder/phrasematch.test.js
```diff
@@ -644,3 +644,51 @@ tape('fuzzyMatchMulti - masks for removed terms at the end of a query', (t) => {
...
+    c.geocoder_address = true;
+    c.geocoder_intersection_token = 'and';
+    c.complex_query_replacer = token.createComplexReplacer([
+        {
+            from: '(.+) & (.+)',
+            to: { regex: true, spanBoundaries: 1, text: '$1 and $2' }
+        }
+    ]);
+
+    const query = termops.tokenize('1st & main st');
```
aarthykc: Just clarifying that this is not a query type that we would support in production, but for the purpose of demonstrating `x` & `y` is converted to `x` and `y` this works fine.

PR #860, opened 2019-06-01
Integrate carmen-core and new and improved fuzzy-phrase

apendleton:
### Context
This branch drops our carmen-cache dependency in favor of [carmen-core](https://github.com/mapbox/carmen-core), and switches our fuzzy-phrase dependency to use the branch described in [this fuzzy-phrase PR](https://github.com/mapbox/fuzzy-phrase/pull/71), which changes the output format of fuzzy-phrase to conform to the input format expected by carmen-core.

Many other ancillary changes have been made to play well with these changes and in general to take this opportunity to clean things up. Details below.

**This branch targets the `carmen-core-dev` branch, not master, so we can merge it now and do more work in subsequent PRs.**

### Summary of Changes
- [x] remove carmen-cache and add carmen-core as a dependency, and switch to phrase-id-supporting fuzzy-phrase
- [x] carmen-core uses a builder pattern, like fuzzy-phrase, so there's an explicit write-only component for indexing and read-only component for query time, rather than a read-write component that can (sort of) be used without doing an explicit build; as such, now that they're the same, I standardized all the places where they get built, rebuilt, or loaded, to always do both together in the same place, where before they sometimes differed
- [x] got rid of the name `cache` to describe things that aren't caches, and got rid of the `_geocoder` name for carmen-cache instances (because wtf); the grid store is now `_gridstore` and the fuzzy-phrase set is now `_fuzzyset`.
- [x] got rid of the `cxxcache.js` and `grid.js` util files
- [x] got rid of grid encoding generally, replaced by passing in JSON formatted according to expectations of carmen-core
- [x] changed how grids are keyed at indexing time: phrases are added to fuzzy-phrase first, which then returns a temporary ID, and that's used as the grid key input; once everything is indexed, fuzzy-phrase is built, and returns a mapping of temporary to final IDs, after which the gridstore's contents are renumbered and the gridstore is finalized as well
- [x] at query time, fuzzy-phrase returns phrase IDs and phrase ID ranges, and those are used as inputs into carmen-core's coalesce (basically, as additional properties on phrasematch objects, formatted according to what's expected by neon-serde on the carmen-core side; this is a little clunky, so we could consider changing the carmen-core side to not require the input to be structured in as complicated a way as it now is, but this works... 🤷‍♂ )
- [x] Updated *lots* of tests

### Next Steps
Nothing else in this PR. In the next PR:

* actually use the `source_phrase_hash` mechanism, which isn't used yet in this PR, to make get_matching_text and text-based collision resolution work better.

cc @mapbox/search


lib/indexer/index.js
```diff
@@ -206,39 +206,25 @@ function update(source, docs, options, callback) {
...
-
-        // steps:
-        //   - pack to a temp directory
-        //   - delete current rocks cache
-        //   - move temp rocks overtop current rocks position
-        //   - create new rocks cache
-
-        const tmpdir = require('os').tmpdir() + '/temp.' + Math.random().toString(36).substr(2, 5);
+        // write grid store
+        source._gridstore.writer.renumber(idMap);
```
aarthykc: Just for my understanding - we write to both the grid and the fuzzy store with temp-ids and after the fuzzy store is built a mapping from temp-ids to permanent ids is created after which the gridstore is renumbered with the final ids of the phrases.

PR #865, opened 2019-07-02
Phrase hash

apendleton:
### Context
This PR builds on #860 to actually use the `source_phrase_hash` mechanism carmen-core introduces.

### Summary of Changes
- [x] introduce a phrase hash generation function that produces a one-byte hash extract of a given phrase. I looked at [this comparison of hashing options](https://github.com/joliss/fast-js-hash-benchmark) and specifically _"If you don't need incremental hashing, and you need more than 500 MB/sec throughput, and you're OK with a native npm dependency, use murmurhash-native for some extra performance_" and went with that recommendation, so the strategy here is to use `murmurhash-native` and take the first byte of the output
- [x] use this hasher to hash all input phrases on each document at index time. We hash phrases at the stage where the've been split apart if they're comma-spliced, but they've otherwise not been processed at all (no case normalization, no token replacement, no whitespace trimming, no tokenization, etc.) to minimize the amount of work we'll have to do at query time to replicate the hashing work. We attach this hash to all grid entries going into the store
- [x] at query time, in the getMatchingText case (for generating the `matching_text` entries on ouptut), change getMatchingText to, rather than rerunning the indexer, hash all phrases on the feature and pick the phrase where the hash matches. In the event of a hash collision, use levenshtein distance between the query phrase and the lowercased feature phrase to pick the most similar one (this introduces an additional new dependency, `leven`, which is also [very performant](https://github.com/sindresorhus/leven#benchmark)).
- [x] at query time, for the ID collision case, hash all the phrases in all the potentially-matching features (as already disambiguated by ID and score), and pick the feature that has a phrase with a matching hash. In the case of more than one feature with a phrase with a matching hash, again, disambiguate using Levenshtein distance (this is somewhat different from the above in that we're choosing amongst different features rather than choosing amongst different phrases on the same feature). Note: this function is also renamed to `matchByText`, to make its name more different from `getMatchingText`.

### Next Steps
- [ ] merge #860 and rebase this branch on top of new carmen-core-dev

Other than that, this branch should be good to go, I think. One thing I might do, though (not sure whether to add it here or do it in a new PR): in the places where I changed how we access `carmen:text` or `carmen:text_XX`, I added support for `carmen:text` being an array instead of a string combined on commas. This is a change we want to make in general, to add a way to index features with labels that contain commas, and it made sense to add support for this behavior in the places that were being rewritten anyway, but to finish the job I'll want to circle back and add tests for it, plus make additional changes to parts of the codebase that aren't touched here.

cc @mapbox/search


lib/util/feature.js
```diff
@@ -314,17 +315,47 @@ function matchScore(features, cover, source) {
...
+        let minDistance = Number.MAX_SAFE_INTEGER;
+        matchingFeatures.forEach((match) => {
+            match.distance = Number.MAX_SAFE_INTEGER;
+            for (const synonym of match.synonyms) {
+                const distance = leven(cover.text, synonym.trim().toLowerCase());
+                if (distance < match.distance) match.distance = distance;
+            }
+            if (match.distance < minDistance) minDistance = match.distance;
+        });
+        matchingFeatures = matchingFeatures.filter((f) => f.distance === minDistance);
```
Nmargolis: Is it possible for there to be more than one closest feature in the case of an id collision? If not, why not keep track of the match with the min distance as we go, instead of doing this additional filtering step?

apendleton:
@Nmargolis:

> Is spacing guaranteed to be the same at index time vs the query text at query time?

So the thing we're hashing isn't the query-time text, it's the text from the feature, both at index time and query time. At index time, we hash the (un-normalized) text to generate the hash, and store that in the grid store. At query time, if we need to select which synonym we want, we re-hash the text from the feature (again, un-normalized) and see which hash matches. If we need to decide which of several features with an ID collision is the one we want, we hash all the un-normalized text on all the features and look for a hash match. We don't have to normalize at query time because we didn't at index time, and it's the same text in all cases (the `carmen:text` or `carmen:text_XX` data from the feature).

lib/util/feature.js
apendleton: Yes, there could be more than one synonym or translation that's the same Levenshtein-distance away.

apendleton:
When I hit merge this was green, but codecov hadn't reported yet apparently, and we're a bit below the patch threshold so now (post-merge) it's ❌. I'll fix it in the next PR, #867.

PR #867, opened 2019-07-22
Prefix work and carmen-core bug fixes

aarthykc:
### Context
This PR contains the following work:

- Prefix store - As a part of https://github.com/mapbox/carmen-core/pull/41 & https://github.com/mapbox/carmen-core/issues/38, we need to be able to group all our data into precomputed buckets. With this change along with indexing all our data we create pre-computed groups of 1024 entries to enable a faster search experience during autocomplete queries.
- Fixes failing index builds because of the occurrence of the word  "constructor"  in our data. Builds failed with the following error: 
```
# build
not ok 1 TypeError: gridPatch.has is not a function
```

### Summary of Changes
- [x] Switches to a new version of carmen-core
- [x] Uses a Map instead of an object for grids while indexing

### Next Steps
- [x] Review 
- [ ] Merge into `carmen-core-dev` branch

cc @mapbox/search

### DIFF
--- a/lib/geocoder/format-features.js
+++ b/lib/geocoder/format-features.js
@@ -411,7 +411,7 @@ function getMatchingText(item, geocoder, requestedLanguage) {
         const qt = item.properties['carmen:query_text'].toLowerCase();
         const phrases = Array.from(hashMatches.keys()).map((a) => { return { key: a, dist: leven(qt, a.toLowerCase()) }; });
         phrases.sort((a, b) => a.dist - b.dist);
-        bestPhrase = phrases[0];
+        bestPhrase = phrases[0].key;
     }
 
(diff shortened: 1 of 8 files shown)


apendleton:
@aarthykc confirming per chat: you've got a test for the `constructor` thing that you're going to pull in from another branch, right?

Also: looks like https://github.com/mapbox/carmen/pull/867/commits/cc696fb99945bd45486bee1278f3bec10949dd8e deletes ~900 lines from the yarn lock file. I think maybe your version of yarn is <1.10 and doesn't include the integrity hashes? If so, might be good to update so we don't churn so much like this.

Nmargolis:
still approved 🌮

PR #871, opened 2019-08-17
Don't aggressively dedupe numerical autocomplete or short address query results

Nmargolis:
### Context
In #846, logic was added to dedupe results based on the text that a query matched in spatialmatch, combined with the geo context. This deduping logic is over-aggressive for address autocomplete queries that are very short, such as numerical-autocomplete or queries that have an address number + ~2 characters. It's possible for there to be several distinct features that match those queries that are in the same postcode, place and region.

This adds some logic to skip that extra deduping step if the address queries are particularly short.

### Summary of Changes
- [x] Adds isShortAddressQuery util to skip deduping if the query is a short address query

### Next Steps
- [x] Add some tests to confirm this behavior

cc @mapbox/search

### DIFF
--- a/test/unit/geocoder/format-features.test.js
+++ b/test/unit/geocoder/format-features.test.js
@@ -675,3 +675,70 @@ test('toFeatures - Consider full context with format and dedupe', (t) => {
     t.equal(results.features[0].place_name, 'Main Street');
     t.end();
 });
+
+
+test('toFeatures - Dont consider full context and spatialmatch text for short address queries', (t) => {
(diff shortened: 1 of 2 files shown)


test/unit/geocoder/format-features.test.js
```diff
@@ -675,3 +675,70 @@ test('toFeatures - Consider full context with format and dedupe', (t) => {
     t.equal(results.features[0].place_name, 'Main Street');
     t.end();
 });
+
+
+test('toFeatures - Dont cosider full context and spatialmatch text for short address queries', (t) => {
```
apendleton: `Dont cosider` -> `Don't consider`

PR #874, opened 2019-08-27
[wip] Tune-up interpolation range evaluation

miccolis:
### Context

While looking into an address cluster whose interpolation range was stripped, it became apparent that we're over-eager to shed interpolation data. This PR does not completely rework how we protect against ranges that would explode index size, it just targets more relevant information - the interpolation ranges themselves, not address points.

A next step here could be to only remove ranges that are problematic, rather than all or them.

It may also be possible to move this check into a different place in the indexing process so that it really is just a protection against abusive number of waffle texts entries and doesn't alter the feature we store in the index. I do worry that this would make it harder to debug these cases, so we'd want to consider that.

### Summary of Changes

- changes the outlier check to examine the interpolation ranges and not the address points. It does not change the underlying approach of using Tucky's fences and stripping all interpolation if a outlier is found. ...not yet anyhow.
- Adds a limit to the width we'd allow in a single range; 100,000. I think this is reasonable.
- Adds unit tests for outlier detection

### Next Steps

- [ ] Test on more real data
- [ ] Consider more changes

cc @mapbox/search


CyanRook:
Closed by
https://github.com/mapbox/carmen/pull/889

PR #448, opened 2016-03-09
Rebalance stacks earlier in the algorithm.

arunasank:

PR #876, opened 2019-09-04
Rebalance tweak

apendleton:
### Context
Our current rebalancing strategy (introduced by #448) assigns equal weight to all matching components, regardless of the number of words in each component, where before components where assigned weights based on the number of words they contained as a fraction of the total number of words in the query.

After having this in production for awhile, we now want to tweak this a little. For the "Martin Luther King Blvd, Seattle, Washington" example from #448, we still want Seattle, Washington to win -- if we can't actually align the feature with the city and state, we want the city and state together to win, because there are more components there. But for examples where there are actual ties (the same number of components in each possible result), #448 results in us instead disambiguating by score, and we now think we should use number of words as a (very minor) hint. So, two components should always beat one component, but if it's one component to one component, length wins.

To do this, this branch keeps component weights approximately equal, but very slightly nudges longer ones over shorter ones.

### Summary of Changes
- [x] change the rebalancing algorithm slightly

### Next Steps
- [x] update all the hard-coded relevances in tests
- [x] (maybe) change the formula for the tweaking to generate nicer numbers -- it seems that the approach so far sometimes introduces undesirable rounding errors
- [x] add an explicit test of an address beating a postcode by token code rather than the postcode winning by score

cc @mapbox/search

### DIFF
--- a/lib/geocoder/spatialmatch.js
+++ b/lib/geocoder/spatialmatch.js
@@ -502,14 +502,18 @@ function rebalance(query, stack) {
     }
 
     const garbage = (query.length === (stackMask.toString(2).split(1).length - 1)) ? 0 : 1;
-    const weightPerMatch = 1 / (garbage + stack.length);
+    const totalLengthBonus = .01 * (garbage + stack.length);
+    const weightPerMatch = (1 / (garbage + stack.length)) - 0.01;
 
     // shallow copy stack into stackClone to prevent cases where a stack's
     // index gets overwritten in deep copies.
     let totalWeight = 0;
     for (let k = 0; k < stack.length; k++) {
         stackClone[k] = stack[k].clone();
-        stackClone[k].weight = weightPerMatch * stack[k].editMultiplier;
+        stackClone[k].weight = (
+            weightPerMatch +
+            (totalLengthBonus * stack[k].weight)
+        ) * stack[k].editMultiplier;
         totalWeight += stackClone[k].weight;
     }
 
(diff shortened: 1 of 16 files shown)


lib/util/round-to.js
```diff
@@ -0,0 +1,6 @@
+'use strict';
+
```
aarthykc: Should we add a test for this function?

lib/geocoder/spatialmatch.js
```diff
@@ -502,18 +503,22 @@ function rebalance(query, stack) {
...
     // shallow copy stack into stackClone to prevent cases where a stack's
     // index gets overwritten in deep copies.
     let totalWeight = 0;
     for (let k = 0; k < stack.length; k++) {
         stackClone[k] = stack[k].clone();
-        stackClone[k].weight = weightPerMatch * stack[k].editMultiplier;
+        stackClone[k].weight = roundTo((
+            weightPerMatch +
+            (totalLengthBonus * stack[k].weight)
+        ) * stack[k].editMultiplier, 8);
```
aarthykc: Was there a reason why we went with 8 places for roundTo()?

apendleton: I wanted something that would eliminate any rounding errors that were farther out in decimal places, and also make the numbers sent to users look a little less crazy, since they had gone from being, like, .5, to .99999999999... . I started with .6, but it turns out if you round here to .6, and more than one subcomponent is rounded up, you can end up with situations where they sum to a value greater than 1 (1.000001 or whatever). So: I did 8 here, and then 6 later for the final output relevance, so that any errors introduced by this rounding (like, 1.00000001) would get rounded away again in the 6-place rounding later. Fiddling around with it a bit, this combo seemed to produce the best results.

lib/util/round-to.js
apendleton: Will do 👍

lib/geocoder/spatialmatch.js
aarthykc: I see, makes sense 👍

lib/util/round-to.js
apendleton: Added.

Issue #847, opened 2019-03-22
Fix feature collision avoidance


apendleton:
### Context
The coalesce process returns a record for each feature we might want to return as part of a result, consisting of a tuple `(relev, score, x, y, id)`, where the ID is the 20 least-significant bits of the actual original feature's ID. We then attempt to retrieve the feature using the `(x, y, id)` component, but sometimes more than one result comes back, because more than one feature on the same tile (the same `(x, y)`) has the same 20 least-significant bits in its ID.

In that case, we first try to disambiguate using the score, by recomputing the 3-bit score value for each feature. If they're *still* tied, since we also know which phrase produced each match, we look at all the labels on the feature (all the `/^carmen:text.*/` properties, break out their synonyms, and see if any of them match the phrase text we got back.

... except this seems to not actually work. #845 made it not crash (under some circumstances it was trying to generate an invalid regular expression), but it will still fail to find the feature we actually want lots (or maybe most) of the time, because it just looks for an occurrence of the cover text inside one of the feature labels. Problems:
* it's case-sensitive (both was before #845 and still is), and all of the phrase text will have been lowercased while in all likelihood, the actual feature text won't have been; the unit test creates a synthetic circumstance under which the case is mixed on both sides, but this won't occur in practice
* it doesn't account for token replacement, which will have been applied to the cover text but not the feature text
* it doesn't account for other transformations that might have occurred to the cover text but not the feature text, such as `###`-ifying of address numbers and, as of recently, rewriting of cover text for intersection labels.

Options:
* we could effectively re-run the indexing process on a given feature to regenerate its phrases, the same way we do to identify `matching_text` -- we might even be able to re-use the same code (or maybe even only run the operation once, and use its results for both purposes?). This process works most of the time, but not every time for reasons we haven't really dug into much.
* we could use a Levenshtein or other fuzzy comparison to see which text is closest to the cover text, with the expectation that this will still give us the right feature despite possible transforms
* both, maybe, if neither is reliable enough alone?

Bottom line, it probably shouldn't be possible for us to retrieve multiple features and then choose to use none of them, as seems to be the case now. We should always pick one, even if our confidence that it's the right one is tenuous.

Separately, whenever we next rework the grid format, we should consider moving to 32-bit IDs to make collisions less likely.

### Next Steps

TBD

@mapbox/search

PR #877, opened 2019-09-04
Use Levenshtein distance for coarse resolution of ID collisions

apendleton:
### Context
#845 touched the `matchText` function in `lib/util/feature.js`, which tries to resolve ID collisions in fetch results from the feature table. That function was crashing, and we made it not crash, but it still didn't work; see #847 for more info. The plan was to fix this behavior with a series of changes in carmen-core and its equivalent carmen branch that would make ID collisions both less likely and easier to disambiguate, but that work has stalled due to unrelated carmen-core performance concerns.

This branch is a band-aid in the meantime. As noted in #847, the previous code almost never worked because it was broken with respect to case (input data was usually mixed case but equality tests were against lowercase cover text), and also didn't handle things like token replacement or other phrase normalization correctly. This branch fixes the case issue, and also uses Levenshtein distance rather than exact text to decide which feature to pick, comparing the cover text with the case-normalized phrase text.

Because we don't have much information to go on here (as compared to in carmen-core), we're conservative: we'll only consider matches with a Levenshtein ratio (`(lensum - ldist) / lensum`) of at least .75. This means the strings will still have to be pretty similar to match: aggressive token replacements on short strings still probably won't work, intersections probably won't work, and address numbers combined with short street names might not work either. Still, in no circumstance are the results worse than they are now, and in many cases they're better. Carmen-core, if/when that ships, will be more reliable.

### Summary of Changes
- [x] tweak the `matchText` function to be less broken

### Next Steps
Nope

cc @mapbox/search

### DIFF
--- a/lib/util/feature.js
+++ b/lib/util/feature.js
@@ -1,6 +1,7 @@
 'use strict';
 const queue = require('d3-queue').queue;
 const termops = require('../text-processing/termops');
+const leven = require('leven');
 
 module.exports = {};
 module.exports.normalize = normalize;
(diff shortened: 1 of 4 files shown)


lib/util/feature.js
```diff
@@ -315,16 +316,27 @@ function matchScore(features, cover, source) {
...
-        let matches = false;
+    const featuresWithRatio = [];
+    for (const feature of features) {
+        let bestRatio = 0;
         for (const key of Object.keys(feature.properties)) {
             if (!feature.properties[key] || !/^carmen:text/.exec(key)) continue;
             feature.properties[key].split(',').forEach((label) => {
-                if (label.indexOf(cover.text) !== -1) matches = true;
+                label = label.trim().toLowerCase();
+                const dist = leven(cover.text, label);
```
Nmargolis: Is there a possibility where the label is "United States of America" and the cover.text is "United States" that this would exclude as a potential match? It seems like before, the idea was if the cover text was found anywhere in the label, that was considered a match, whereas this seems to be looking for similarity between entire strings.

apendleton: This is possible, yes, but we check each phrase on the feature, so if the feature had both "United States" and "United States of America" as synonyms, we'd check them both separately, and somehow, in order for us to even get here, the phrase "united states" would have to have made it into the index in the first place, which would pretty much have to mean it'd have to have been present as a synonym.

The other possibility, I suppose, would be that it could have gotten there by having words dropped due to frequency. But, if that was the goal, the old way was a pretty unreliable way to do it, because it would rely on the words being dropped having been at the beginning or the end rather than in the middle (which is equally likely). I'm not aware of a "Levenshtein-and-also-contains" measure, so I think if we're picking between the two, Levenshtein is much better, especially because dropped words *are* expressable as a Levenshtein distance (you can model them as having deleted each individual letter in the now-missing word).

apendleton: Also, as per the original PR text, the old one just didn't work at all, regardless of its intent, because it didn't normalize for case, so I'm not particularly worried about regressions here.

PR #881, opened 2019-09-25
Reduce verifymatch limits by half for stopgap performance increase

Nmargolis:
### Context
The impact of the increase of VERIFYMATCH_STACK_LIMIT from 20 to 100, and the context backfill resulting in loading up to 40 contexts when previously there was a max of 10, was more severe for forward geocodes than anticipated. This PR reduces those limits by half until a more optimal solution can be reached.

### Summary of Changes
- [x] Decrease `VERIFYMATCH_STACK_LIMIT` to 50
- [x] Decrease `MAX_CONTEXTS_LIMIT` to 20

cc @mapbox/search


aarthykc:
🌮

PR #886, opened 2019-10-07
Drop frequent words while indexing

aarthykc:
### Context
From a recent address triage sprint we found out that 7% of the failures was because of `missing_input` which meant that the user didn't add information like the cardinal direction such as `NW`, `SW` or they left out the ordinal such as `3` instead of `3rd` or they didn't add the way type such as `street` or `Avenue`. 

Currently the way we identify and drop frequent words like the `way type` or the `cardinal directions` is by using a version of the [tf-idf](https://en.wikipedia.org/wiki/Tf%E2%80%93idf) logic. Given that we only index a few 1000 documents at a time, this method is not always great for identifying common words.

In the new indexing strategy the idea is that we're going to provide carmen with a list of frequent words which we will use while indexing. The idea is to use the list of frequent words, dropping them while indexing and bumping the relevance. For example:

Consider the following feature: 
```
const address = {
            id:1,
            properties: {
                'carmen:text':'North Capitol Street',
                'carmen:center':[0,0],
            },
            geometry: {
                type: 'Point',
                coordinates: [0,0]
            }
        };
```
_North and Street here are cardinals and way types respectively_ 

If we index this feature without `North` or `Street` and bump the relevance of the phrases `North Capitol` and `Capitol Street` to `0.8` while indexing - we should be able to surface the feature even if the user does not type `North` or `Street`.

### Summary of Changes
- [x] Changes index.js to take add frequent word list on the source 
- [x] Changes to getIndexablePhrases to index words without frequent words from the list
- [x] tests

### Next Steps
- [x] Review 
- [ ] merge 

cc @mapbox/search

### DIFF
--- a/lib/text-processing/termops.js
+++ b/lib/text-processing/termops.js
@@ -716,9 +716,9 @@ function getIndexablePhrases(text, freq, frequentWords) {
     for (let i = 0; i < perms.length; i++) {
         // Indexing optimization.
         const relev = perms[i].relev;
-        const text = perms[i].join(' ');
-
         if (relev < 0.8) break;
+
+        const text = perms[i].join(' ');
         const etext = normalizeText(text);
 
         // Encode canonical phrase.
--- a/lib/text-processing/termops.js
+++ b/lib/text-processing/termops.js
@@ -759,7 +759,6 @@ function getWeights(tokens, freq, frequentWords) {
     let weightsum = 0;
     const weights = [];
     const totalfreq = freq['__COUNT__'][0] || 1;
-    const regex = frequentWords ?  new RegExp(frequentWords.join('|').toLowerCase()) : false;
     let numTokens = false;
 
     // Determine weights of all terms relative to one another.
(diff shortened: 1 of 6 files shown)


apendleton:
Capturing from voice: what we want the spec to be here is, for any phrase that contains any number of words in the frequent words list, for every permutation we generate for which exactly one frequent word has been dropped (so, it's identical to the original phrase except that it's missing that one word), we should guarantee that it has a relevance of 0.8.

lib/text-processing/termops.js
```diff
@@ -752,12 +752,14 @@ function sortByRelev(a, b) {
...
  */
-function getWeights(tokens, freq) {
+function getWeights(tokens, freq, frequentWords) {
     let i = 0;
     let termfreq = 0;
+    let termWeight = 0;
     let weightsum = 0;
     const weights = [];
     const totalfreq = freq['__COUNT__'][0] || 1;
+    const regex = frequentWords ?  new RegExp(frequentWords.join('|').toLowerCase()) : false;
```
apendleton: In general, because of things like https://github.com/mapbox/carmen/pull/845 , I think we should get out of the habit of assuming we can safely take arbitrary strings from who knows where and make regexes out of them (what if they have dots or pluses or parentheses?). Instead of lowercasing all the words and putting them in a regex, can we lowercase them all and put them in a `Set`? and then use `set.has()` instead of `regex.test()` further down.

apendleton: Now that I think about it, we can probably construct that set in the Carmen constructor right from the beginning, rather than doing it every time getWeights is called, and just pass a `Set` around instead of an array.

lib/text-processing/termops.js
```diff
@@ -766,13 +768,18 @@ function getWeights(tokens, freq) {
         if (/#/.test(tokens[i])) {
             numTokens = true;
             weights[i] = -1;
+        } else if (regex) {
```
apendleton: So if there's a list of frequent words at all on this whole index, we go down this path instead of the `else` path, right? Doesn't that mean this changes how frequency is calculated for every phrase in the index, even if this particular phrase doesn't have any frequent words in it? The math looks pretty different here than in the other branch, and it seems like it would make the old kind of frequency analysis not work, unless I'm missing something?

It seems like what we want is just to override the weight on specific words if they match the regex, and leave the calculation for the rest of the words exactly the same as it was before, no?

lib/text-processing/termops.js
aarthykc: yeah that works!

PR #888, opened 2019-10-09
Feature backfill

Nmargolis:
### Context
The previous verifymatch stack limit increase and context backfill PR #879 allows for better results, but also is a significant performance decrease. Instead of reading data for 50 features at the beginning of verifymatch, this adds a similar backfill mechanism for feature loading, where additional features (on top of the initial 20) are only loaded if there aren't a full 20 verified features after verifying the first batch. Features are then loaded as needed, until there are either 20 verified features, or there have been a total of 50 loaded.

### Summary of Changes
- [x] Adds feature backfill mechanism in verifymatch
- [x] Updates the spatialmatch stack limit constant to 40, the number most often actually passed in as an option
- [x] Adds a heuristic for stopping the feature backfill process if there is a reduction in spatialmatch relevance (see #887 )

### Next Steps
- [x] Clean up some stray todos
- [x] Make it clearer how and when the partialNumberLimit is applied

cc @mapbox/search

### DIFF
--- a/lib/geocoder/verifymatch.js
+++ b/lib/geocoder/verifymatch.js
@@ -60,14 +60,15 @@ function verifymatch(query, stats, geocoder, matched, options, callback) {
         const { spatialmatches, verified, batchSize, partialNumberCount, matchesSeen, startPos } = featureOpts;
 
         let spatialmatchesChunk = [];
+        let stopEarly = false;
         const backfill = [];
+        let batchPartialNumberCount = 0;
         // Limit initial feature check.
         if (spatialmatches.length > batchSize) {
             // Set the partialNumber limit to be up to 80% of the total stack limit,
             // depending on how many partial number results there already are
             const partialNumberLimit = (0.8 * options.verifymatch_stack_limit) - partialNumberCount;
 
-            let batchPartialNumberCount = 0;
             for (let i = 0; i < spatialmatches.length; i++) {
                 // Skip covers if their source isn't allowed by types & stacks filters.
                 if (options.types || options.stacks) {
@@ -77,13 +78,19 @@ function verifymatch(query, stats, geocoder, matched, options, callback) {
                     if (!source) console.error(new Error('Misuse: source undefined for idx ' + firstCover.idx));
                     if (!source || !filter.sourceAllowed(source, options)) continue;
                 }
+
+                if (verified.length && spatialmatches[i].relev < verified[0].properties['carmen:spatialmatch'].relev) {
+                    stopEarly = true;
+                    break;
+                }
+
                 // Enforce partialnumber limit
                 if (spatialmatches[i].partialNumber) {
-                    batchPartialNumberCount++;
                     if (batchPartialNumberCount > partialNumberLimit) {
                         backfill.push(spatialmatches[i]);
                         continue;
                     }
+                    batchPartialNumberCount++;
                     spatialmatchesChunk.push(spatialmatches[i]);
                 }
                 else {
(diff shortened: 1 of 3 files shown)


lib/geocoder/verifymatch.js
```diff
@@ -21,7 +20,7 @@ module.exports.sortContext = sortContext;
 
 /**
 * verifymatch - results from spatialmatch are now verified by querying real geometries in vector tiles
-*
+*q
```
apendleton: ?

lib/geocoder/verifymatch.js
```diff
@@ -31,39 +30,115 @@ module.exports.sortContext = sortContext;
...
+            results:  [],
+            goodFeatureCount: 0,
+            batch: verified.slice(0, options.limit_verify),
+            backfill: verified.slice(options.limit_verify)
+        };
+        verifyContextChunk(geocoder, verifiedContexts, sets, options, callback);
+    });
+
+
+    function verifyFeatureChunk(query, geocoder, featureOpts, options, done) {
```
apendleton: Is there a reason this is defined nested inside the parent function? I guess we're probably closing on some shared state?

lib/geocoder/verifymatch.js
```diff
@@ -361,6 +434,7 @@ function verifyFeatures(query, geocoder, spatialmatches, loaded, options) {
             byText[text] = true;
         }
     }
+    // TODO: or maybe this is where partialnumbercount needs to be calculated
```
apendleton: Are going to do this?

lib/geocoder/verifymatch.js
```diff
@@ -31,39 +30,115 @@ module.exports.sortContext = sortContext;
...
+            if (err) return callback(err);
+            const verifiedChunk = afterFeatures(loaded, options, query, geocoder, spatialmatchesChunk, startPos);
+            verified.push(...verifiedChunk);
+            // Base cases
+            if (stopEarly ||
+                backfill.length === 0 ||
+                verified.length >= options.verifymatch_stack_limit ||
+                matchesSeen + loaded.length >= VERIFYMATCH_MAX_FEATURES_LIMIT
+            ) {
+                return done(null, verified);
```
apendleton: Nit, but I wonder if, since there are so many nested things at this point, we should get away from `callback` and `done` and name them something that expresses which thing we're finishing. I had to hunt around to figure out which layer of nesting this gets us out of.

lib/geocoder/verifymatch.js
```diff
@@ -367,39 +530,34 @@ function verifyFeatures(query, geocoder, spatialmatches, loaded, options) {
...
+ * loadContext returns the heirachy of the a that makes up each result's
+ * geographic context, as provided by {@link context}
+ *
+ * @param {Object} geocoder - a geocoder datasource
+ * @param {Object} feature - geojson feature
+ * @param {Object} sets - covers for the features that have been verified
+ * @param {Object} options -  passed through the geocode function in geocode.js
+ * @param {Function} afterContextChunk - callback called after all features are loaded
+ * with the features in the the appropriate heirachical order
+ * @rewturns {undefined}
```
Nmargolis: oops, will fix typo

lib/geocoder/verifymatch.js
Nmargolis: Yes, it requires access to the callback and other shared variables in scope, and calls other functions that require those variables

lib/geocoder/verifymatch.js
Nmargolis: nope, removed

lib/geocoder/verifymatch.js
Nmargolis: I have renamed a bunch of the callbacks (and created named callback functions). Many of these are defined within other functions as well so they can have access to the arguments passed to the original functions, in addition to the return values

lib/geocoder/verifymatch.js
```diff
@@ -31,70 +30,95 @@ module.exports.sortContext = sortContext;
 * @param {Function} callback - callback function which is called with the verified indexes in the correct hierarchical order
 */
 function verifymatch(query, stats, geocoder, matched, options, callback) {
-    let spatialmatches = matched.results;
+    const spatialmatches = matched.results;
     const sets = matched.sets;
-
     options.limit_verify = options.limit_verify || 10;
-    // Limit covers based on whether their source is allowed by types & stacks filters.
-    if (options.types || options.stacks) {
```
Nmargolis: This has been moved into the getSpatialmatchesChunk function

Nmargolis:
For posterity, here's a flow chart of how this approach works:
Verifymatch-backfill-flow-chart

PR #889, opened 2019-10-14
Address Outlier Improvement

CyanRook:
### Context
Currently, if an outlier in the address data is detected, all the interpolation ranges are nulled. Now, we will just truncate the interpolation ranges to where the non outlier data is located. This came up as an issue where we were being overly aggressive in truncating interpolation ranges.

#### Why do we do this?
Occasionally data will include addresses that are outside the normal address range for a street.  This can cause the interpolation ranges to be far larger than necessary.  This follows down to the indexing stage and suddenly the indexer has a large increase in the number of entries that don't represent the actual data.  For example, if the addressnumbers are [1, 5, 10, 5000], 5000 is clearly way outside the norm for these addresses.

#### Previous Approach
The previous approach had an outlier detection scheme that looked at the first and third quartile and then extrapolated from there to look for outliers.  If it found outliers, it eliminated all the interpolation ranges for that feature. This was a bit too aggressive in both eliminating good interpolation ranges and detected outliers when there were not any due to non uniform density of address spacing.

#### New Approach
This is a modified version of the previous approach.  Instead of looking at the first and third quartile, the detection algorithm looks at the the 10th and 90th percentile and then extrapolates from there.  It also carries back the percentiles and limits from these calculations and uses this data to limit interpolation ranges instead of eliminating them.

### Summary of Changes
- [x] Change outlier detection to return bounds on non-outlier data
- [x] Adjust outlier detection constants
- [x] Change interpolation ranges to match outlier bounds
- [x] Write unit tests

### Next Steps

cc @mapbox/search


lib/indexer/indexdocs.js
```diff
@@ -539,38 +553,42 @@ function generateFrequency(docs, simpleReplacer, complexReplacer, globalTokens,
...
-                return true;
-            }
-            else return false;
+            // Calculate the limits used for detecting outliers
+            lowerLimit = lowerPercentile - (deltaPerPercentile * (targetPercentile + errorPercentile));
+            upperLimit = upperPercentile + (deltaPerPercentile * (targetPercentile + errorPercentile));
+
+            const lowerBound = sortedArray.find((value) => value >= lowerLimit);
+            const upperBound = sortedArray.slice().reverse().find((value) => value <= upperLimit);
+            const outlierCont = sortedArray.filter((value) => (value < lowerLimit || value > upperLimit)).length;
```
aarthykc: Nitpick: outlierCount?

CyanRook: 👍 Also made some of the variables more clearly named and used destructuring the output to named variables instead of indexing the array output.

PR #895, opened 2019-10-21
Improve results for omitted interpolation features in proximity search

apendleton:
### Context
#200 introduced a mechanism for dealing with situations where a sought address number has neither an exact address point match in a given cluster, not an interpolation match, but is close to the ends of the interpolation ranges in question. In this instance, we allow for reading past the end of the interpolation range to approximate a result. These features have the `omitted` property set to true on them, and are down-ranked as compared to other address results of similar relevance.

For proximity results specifically, this can mean very-far-away non-omitted results can beat nearby omitted results, which leads to a poor user experience. This PR changes that behavior, so that for proximity features, results within the proximity radius have their scoredist (which factors in distance) considered before their omitted-ness. Features outside the proximity radius are unchanged, and non-proximity searches are also unchanged (that is to say, absent any information about the user's location, non-omitted results are still preferred over omitted results, all else being equal. 

### Summary of Changes
- [x] for proximity results, we calculate a composite relevance/scoredist value so that scoredist can beat relevance in some circumstances; this composite value no longer factors in omitted-ness
- [x] a new inside_radius temporary property is calculated on features to aid in sorting
- [x] for situations where features are being sorted against one another, composite relevance/scoredist is tied, and inside_radius is different (meaning one feature is far away and one is close), scoredist is considered before omitted-ness. In practice this will likely occur only for far-away features, as for close features, scoredists will typically differ
- [x] added a test

### Next Steps
Nope

cc @mapbox/search


lib/geocoder/verifymatch.js
```diff
@@ -749,12 +750,18 @@ function sortContext(a, b) {
     if (ar > br) return -1;
     if (ar < br) return 1;
 
+    // omitted difference
+    const omitted = ((a[0].geometry && a[0].geometry.omitted ? 1 : 0) - (b[0].geometry && b[0].geometry.omitted ? 1 : 0));
+    if (omitted !== 0 && a[0].properties['carmen:inside_radius'] === b[0].properties['carmen:inside_radius']) return omitted;
```
apendleton: We consider omitted before scoredist (the current-master behavior) if both features being compared are inside the radius, both are outside the radius, or inside_radius is null for both features (as will be the case for non-proximity searches)

lib/geocoder/verifymatch.js
```diff
@@ -749,12 +750,18 @@ function sortContext(a, b) {
...
+
     // sort by scoredist
     const as = a[0].properties['carmen:scoredist'] || 0;
     const bs = b[0].properties['carmen:scoredist'] || 0;
     if (as > bs) return -1;
     if (as < bs) return 1;
 
+    if (omitted !== 0) return omitted;
```
apendleton: If we've gotten this far, we didn't consider omitted last time, so we'll consider it now, after scoredist; this will be the case for proximity searches where we're comparing a nearby feature with a far-away feature.

lib/geocoder/verifymatch.js
```diff
@@ -307,6 +307,9 @@ function verifyFeatures(query, geocoder, spatialmatches, loaded, options) {
 
             feat.properties['carmen:distance'] = feat.properties['carmen:distance'] ||
                 proximity.distance(options.proximity, feat.properties['carmen:center'], cover);
+            feat.properties['carmen:inside_radius'] = options.proximity ?
```
CyanRook: nit: If splitting across lines, can we put the operator at the beginning so it is easier to read as an extension of the previous line?
```js
feat.properties['carmen:inside_radius'] = options.proximity
    ? feat.properties['carmen:distance'] < proximity.scaleRadius(feat.properties['carmen:zoom'])
    : null;
```

lib/geocoder/verifymatch.js
```diff
@@ -749,12 +750,18 @@ function sortContext(a, b) {
     if (ar > br) return -1;
     if (ar < br) return 1;
 
+    // omitted difference
```
CyanRook: Nit: Drop the comment and just label the variable `omittedDifference`

lib/geocoder/verifymatch.js
apendleton: We don't do this anywhere else in Carmen: https://github.com/mapbox/carmen/blob/master/lib/geocoder/context.js#L603 , https://github.com/mapbox/carmen/blob/master/lib/geocoder/format-features.js#L181-L183 , https://github.com/mapbox/carmen/blob/master/lib/geocoder/phrasematch.js#L40-L42 etc. I think you could argue either way -- that it's more important to signal that a line is an extension of the previous line, or that it's more important to signal that a given line continues onto the next line, but either way, I'm not keen on starting a new trend here; we seem to have settled on this style.

lib/geocoder/verifymatch.js
```diff
@@ -307,6 +307,9 @@ function verifyFeatures(query, geocoder, spatialmatches, loaded, options) {
 
             feat.properties['carmen:distance'] = feat.properties['carmen:distance'] ||
                 proximity.distance(options.proximity, feat.properties['carmen:center'], cover);
+            feat.properties['carmen:inside_radius'] = options.proximity ?
+                feat.properties['carmen:distance'] < proximity.scaleRadius(feat.properties['carmen:zoom']) :
```
Nmargolis: Documenting from chat: the carmen:zoom property may not be set yet here, so the scaleRadius function will use the default constant for z14. Could use cover.zoom instead since that is defined at this point.

PR #902, opened 2019-10-28
Partial credit for bad alignment

apendleton:
### Context
We sometimes encounter situations where a user queries for an address that assumes an alignment of features that doesn't exist in our data (e.g., an address in a postcode when in reality, while we have that address in our data, we don't think it's in that postcode). This might happen because our data is wrong, or because a colloquial understanding of a city might exist that doesn't reflect its legal/administrative boundaries. In current Carmen, we treat all misalignments as equally bad: they contribute nothing to the total relevance of a result. This is true regardless of if the features we tried to align are right next to each other or across the country from each other. Often, the close-but-not-quite result is actually much higher-signal than other kinds of misalignemtns, and while we want to prefer actual alignments over these kinds of barely-misaligned results, the barely-off ones are better than ones that aren't close at all.

This branch creates such a partial credit mechanism, specifically for cases where a result is surfaced by `coalesce`, meaning it tile-aligned, but fails in context retrieval, meaning it didn't exactly align. We now detect these cases and assign them partial credit.

Approach:
* we iterate through elements of the context in context verification, and if we encounter an item that wasn't in the user's query, rather than just skipping it altogether, we check if there was a different item in the user's query as surfaced by coalesce from the same index
* if there was, we book-keep this almost-matched feature in parallel to the regular info about the match: it has its own mask, and its own relevance. This is so that if other elements of the context *do* match correctly and claim the same tokens as our close-but-not-quite match (in other words, their masks overlap), the actual match gets considered and applied first.
* after iterating through the context, if our close-but-not-quite mask doesn't overlap with our actual mask, we add the close-but-not-quite relevance to the total relevance with a penalty

### Summary of Changes
- [x] add new logic to `verifyContext` as described above

### Next Steps
- [x] still needs more tests

cc @mapbox/search


apendleton:
Slight update: I tweaked the logic a little to make it stricter. Specifically, low-zoom features (basically, countries or regions) are no longer eligible, so you have to have at least gotten the country and state right in your query. Also, the text we're giving partial credit for now has a minimum length (must be more than three characters) so that we can be reasonably confident you actually meant the thing and it's not just noise -- turns out a lot of features are next to something that starts with "san".

Those changes don't sacrifice any actual wins from the earlier logic, but reduce regressions and jitter a lot.

PR #908, opened 2019-11-05
VT Query Direct Hit Polygon

CyanRook:
### Context
When upgrading from mapnik to VTQuery, it was unrealized at the time that mapnik did not support distance to polygons while VTQuery did.  This caused reverse geocode results return polygons that the point was not in.  This change returns to the original functionality which requires reverse geocoding points to be within the polygons it returns.  This does not affect behavior around point or line geometries.

### Summary of Changes
* Update VTQuery
* Enable direct hit flag in VT Query
* Only direct hits on polygons will match
* Add .idea to .gitignore

### Next Steps

cc @mapbox/search

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,4 @@ npm-debug.log
 /.nyc_output/
 coverage.lcov
 yarn-error.log
+.idea/
--- a/lib/geocoder/context.js
+++ b/lib/geocoder/context.js
@@ -343,7 +343,8 @@ function nearestPoints(source, lon, lat, scoreFilter, callback) {
             radius: 1000,
             layer: source.geocoder_layer,
             limit: 5,
-            dedupe: true
+            dedupe: true,
+            direct_hit_polygon: true
         }, afterQuery);
     }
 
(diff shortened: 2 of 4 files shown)


.gitignore
```diff
@@ -10,3 +10,4 @@ npm-debug.log
 /.nyc_output/
 coverage.lcov
 yarn-error.log
+.idea/
```
aarthykc: Was the addition of this line intentional?

lib/geocoder/context.js
```diff
@@ -548,7 +549,8 @@ function contextVector(source, lon, lat, full, matched, language, scoreFilter, r
                 layer: source.geocoder_layer,
                 limit: 5,
                 dedupe: true,
-                'basic-filters': ['all', [['carmen:score', '>=', 0]]]
+                'basic-filters': ['all', [['carmen:score', '>=', 0]]],
```
aarthykc: Not a blocker for this release, but consistency around property name casing would be great!

.gitignore
CyanRook: It's not critical for this release, but I've been meaning to put it in for a while.  The .idea directory gets generated by the Jetbrains IDEs.  Adding it to the .gitignore also prevents that directory from being bundled up by `npm publish`.

PR #911, opened 2019-11-15
Scoredist decode hack

Nmargolis:
### Context
Scores are encoded on a log scale between 1 and 7 for storage purposes. The scoredists coming out of carmen-cache are based on this log-encoded score. In spatialmatch, there is an attempt to "decode" the scoredist by accounting for the scorefactor of the data source. This isn't an exactly inverse mathmatical operation because the scoredist can factor in distance, in addition to the score.

Scoredists coming out of carmen-cache were getting "decoded" to a value that was always equal to 1 if the scorefactor of an index was 1. For example, matches with a scoredist of 1 coming out of coalesce and matches with a scoredist of 6 would both get "decoded" to a scoredist of 1 if the scorefactor of an index is 1.

This PR adds a small boost to the scorefactor to avoid losing the differentiation between relatively low scoredists. A value of 1.01 was somewhat arbitrarily chosen, but it intersects nicely with the values that the other equation for expanding scoredists (if the scoredist is greater than 7) produces. That way a scoredist that's less than 7 coming out of coalesce should not ever result in a higher decoded scoredist than one that's greater than 7 coming out of coalesce.

Screen Shot 2019-11-15 at 1 54 24 PM

Since the original reason for applying a different operation to scoredists that are greater than 7 vs scoredists that are less than 7 is unknown, this PR leaves that intact.

A more robust solution would probably be to decode the scores before calculating the scoredist in carmen-cache in the future.

### Summary of Changes
- [ ] Adds a small boost to the scorefactor when "decoding" scoredists
- [ ] Adds an argument to decode3BitLogScale for whether or not to round to the nearest integer, since other operations rely on the output of that function being an integer.

cc @mapbox/search

### DIFF
--- a/lib/geocoder/spatialmatch.js
+++ b/lib/geocoder/spatialmatch.js
@@ -574,7 +574,7 @@ function Cover(cacheCover, phrasematch) {
     this.tmpid = cacheCover.tmpid;
     this.distance = cacheCover.distance;
     this.score = termops.decode3BitLogScale(cacheCover.score, phrasematch.scorefactor);
-    this.scoredist = cacheCover.scoredist > 7 ? (phrasematch.scorefactor / 7) * cacheCover.scoredist : termops.decode3BitLogScale(cacheCover.scoredist, phrasematch.scorefactor);
+    this.scoredist = cacheCover.scoredist > 7 ? (phrasematch.scorefactor / 7) * cacheCover.scoredist : termops.decode3BitLogScale(cacheCover.scoredist, Math.min(phrasematch.scorefactor, 1.01));
     this.scorefactor = phrasematch.scorefactor;
     this.matches_language = cacheCover.matches_language;
     this.prefix = phrasematch.prefix;
--- a/lib/geocoder/spatialmatch.js
+++ b/lib/geocoder/spatialmatch.js
@@ -573,8 +573,8 @@ function Cover(cacheCover, phrasematch) {
     this.idx = cacheCover.idx;
     this.tmpid = cacheCover.tmpid;
     this.distance = cacheCover.distance;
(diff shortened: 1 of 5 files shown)


lib/geocoder/spatialmatch.js
```diff
@@ -574,7 +574,7 @@ function Cover(cacheCover, phrasematch) {
     this.tmpid = cacheCover.tmpid;
     this.distance = cacheCover.distance;
     this.score = termops.decode3BitLogScale(cacheCover.score, phrasematch.scorefactor);
-    this.scoredist = cacheCover.scoredist > 7 ? (phrasematch.scorefactor / 7) * cacheCover.scoredist : termops.decode3BitLogScale(cacheCover.scoredist, phrasematch.scorefactor);
+    this.scoredist = cacheCover.scoredist > 7 ? (phrasematch.scorefactor / 7) * cacheCover.scoredist : termops.decode3BitLogScale(cacheCover.scoredist, Math.max(phrasematch.scorefactor, 1.01), 3);
```
CyanRook: Why 1.01 instead of 1.0?

CyanRook: For posterity, this is to ensure the value is strictly above 1.0 except when scorefactor is 0.

Nmargolis: Correct, it's to ensure that there is differentiation in scoredists even when the scorefactor is 1, because 1 to the power of any number (which is what the decode3BitLogScale function does) will always return 1. The actual number doesn't matter that much (it could've been .99) but as I'll note in the PR description, 1.01 has a nice intersect with the other way of expanding out the scoredist if the scoredist is greater than 7.

PR #912, opened 2019-11-20
Handle missing ordinals 

aarthykc:
### Context
A recent address triage process revealed that ~7% of the failures came from details missing data in input which means that the query left out details like the cardinal directions NW, SW), or way types (street, road etc) or ordinals. Out of the 7% of failures where the user failed to include details in the query, users most frequently left out ordinals like `th` in `7th street northwest`. A solution to this problem was to index both forms of the feature, one with the ordinal and one without at a reduced relevance. For example, say we have the following feature:
```
const address = {
            id:1,
            properties: {
                'carmen:text':'4th Street Northwest',
                'carmen:center':[0,0],
            },
            geometry: {
                type: 'Point',
                coordinates: [0,0]
            }
        };
```
The idea here is to index both `4th Street Northwest` as well as `4 Street Northwest` but the phrase `4 Street Northwest` would have a lower relevance (0.8).

This would only affect our indexing process and will not affect our behaviour during query time.

### Summary of Changes
- [x] Add a flag to complex token replacement on the indexing side which suggests relevance reduction
- [x] Add functionality to reduce the relevance of a phrase if a flag is set
- [x] Add tests that are indicative of ordinal dropping token replacement - check to see if we get back a result after dropping the ordinal

### Next Steps
- [x] Review 
- [ ] Merge
- [ ] Release + changelog

cc @mapbox/search

### DIFF
--- a/index.js
+++ b/index.js
@@ -204,8 +204,8 @@ function Geocoder(indexes, options) {
             }
             source.categorized_replacement_words = token.categorizeTokenReplacements(info.geocoder_tokens);
             source.simple_replacer = token.createSimpleReplacer(source.categorized_replacement_words.simple);
-            source.complex_query_replacer = token.createComplexReplacer(source.categorized_replacement_words.complex);
-            source.complex_indexing_replacer = token.createComplexReplacer(source.categorized_replacement_words.complex, { includeUnambiguous: true });
+            source.complex_query_replacer = token.createComplexReplacer(source.categorized_replacement_words.complex, { reduceRelevance: false });
+            source.complex_indexing_replacer = token.createComplexReplacer(source.categorized_replacement_words.complex, { includeUnambiguous: true , reduceRelevance: true });
             source.format_helpers = options.formatHelpers;
 
             source.categories = false;
--- a/index.js
+++ b/index.js
@@ -204,8 +204,8 @@ function Geocoder(indexes, options) {
             }
             source.categorized_replacement_words = token.categorizeTokenReplacements(info.geocoder_tokens);
             source.simple_replacer = token.createSimpleReplacer(source.categorized_replacement_words.simple);
-            source.complex_query_replacer = token.createComplexReplacer(source.categorized_replacement_words.complex, { reduceRelevance: false });
-            source.complex_indexing_replacer = token.createComplexReplacer(source.categorized_replacement_words.complex, { includeUnambiguous: true , reduceRelevance: true });
+            source.complex_query_replacer = token.createComplexReplacer(source.categorized_replacement_words.complex, { includeRelevanceReduction: false });
+            source.complex_indexing_replacer = token.createComplexReplacer(source.categorized_replacement_words.complex, { includeUnambiguous: true , includeRelevanceReduction: true });
             source.format_helpers = options.formatHelpers;
 
             source.categories = false;
--- a/lib/indexer/indexdocs.js
+++ b/lib/indexer/indexdocs.js
@@ -66,7 +66,7 @@ function indexdocs(docs, source, options, callback) {
     if (TIMER) console.time('update:freq');
     let freq;
     try {
-        freq = generateFrequency(docs, source.simple_replacer, source.complex_query_replacer, options.tokens, parseInt(source.maxscore, 10), source.lang.has_languages);
+        freq = generateFrequency(docs, source.simple_replacer, source.complex_indexing_replacer, options.tokens, parseInt(source.maxscore, 10), source.lang.has_languages);
     } catch (err) {
         return callback(err);
     }
--- a/lib/indexer/indexdocs.js
+++ b/lib/indexer/indexdocs.js
@@ -66,7 +66,7 @@ function indexdocs(docs, source, options, callback) {
     if (TIMER) console.time('update:freq');
     let freq;
     try {
-        freq = generateFrequency(docs, source.simple_replacer, source.complex_indexing_replacer, options.tokens, parseInt(source.maxscore, 10), source.lang.has_languages);
+        freq = generateFrequency(docs, source.simple_replacer, source.complex_query_replacer, options.tokens, parseInt(source.maxscore, 10), source.lang.has_languages);
     } catch (err) {
         return callback(err);
     }
--- a/lib/text-processing/termops.js
+++ b/lib/text-processing/termops.js
@@ -451,7 +451,7 @@ function getIndexableText(simpleReplacer, complexReplacer, globalReplacer, doc, 
                 const intersectionVariants = token.enumerateTokenReplacements(complexReplacer, tokenize(intersections[l]));
 
                 for (const intersectionVariant of intersectionVariants) {
-                    let intersectionTokens = normalizeQuery(tokenize(intersectionVariant)).tokens;
+                    let intersectionTokens = normalizeQuery(tokenize(intersectionVariant.phrase)).tokens;
 
                     // simpleReplacer to covert things like northwest => nw
                     intersectionTokens = simpleReplacer ? simpleReplacer.replacer(intersectionTokens) : intersectionTokens;
@@ -470,7 +470,7 @@ function getIndexableText(simpleReplacer, complexReplacer, globalReplacer, doc, 
         const variants = token.enumerateTokenReplacements(complexReplacer, tokenize(text));
 
         for (const variant of variants) {
-            const tokenized = tokenize(variant);
+            const tokenized = tokenize(variant.phrase);
             const encoded = normalizeQuery(tokenized).tokens;
 
             // do simple token replacements without regexes -- just look them up in a dict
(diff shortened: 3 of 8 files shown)


lib/indexer/indexdocs.js
```diff
@@ -66,7 +66,7 @@ function indexdocs(docs, source, options, callback) {
     if (TIMER) console.time('update:freq');
     let freq;
     try {
-        freq = generateFrequency(docs, source.simple_replacer, source.complex_query_replacer, options.tokens, parseInt(source.maxscore, 10), source.lang.has_languages);
+        freq = generateFrequency(docs, source.simple_replacer, source.complex_indexing_replacer, options.tokens, parseInt(source.maxscore, 10), source.lang.has_languages);
```
apendleton: Hm. We should be confident that this doesn't slow us way down. The reason we used the query replacer before was, in part, a sense that if we're just counting tokens, it doesn't really matter, and the indexing one has a bunch of complex/slow stuff in it. Did this change fix something specific?

index.js
```diff
@@ -204,8 +204,8 @@ function Geocoder(indexes, options) {
             }
             source.categorized_replacement_words = token.categorizeTokenReplacements(info.geocoder_tokens);
             source.simple_replacer = token.createSimpleReplacer(source.categorized_replacement_words.simple);
-            source.complex_query_replacer = token.createComplexReplacer(source.categorized_replacement_words.complex);
-            source.complex_indexing_replacer = token.createComplexReplacer(source.categorized_replacement_words.complex, { includeUnambiguous: true });
+            source.complex_query_replacer = token.createComplexReplacer(source.categorized_replacement_words.complex, { reduceRelevance: false });
+            source.complex_indexing_replacer = token.createComplexReplacer(source.categorized_replacement_words.complex, { includeUnambiguous: true , reduceRelevance: true });
```
apendleton: I like `reduceRelevance` as the flag on individual replacements. As the flag when creating the replacer, it feels weirder. The replacer should reduce relevance? When/why? What we want to communicate is that we want the replacer to include replacements where reduceRelavence is true.

lib/text-processing/termops.js
```diff
@@ -498,15 +499,19 @@ function getIndexableText(simpleReplacer, complexReplacer, globalReplacer, doc,
             for (const key of keys) {
                 indexableText[key] = indexableText[key] || new Set();
                 for (const language of langs) indexableText[key].add(language);
+                if (variant.reduceRelevance) {
+                    indexableText[key].reduceRelevance = variant.reduceRelevance;
```
apendleton: so this is setting properties on a Set, right? I can live with that, I think, but seems a little clunky

lib/text-processing/termops.js
```diff
@@ -641,8 +648,12 @@ function permutations(terms, weights, all, frequentWords) {
             }
             if (weights) relev += (weights[j] || 0);
         }
+
         if (weights) {
-            permutation.relev  = (wordDropped && (permutation.length === terms.length - 1)) ? Math.max(0.8, Math.round(relev * 5) / 5) : Math.round(relev * 5) / 5;
+            if (reduceRelevance === true && permutation.length === length) permutation.relev = 0.8;
+            else {
```
apendleton: is there ever a circumstance where we would fall into this branch of the if statement while reduceRelevance is true? if so, do we get the right results here?

lib/text-processing/token.js
```diff
@@ -33,22 +33,22 @@ const escapeRegExp = require('./closest-lang').escapeRegExp;
  * @name createComplexReplacer
  *
  * @param {TokenizedQuery} tokens - tokenized text
- * @param {object} inverseOpts - options for inverting token replacements
- * @param {object} inverseOpts.includeUnambiguous - options for
+ * @param {object} options - options for token replacements
+ * @param {object} options.includeUnambiguous - options for whether the keys are unique
```
apendleton: that's not what this does... if this is true, the the complex replacer generates replacements in both directions, forwards and backwards, so like rd => road but also road => rd, but only if they're unambiguous. So like, street => st and saint => st, but not st => ...? because that one is ambiguous.

also, can we document the new property we're adding?

lib/text-processing/termops.js
apendleton: (the answer maybe 'yes', but I'm not 100% sure)

lib/text-processing/token.js
```diff
@@ -84,6 +84,7 @@ module.exports.createComplexReplacer = function(tokens, inverseOpts) {
         let replacementOpts = {};
         if (typeof orig_to == 'object') {
             if (orig_to.text === undefined) continue;
+            if (tokenPair.to.reduceRelevance === true && options.reduceRelevance === false) continue;
```
apendleton: can this just be `(tokenPair.to.reduceRelevance && !options.reduceRelevance)` ? they're already booleans... (also, per above, using the same name to two different concepts, the purpose of a given replacement and the way the replacer is configured, feels confusing)

lib/text-processing/token.js
```diff
@@ -122,6 +123,7 @@ module.exports.createComplexReplacer = function(tokens, inverseOpts) {
                     entry.spanBoundaries = from.split(/\s/).length - 1;
                 }
             }
+            if (tokenPair.to.reduceRelevance === true && options.reduceRelevance === true) entry.reduceRelevance = true;
```
apendleton: same as above: shouldn't need `=== true`

lib/indexer/indexdocs.js
aarthykc: I see, it wasn't clear to me why we were using query time tokens for generating the frequency. Okay, that makes sense to me. Didn't change anything specific!

lib/text-processing/token.js
aarthykc: Ah okay, it wasn't documented before -- can change!

lib/text-processing/termops.js
aarthykc: per chat, this behavior seems to surface results better, but going to add a test to capture current behaviour that if a phrase has an ordinal missing and a frequent word, we still want the relevance to be 0.8

lib/text-processing/termops.js
aarthykc: yup, need the property to propagate to getIndexablePhrases

index.js
aarthykc: changed `reduceRelevance` -> `includeOrdinalReplacement`

lib/text-processing/termops.js
aarthykc: Added!

lib/text-processing/termops.js
miccolis: Oh please reconsider! Can the set member be an object?

aarthykc: resolved this via chat, that we're adding the property to the object `indexableText`.

lib/text-processing/termops.js
```diff
@@ -641,8 +648,12 @@ function permutations(terms, weights, all, frequentWords) {
             }
             if (weights) relev += (weights[j] || 0);
         }
+
         if (weights) {
-            permutation.relev  = (wordDropped && (permutation.length === terms.length - 1)) ? Math.max(0.8, Math.round(relev * 5) / 5) : Math.round(relev * 5) / 5;
+            if (reduceRelevance && permutation.length === length) permutation.relev = 0.8;
+            else {
+                permutation.relev  = (wordDropped && (permutation.length === terms.length - 1)) ? Math.max(0.8, Math.round(relev * 5) / 5) : Math.round(relev * 5) / 5;
```
CyanRook: Nit picky:
Can we get rid of the magic number especially as its used twice?
const minRelavence = 0.8;
let roundedRelevance = Math.round(relev *5)/5;

lib/text-processing/token.js
```diff
@@ -318,22 +323,26 @@ module.exports.enumerateTokenReplacements = function(replacers, text) {
     // Iterate through the versions of each input token and assemble possible
     // combinations of the subsequent tokens.
     const out = [];
-    const assemble = (a, i, o) => {
+    const assemble = (a, i, o, m) => {
```
CyanRook: Nit pick: 
What are a, i, o, m?
3 unnamed variables is already bad, 4 is worse
i is fine as its obviously the index, but the others are ???

test/acceptance/geocode-unit.tokens.test.js
```diff
@@ -578,3 +578,91 @@ tape('teardown', (t) => {
...
+
+// test to see if phrases that have both frequentWords and reduceRelevance are indexed
+// we need the resultant relevance to be 0.8 otherwise we lose the functionality of being able to surface 4 street northwest
+(() => {
+    const conf = {
+        address: new mem({
+            maxzoom: 6,
+            geocoder_frequent_word_list: ['Street'],
+            geocoder_tokens: {
+                '([0-9]+)(?:st|nd|rd|th)': {
```
CyanRook: Nit pick:
Leave a comment near regex that explains what they match or better yet, a const variable that explicitly names it.

test/acceptance/geocode-unit.tokens.test.js
```diff
@@ -578,3 +578,91 @@ tape('teardown', (t) => {
...
+        });
+    });
+})();
+
+(() => {
+    const conf = {
+        address: new mem({
+            maxzoom: 6,
+            geocoder_tokens: {
+                '([0-9]+)(?:st|nd|rd|th)': {
```
CyanRook: Nit Pick:
Leave a comment near regex that explains what they match or better yet, a const variable that explicitly names it.

PR #917, opened 2020-01-21
Carmen-core final

apendleton:
### Context
This PR incorporates all the changes necessary to move our grid storage backend from carmen-cache to carmen-core. It supercedes all the previous PRs to this effect.

It includes some work that has already been reviewed:
- [x] #860 -- this is the main initial carmen-core functionality PR
- [x] #865
- [x] #867
- [x] #869

Additionally, as compared to the earlier `carmen-core-dev` branch, it's up to date with master, and includes some additional work:
- [x] switches the mechanism for adding grid data to the store at index time to use the new, faster `compactAppend` mechanism (see mapbox/carmen-core#60)
- [x] extracts bin-boundary information from fuzzy-phrase and sends it into carmen-core for more-efficient pre-computing of autocomplete results (see, again, mapbox/carmen-core#60 as well as mapbox/fuzzy-phrase#72)
- [x] updates a bunch of tests that depended on previous, arbitrary ordering of tied results in carmen-cache to match the different, but still arbitrary and fine, ordering from carmen-core

### Next Steps
- [x] move to released carmen-core and node-fuzzy-phrase

cc @mapbox/search


test/acceptance/geocode-unit.languageMode.test.js
```diff
@@ -85,8 +85,8 @@ const addFeature = require('../../lib/indexer/addfeature'),
         c.geocode('c', { language: 'en', languageMode: 'strict' }, (err, res) => {
             t.ifError(err);
             t.equal(res.features.length, 2, 'filters features to just those with "en" (x2)');
-            t.equal(res.features[0].place_name, 'China', '0 - China');
-            t.equal(res.features[1].place_name, 'Canada', '1 - Canada');
+            t.equal(res.features[0].place_name, 'Canada', '0 - Canada');
```
CyanRook: Is ordering important here or is it deterministically random?

test/acceptance/geocode-unit.proximity.test.js
```diff
@@ -146,7 +146,7 @@ tape('forward country - single layer - limit', (t) => {
     c.geocode('country', { limit_verify: 1 }, (err, res) => {
         t.ifError(err);
         t.equals(res.features[0].place_name, 'country', 'found country');
-        t.equals(res.features[0].id, 'country.1', 'found country.1');
+        t.equals(res.features[0].id, 'country.2', 'found country.2');
```
CyanRook: Why did this change?

test/unit/geocoder/phrasematch.test.js
```diff
@@ -332,15 +334,13 @@ tape('fuzzyMatchMulti - correct address permutations', (t) => {
...
-                { phrase: ['main'], mask: 2, ender: false, endingType: 0 },
-                { phrase: ['street'], mask: 4, ender: true, endingType: 0 },
-                { phrase: ['1##'], mask: 1, ender: false, endingType: 0 }
+                { phrase: ['100','main','street'], mask: 7, ender: true, ending_type: 0, phrase_id_range: [0, 0] },
+                { phrase: ['1##','main','street'], mask: 7, ender: true, ending_type: 0, phrase_id_range: [0, 0] },
+                { phrase: ['100','main'], mask: 3, ender: false, ending_type: 0, phrase_id_range: [0, 0] },
+                { phrase: ['main','street'], mask: 6, ender: true, ending_type: 0, phrase_id_range: [0, 0] },
+                { phrase: ['1##','main'], mask: 3, ender: false, ending_type: 0, phrase_id_range: [0, 0] },
+                { phrase: ['main'], mask: 2, ender: false, ending_type: 0, phrase_id_range: [0, 0] },
+                { phrase: ['street'], mask: 4, ender: true, ending_type: 0, phrase_id_range: [0, 0] },
```
CyanRook: Why did these other matches drop off?

test/unit/geocoder/phrasematch.test.js
```diff
@@ -359,21 +359,15 @@ tape('fuzzyMatchMulti - correct address permutations: all numbers', (t) => {
...
-                { phrase: ['3##'], mask: 4, ender: true, endingType: 0 }
+                { phrase: ['100', '200', '300'], mask: 7, ender: true, ending_type: 0, phrase_id_range: [0, 0] },
+                { phrase: ['3##', '100', '200'], mask: 7, ender: false, ending_type: 0, phrase_id_range: [0, 0] },
+                { phrase: ['1##', '200', '300'], mask: 7, ender: true, ending_type: 0, phrase_id_range: [0, 0] },
+                { phrase: ['3##', '200'], mask: 6, ender: false, ending_type: 0, phrase_id_range: [0, 0] },
+                { phrase: ['1##', '200'], mask: 3, ender: false, ending_type: 0, phrase_id_range: [0, 0] },
+                { phrase: ['2##', '300'], mask: 6, ender: true, ending_type: 0, phrase_id_range: [0, 0] },
+                { phrase: ['200', '300'], mask: 6, ender: true, ending_type: 0, phrase_id_range: [0, 0] },
+                { phrase: ['100', '200'], mask: 3, ender: false, ending_type: 0, phrase_id_range: [0, 0] },
+                { phrase: ['2##', '100'], mask: 3, ender: false, ending_type: 0, phrase_id_range: [0, 0] },
```
CyanRook: Why did the other matches drop off?

test/unit/geocoder/templating.test.js
```diff
@@ -354,7 +354,7 @@ const addFeature = require('../../../lib/indexer/addfeature'),
         c.geocode('Main st', {}, (err, res) => {
             t.ifError(err);
             t.equals(res.features.length, 2);
-            t.deepEqual(res.features.map((v) => v.place_name), ['Main st', 'Main street']);
+            t.deepEqual(res.features.map((v) => v.place_name).sort(), ['Main st', 'Main street']);
```
CyanRook: Why is sort needed?

test/unit/text-processing/termops.getIndexablePhrases.test.js
```diff
@@ -9,16 +9,16 @@ test('termops.getIndexablePhrases', (t) => {
...
             'relev': 1,
-            'text': 'main st',
             'phrase': 'main st',
+            'hash': 0
         },
         {
             'relev': 0.8,
-            'text': 'main',
-            'phrase': 'main'
+            'phrase': 'main',
```
CyanRook: Is text no longer important?

lib/geocoder/format-features.js
```diff
@@ -378,51 +379,48 @@ function getMatchingText(item, geocoder, requestedLanguage) {
     const textKey = closest ? 'carmen:text_' + closest : 'carmen:text';
     const closestText = item.properties[textKey].split(',')[0];
 
-    if (item.properties['carmen:matches_language'] && (requestedLanguage ? closest === requestedLanguage : true) && !/,/.exec(item.properties[textKey])) return;
+    const hasMultiple = (Array.isArray(item.properties[textKey]) && item.properties[textKey].length > 1) || /,/.exec(item.properties[textKey]);
```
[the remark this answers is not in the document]
CyanRook: Apparently this is a way to do regex in JS - looks weird

docs/index-structure.md
```diff
@@ -2,7 +2,7 @@
 
 There are two types of index stores in Carmen.
 
-- `cxxcache` is used for storing the `grid`, and `freq` indexes.
+- `gridstore` is used for storing the `grid` index.
```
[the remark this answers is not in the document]
apendleton: Carmen hasn't used it for awhile (I think it went away over a year ago?), but carmen-cache still had the ability to represent this kind of information, and we didn't bother to implement it in carmen-core.

lib/geocoder/format-features.js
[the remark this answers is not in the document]
apendleton: Indeed, this is a regex

lib/geocoder/spatialmatch.js
```diff
@@ -1,13 +1,13 @@
...
 const queue = require('d3-queue').queue;
-const coalesce = require('@mapbox/carmen-cache').coalesce;
+const coalesce = require('@mapbox/carmen-core').coalesce;
 const bbox = require('../util/bbox.js');
 const cheapRuler = require('cheap-ruler');
 const roundTo = require('../util/round-to.js');
 const termops = require('../text-processing/termops');
 const constants = require('../constants');
-const PREFIX_SCAN = require('@mapbox/carmen-cache').PREFIX_SCAN;
+const ENDING_TYPE = require('@mapbox/node-fuzzy-phrase').ENDING_TYPE;
```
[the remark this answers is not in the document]
apendleton: There are three kinds of autocomplete-y ness you might want to do, no autocomplete (so, the match has to be a complete match), or regular autocomplete, where the phrase you're looking up can be a prefix of the thing we match, or word-boundary-prefix, which is like the second one, but only allowing for things that hit at a word boundary. This is mostly relevant for token-replacement situations, so like, if your query is "saint" and we replace that with "st", we don't want to match "steven", just "st paul". This thing is just constants (0, 1, 2) for those three types, and both carmen-cache and fuzzy-phrase supplied the same constants, but we don't have carmen-cache anymore, so we get them now from fuzzy-phrase instead. But they're the same things as they were before.

lib/indexer/index.js
```diff
@@ -153,40 +153,42 @@ function update(source, docs, options, callback) {
...
                                 }
-                                langArg.push(source.lang.lang_map[lang]);
+                                lang_set.push(source.lang.lang_map[lang]);
                             }
-                            if (langArg.length === 0) langArg = null;
+                            if (lang_set.length === 0) lang_set = null;
+                        }
+                        const key = { phrase_id, lang_set };
+                        for (const chunk of chunks) {
+                            gridBuilder.compactAppend(key, chunk.relev, chunk.score, chunk.id, chunk.source_phrase_hash, chunk.coords);
```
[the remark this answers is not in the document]
apendleton: In the common case where we're adding a bunch of tiles for the same feature (because it covers a lot of area), we'll have a set of tile coordinates that all correspond to the same phrase, relevance, score, ID, and phrase hash. `chunk` here is a set of coordinates for the same phrase that all have the same relevance, score, ID, and hash. It's like "a bunch of different things that have some common properties and so can be inserted as a unit for performance reasons." Open to other names, but it's sort of a nebulous concept anyway.

lib/util/feature.js
```diff
@@ -309,41 +309,59 @@ function storableProperties(properties, type) {
...
+            if (feature.properties[key] && key.match(/^carmen:text/)) {
+                const synonyms = Array.isArray(feature.properties[key]) ?
+                    feature.properties[key] :
+                    feature.properties[key].split(',');
+                const matchingSynonyms = new Set();
+                for (const synonym of synonyms) {
+                    const hash = termops.phraseHash(synonym);
+                    if (hash === cover.source_phrase_hash) {
+                        matchingSynonyms.add(synonym);
+                    }
```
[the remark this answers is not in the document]
apendleton: Yeah, the limiting in master actually came after the carmen-core version, as a hack because we weren't ready to ship carmen-core yet. The carmen-core approach is "if there's an ID collision, disambiguate by phrase hash, and if there's *still* a tie, use Levenshtein." In master, there is no phrase hash and IDs are shorter and more likely to collide, so just using Levenshtein will happen more and be less reliable, so I wanted an extra seatbelt -- only use the Levenshtein result if we're actually pretty confident in it. But I think we have more signal in carmen-core so that seatbelt shouldn't be necessary.

test/acceptance/geocode-unit.language-universal.test.js
```diff
@@ -119,7 +120,7 @@ tape('Find features using universal text', (t) => {
 });
 
 tape('Finds and ranks features using universal text with language codes', (t) => {
-    c.geocode('HO', { language: 'en' }, (err, res) => {
+    c.geocode('HO', { language: 'fr' }, (err, res) => {
```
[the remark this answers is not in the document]
apendleton: So... I think this test was just wrong. We have a universal_text mechanism that lets you supply a translation for a feature that's valid (so, unpenalized) for all languages. We use this for things like airport codes, which aren't language-specific. The idea behind this test seemed to be "if there's a match that matches on universal text, it should always come back first no matter what," but that's not how universal_text works and not (as far as I know) how it ever worked: universal text and in-language text are both treated the same (as unpenalized); they both beat out-of-language matches but are tied with each other. This query had two perfect matches, one for an in-language match in English, and one for a universal-text match. The test seemed to be asserting that the universal match was coming back first because it always was supposed to, but it wasn't, it was just coming back before the English one by chance; these results were effectively tied. I think a more useful thing to test was that universal text would come back before an out-of-language match, so I changed the search language such that the English translation would be out-of-language but the universal one would still be unpenalized.

test/acceptance/geocode-unit.languageMode.test.js
apendleton: It's deterministic but arbitrary; it used to be ascending by feature ID and is now descending by feature ID, since there's no prox point, but are perfect text matches, and neither has a score.

test/acceptance/geocode-unit.proximity.test.js
apendleton: These features both match this query and differ only by centerpoint and ID; without a proximity point their order is arbitrary and now the other one comes back first.

test/unit/geocoder/templating.test.js
apendleton: Again, the arbitrary order changed; both of these features match equally well (identical centerpoint, perfect text match, no score). After having changed a bunch of the earlier tests to explicitly swap things around, I wondered if for these ones where the order doesn't matter, we should just sort them in the tests to insulate ourselves from arbitrary order changes?

test/unit/text-processing/termops.getIndexablePhrases.test.js
apendleton: Yeah, we used to have two properties that always had the same contents, `phrase` and `text`, because even farther back, carmen-cache used hashes of the text as keys, and one of them contained the original hash but the other one contained its hash, and once we moved to keying by text directly, it was easier to just keep the thing we were using as the key named the same thing, so we just had the same string duplicated twice. Seemed easier now to just have it in there the once, as long as we were rewriting it anyway.

test/unit/geocoder/phrasematch.test.js
apendleton: These went away because of the phrasematch performance optimization where we no longer attempt to match number-only results in address indexes without a proximity point. `geocoder_address` is set to true, below, and we do the phrasematch after that without supplying prox or any other options, so we eliminate `100` and `1##` from consideration.

test/unit/geocoder/phrasematch.test.js
apendleton: See above; these are number-only subqueries in an address index without a prox point.

lib/indexer/index.js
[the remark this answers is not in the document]
CyanRook: Ah okay.  I don't have a better name unless `coordinates` worked, but I'm not sure that captures all that is here.  Not super important, but one thing I had trouble following.

boblannon:
🎉

miccolis:
🎉

PR #919, opened 2020-02-11
Feature prox radius

Nmargolis:
### Context
This PR makes it possible for a feature to have a custom proximity radius for calculating scoredist. This allows for surfacing important features, like airports, from farther away, without using a scoring adjustment that would increase the max score of an index. Increasing the max score of an index has wider ramifications, since the max score is used to encode and decode scores, and compare features across indexes. In addition, this doesn't change the global distribution of scores, it only affects the scoredist calculations when proximity is provided

### Summary of Changes
- [x] Adds `carmen:proximity_radius` as a storable property on a feature for indexing
- [x] Adds a radius option to the proximity.scoredist calculation, which overrides the default zoom-based proximity radius constant
- [x] Passes the feature's `carmen:proximity_radius` property into the scoredist calculation

### Next Steps
- [ ] Release a new carmen version

cc @mapbox/search

### DIFF
--- a/lib/util/proximity.js
+++ b/lib/util/proximity.js
@@ -91,9 +91,9 @@ function center2zxy(center, z) {
  * @param {String} zoom The vector tile zoom level of the index the feature is part of.
  * @return {Number} proximity adjusted score value between 1 and 5,000
  */
-function scoredist(score, minScore, maxScore, dist, zoom) {
+function scoredist(score, minScore, maxScore, dist, zoom, radius) {
     const scoreVal = scoreWeight(score, minScore, maxScore);
-    const distVal = distWeight(dist, zoom);
+    const distVal = distWeight(dist, zoom, radius);
     return distVal * scoreVal;
 }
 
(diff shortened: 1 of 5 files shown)


lib/util/proximity.js
```diff
@@ -91,9 +91,9 @@ function center2zxy(center, z) {
  * @param {String} zoom The vector tile zoom level of the index the feature is part of.
  * @return {Number} proximity adjusted score value between 1 and 5,000
  */
```
apendleton: Can the doc strings be updated to include the new parameter?

lib/util/proximity.js
```diff
@@ -123,8 +123,8 @@ function scoreWeight(score, minScore, maxScore) {
  * @param {String} zoom The vector tile zoom level of the index the feature is part of.
  * @return {Number} scaled score value
  */
```
apendleton: Likewise, can the doc strings be updated?

PR #920, opened 2020-03-11
Rust stackable coalesce

apendleton:
### Context
This PR removes a bunch of compute-intensive stuff from `spatialmatch` and moves it into carmen-core (mostly `stackable`). The new Rust implementations are in https://github.com/mapbox/carmen-core/pull/64

### Summary of Changes
- [x] before, we took all of our phrasematches, produced a series of hypothetical "stacks" based on how they might align, and then submitted each stack into "coalesce" in carmen-core. We now submit all of the phrasematches, unstacked, and both the stacking and the coalescing happen in Rust, using a new tree-traversal approach more throughly detailed in the PR linked above. So from Carmen's perspective, we've mostly just deleted a bunch of code here
- [x] the `rebalance` mechanism, introduced in #924 , has changed pretty dramatically, because the stage where it previously occurred, between `stackable` and `coalesce`, is no longer accessible to Javascript, and the things it operated on (the stacks) no longer have a direct analogue in the new tree-based model. We still have a rebalance operation now, but it works somewhat differently, and operates *after* results come back from coalesce. In general, it's less aggressive than the previous model, so some queries that only worked because of rebalancing before no longer do, and conversely, some queries that were prevented from working by rebalance now work
- [x] several new pieces of metadata have been added to the gridcache structure itself to allow for smarter processing in the new stackable and coalesce logic. Some of these things were properties previously exposed via `phrasematch` objects but that are always the same for a given index, and so are now sent at index instantiation time instead (e.g., the index's ID and type); others were not previously available in carmen-cache/carmen-core, but are now needed for the new, smarter coalesce logic, such as each index's overall bounding box (expressed in tile coords, and potentially decomposed into multiple bounding boxes to allow for antemeridian-crossing bounding boxes).
- [x] we don't have the same hard cutoffs around relevance now that we used to, so some tests now include tests that return more results than they used to

Most of the work in this PR was separately reviewed as sub-PRs; see the links to those PRs in the commit history below.

### Next Steps

cc @mapbox/search


apendleton:
Going to merge again despite lack of coverage because codecov is still borked.

PR #921, opened 2020-03-16
Gridstore opts

apendleton:
### Context
This is the companion PR to mapbox/carmen-core#72, which moves a bunch of fields from phrasematches into the gridstore type, and makes some other changes to how phrasematches are structured and coalesce is called.

### Summary of Changes
- [x] The big one is the refactor of index.js in Carmen, which includes the Carmen constructor that does all of the setup for new geocoding instances. There are a bunch of items that are currently properties on each carmen `source` object that get copied onto every phrasematch produced by that source (idx, zoom, ndx/nmask, bmask, etc.). I wanted to just move those into the store up front, at construction time, rather than send them over the wire on every call. Unfortunately, in master we construct the gridstore object before many of these values are calculated and set, so the order of operations in the construction of new Carmen objects had to change.
- [x] while I was at it, I simplified the structure a bit. It was super callback-hell-y, with an order of operations that was hard to follow. Hopefully it's easier now?
- [x] I changed phrasematch to just return an array rather than a nested thing, since the nested thing just had properties that are now in the gridstore
- [x] I dropped some fields from the phrasematches that get passed in; we were doing some copying from the subquery onto the phrasematch, but we don't actually need any of those copied fields on the rust side
- [x] dropped the editMultiplier, which was a way of keeping both full and penalized relevances around for the complicated heuristics we used to do to rank stacks. We don't rank stacks anymore, so this shouldn't be necessary anymore and we can just penalize the weights directly

### Next Steps
No?

cc @mapbox/search

### DIFF
--- a/index.js
+++ b/index.js
@@ -355,6 +355,8 @@ function Geocoder(indexes, options) {
                     // read case: we'll be creating a GridStore and storing it in _gridstore.reader
                     source._gridstore = {
                         reader: new carmenCore.GridStore(gridStoreFile, {
+                            idx: source.idx,
+                            zoom: source.zoom,
                             type_id: source.ndx,
                             non_overlapping_indexes: source.bmask,
                             coalesce_radius: source.geocoder_coalesce_radius || constants.COALESCE_PROXIMITY_RADIUS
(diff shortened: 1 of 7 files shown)


apendleton:
@aarthykc good call, that was my mistake. It's a change necessary to get rid of the editMultiplier, which is no longer necessary, but I missed that let/const change. Since I see that you've addressed it in your PR, going to go ahead and merge.

PR #924, opened 2020-03-24
Rebalance

aarthykc:
### Context
This PR resurrects rebalance from the dead post stackable-coalesce. There's more work that needs to be done here though - in it's current state it's causing relevances to be higher than usual. Will update the PR with more details after poking at the different penalties applied in verifymatch. Another her change in this version of rebalance is that we're using the relevance instead of the weight, which is probably causing the slight change in relevance.

### Summary of Changes
- [ ]

### Next Steps

cc @mapbox/search


lib/geocoder/spatialmatch.js
```diff
@@ -43,7 +43,10 @@ function spatialmatch(query, phrasematchResults, options, callback) {
     function done(err, results) {
         if (err) return callback(err);
         results = results || [];
-        const combined = results.map((result) => new Spatialmatch(result, preparedPhrasematches.flattenedPhrasematches));
+        const rebalancedResults = results.map((result) => rebalance(query, result, preparedPhrasematches.flattenedPhrasematches));
+        const combined = rebalancedResults.map((result) => new Spatialmatch(result, preparedPhrasematches.flattenedPhrasematches));
+        combined.sort(sortByRelevLengthIdx);
+        combined.sort(sortByRelev);
```
apendleton, line 49: Why are we sorting this twice with different criteria? I'm pretty sure one of these was meant to sort stacks and the other was meant to sort phrasematches.

lib/geocoder/spatialmatch.js
```diff
@@ -43,7 +43,10 @@ function spatialmatch(query, phrasematchResults, options, callback) {
     function done(err, results) {
         if (err) return callback(err);
         results = results || [];
-        const combined = results.map((result) => new Spatialmatch(result, preparedPhrasematches.flattenedPhrasematches));
+        const rebalancedResults = results.map((result) => rebalance(query, result, preparedPhrasematches.flattenedPhrasematches));
+        const combined = rebalancedResults.map((result) => new Spatialmatch(result, preparedPhrasematches.flattenedPhrasematches));
```
apendleton, line 47: Can we do these steps as a single map rather than two? like, `new Spatialmatch(rebalance(...` so that we don't allocate an extra array in the middle?

lib/geocoder/spatialmatch.js
```diff
@@ -95,32 +97,36 @@ function spatialmatch(query, phrasematchResults, options, callback) {
...
     let totalWeight = 0;
-    for (let k = 0; k < stack.length; k++) {
-        stackClone[k] = stack[k].clone();
-        stackClone[k].weight = roundTo(
+    for (let k = 0; k < stack.entries.length; k++) {
+        const entry = stack.entries[k];
+        const phrasematch = flattenedPhrasematches[entry.phrasematch_id];
+        if (stack.entries[k].grid_entry.relev < phrasematch.weight) {
+            stackClone.entries[k].grid_entry.relev = stack.entries[k].grid_entry.relev;
+            continue;
```
CyanRook, line 120: We are dropping the relevance from the total weight here and thus losing relevance on good matches.

lib/geocoder/spatialmatch.js
aarthykc: Good catch! Was left behind during debugging which I should have cleaned up!

lib/geocoder/spatialmatch.js
aarthykc: Done ✅

PR #925, opened 2020-03-25
Structure change tests

aarthykc:
### Context
This PR fixes tests that were failing because of the change the structure of phrasematch and other structural changes in the options of the gridstore.

### Next Steps
- [ ] Get this reviewed
- [ ] Merge into rust-stackable-coalesce

cc @mapbox/search


lib/indexer/index.js
```diff
@@ -230,7 +230,7 @@ function store(source, callback) {
         source._gridstore.writer = null;
         source._gridstore.reader = new carmenCore.GridStore(gridStoreFile, {
             idx: source.idx,
-            zoom: source.zoom,
+            zoom: isNaN(source.zoom) ? 6 : source.zoom,
```
apendleton: Is this one also necessary if we do the one above? I think ideally we make sure source.zoom is a number, and then always pass in that number, rather than letting source.zoom stay NaN but passing a number into carmen-core... I think really weird stuff will happen if the index's zoom and carmen-core's zoom disagree

test/acceptance/geocode-unit.cutoffs.test.js
```diff
@@ -126,51 +126,4 @@ const addFeature = require('../../lib/indexer/addfeature'),
...
-    tape('stackable_limit > 1', (t) => {
-        // providing parameter stackable_limit>1 determines the number of stacks considered for type filtering
-        c.geocode('place 1 united', { autocomplete: true }, (err, res) => {
-            t.ifError(err);
-            t.deepEquals(res.features[0].place_name, 'place 1, United States', 'returns place 1 from United States');
-            t.deepEquals(res.features[0].center, [0,0], 'Center for place 1 from United States');
-            t.deepEquals(res.features[1].center, [0,1], 'Includes results for id.50 place 1');
-            t.end();
-        });
-    });
```
apendleton: I assume this is going away because we're getting rid of the stackable_limit concept?

aarthykc: yes it is!

test/unit/geocoder/phrasematch.test.js
```diff
@@ -268,7 +267,7 @@ tape('fuzzyMatchWindows - expanded & removed term', (t) => {
     const clone = JSON.parse(JSON.stringify(query));
     phrasematch(c, query, {}, (err, results, source) => {
         t.error(err);
-        // t.equal(results.phrasematches.length, 9);
+        // t.equal(results.length, 9);
```
apendleton, line 270: Should this still be commented out?

lib/indexer/index.js
apendleton: We should still address this, and make sure it's never possible for source.zoom to be a NaN, so that this check becomes unnecessary.

PR #926, opened 2020-04-09
Redo Rebalance

CyanRook:
### Context

Current rebalance is causing a huge regression in pass rates.  This to attempt to improve this while still bringing in some form of rebalancing.

### Summary of Changes
- [x] Rebalance occurs after stackable and coalesce instead of between them
- [x] A balance is attempted to be struck between number of stack elements and the coverage of those individual elements

### Next Steps

cc @mapbox/search

### DIFF
--- a/lib/geocoder/spatialmatch.js
+++ b/lib/geocoder/spatialmatch.js
@@ -43,7 +43,8 @@ function spatialmatch(query, phrasematchResults, options, callback) {
     function done(err, results) {
         if (err) return callback(err);
         results = results || [];
-        const combined = results.map((result) => new Spatialmatch(result, preparedPhrasematches.flattenedPhrasematches));
+        const combined = results.map((result) => new Spatialmatch(rebalance(query, result, preparedPhrasematches.flattenedPhrasematches), preparedPhrasematches.flattenedPhrasematches));
+        combined.sort(sortByRelev);
 
         // Ascending and Descending order here refers to being able to support
         // `address, place, region, country` and `country, region, place, address`
@@ -77,7 +78,6 @@ function spatialmatch(query, phrasematchResults, options, callback) {
                 filteredSpatialmatches.push(spatialmatch);
             }
         }
-
         return callback(null, { results: filteredSpatialmatches, sets: sets, waste: [] });
     }
 }
@@ -95,32 +95,36 @@ function spatialmatch(query, phrasematchResults, options, callback) {
  * @param {Array} stack results for a subquery combination
  * @returns {Array} - rebalanced stack
  */
-function rebalance(query, stack) {
+function rebalance(query, stack, flattenedPhrasematches) {
     let stackMask = 0;
-    const stackClone = [];
+    const stackClone = Object.assign({}, stack);
 
-    for (let i = 0; i < stack.length; i++) {
-        stackMask |= stack[i].mask;
+    for (let i = 0; i < stack.entries.length; i++) {
+        stackMask |= stack.entries[i].mask;
     }
 
     const garbage = (query.length === (stackMask.toString(2).split(1).length - 1)) ? 0 : 1;
-    const totalLengthBonus = .01 * (garbage + stack.length);
-    const weightPerMatch = (1 / (garbage + stack.length)) - 0.01;
+    const totalLengthBonus = .01 * (garbage + stack.entries.length);
+    const weightPerMatch = (1 / (garbage + stack.entries.length)) - 0.01;
 
     // shallow copy stack into stackClone to prevent cases where a stack's
     // index gets overwritten in deep copies.
     let totalWeight = 0;
-    for (let k = 0; k < stack.length; k++) {
-        stackClone[k] = stack[k].clone();
-        stackClone[k].weight = roundTo(
+    for (let k = 0; k < stack.entries.length; k++) {
+        const entry = stack.entries[k];
+        const phrasematch = flattenedPhrasematches[entry.phrasematch_id];
+        if (stack.entries[k].grid_entry.relev < phrasematch.weight) {
+            stackClone.entries[k].grid_entry.relev = stack.entries[k].grid_entry.relev;
+            continue;
+        }
+        stackClone.entries[k].grid_entry.relev = roundTo(
             weightPerMatch +
-            (totalLengthBonus * stack[k].weight)
+            (totalLengthBonus * stack.entries[k].grid_entry.relev)
             , 8);
-        totalWeight += stackClone[k].weight;
+        totalWeight += stackClone.entries[k].grid_entry.relev;
     }
 
     stackClone.relev = Math.min(roundTo(totalWeight, 8), 1);
-
     return stackClone;
 }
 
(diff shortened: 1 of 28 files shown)


test/acceptance/geocode-unit.near-alignment.test.js
```diff
@@ -158,7 +158,7 @@ tape('build queued features', (t) => {
 tape('Check correctly aligned one', (t) => {
     c.geocode('100 main st atlanta georgia 80139', { limit_verify: 10 }, (err, res) => {
         t.ifError(err);
-        t.equals(res.features.length, 2);
+        t.equals(res.features.length, 4);
```
CyanRook, line 161: This seems like a carmen-core change? I'm not creating new features....

test/unit/geocoder/spatialmatch.rebalance.test.js
```diff
@@ -1,79 +1,105 @@
...
     stack.relev = 1;
 
-    const rebalanced = rebalance(query, stack);
+    const rebalanced = rebalance(query, stack, phraseMatches);
     t.equal(rebalanced.relev, 1, 'relev = 1');
-    t.equal(rebalanced[0].weight, 0.26, 'weight = 0.25');
-    t.equal(rebalanced[1].weight, 0.24666667, 'weight = 0.25');
-    t.equal(rebalanced[2].weight, 0.24666667, 'weight = 0.25');
-    t.equal(rebalanced[3].weight, 0.24666667, 'weight = 0.25');
+    t.equal(rebalanced.entries[0].grid_entry.relev, 0.5, '1## main st weight = 0.5');
```
CyanRook, line 40: A query that is completely covered by the feature stack no longer gets changed, this is because something weird with mystery changes in stack relevance that I haven't been able to track down yet.

lib/geocoder/spatialmatch.js
```diff
@@ -97,34 +97,46 @@ function spatialmatch(query, phrasematchResults, options, callback) {
...
     for (let i = 0; i < stack.entries.length; i++) {
         stackMask |= stack.entries[i].mask;
     }
+    const stackQueryCoverage = stackMask.toString(2).split(1).length - 1;
+    const missingQueryTokens = query.length - stackQueryCoverage;
+    const stackLength = stack.entries.length;
+    const averageStackCoverage = stackQueryCoverage / stackLength;
+    const entryTokenBonus = 0.01;
+
+    let stackWeight = 1 / stackLength;
```
CyanRook, line 110: This is used to account for whether or not we cover the whole query

lib/geocoder/spatialmatch.js
```diff
@@ -97,34 +97,46 @@ function spatialmatch(query, phrasematchResults, options, callback) {
...
-            stackClone.entries[k].grid_entry.relev = stack.entries[k].grid_entry.relev;
-            continue;
+    let expectedWeight = 0;
+    for (let k = 0; k < stackLength; k++) {
+        const entry = JSON.parse(JSON.stringify(stack.entries[k]));
+        expectedWeight += entry.grid_entry.relev;
+        if (missingQueryTokens > 0) {
+            const entryQueryCoverage = entry.mask.toString(2).split(1).length - 1;
+            const entryTokenDifference = entryQueryCoverage - averageStackCoverage;
+            const entryWeight = entryQueryCoverage / query.length;
```
CyanRook, line 129: This is to track reductions to an entries relevance that aren't due to the entries coverage of the query.  We need to be able to recreate theses discounts during the rebalancing of the values

lib/geocoder/spatialmatch.js
```diff
@@ -97,34 +97,46 @@ function spatialmatch(query, phrasematchResults, options, callback) {
...
-            weightPerMatch +
-            (totalLengthBonus * stack.entries[k].grid_entry.relev)
-            , 8);
-        totalWeight += stackClone.entries[k].grid_entry.relev;
+        totalWeight += entry.grid_entry.relev;
+        stackClone.entries.push(entry);
     }
 
-    stackClone.relev = Math.min(roundTo(totalWeight, 8), 1);
+    const stackPenalty = expectedWeight - stack.relev;
```
CyanRook, line 137: This is to track stack relevance penalties that aren't tied to specific entries.

lib/geocoder/spatialmatch.js
```diff
@@ -97,34 +97,46 @@ function spatialmatch(query, phrasematchResults, options, callback) {
...
-    const stackClone = Object.assign({}, stack);
 
     for (let i = 0; i < stack.entries.length; i++) {
         stackMask |= stack.entries[i].mask;
     }
+    const stackQueryCoverage = stackMask.toString(2).split(1).length - 1;
+    const missingQueryTokens = query.length - stackQueryCoverage;
+    const stackLength = stack.entries.length;
+    const averageStackCoverage = stackQueryCoverage / stackLength;
+    const entryTokenBonus = 0.01;
```
aarthykc, line 108: what's the purpose of an `entryTokenBonus`?

test/unit/geocoder/spatialmatch.rebalance.test.js
```diff
@@ -1,79 +1,105 @@
 'use strict';
 
-/*
+
 const rebalance = require('../../../lib/geocoder/spatialmatch').rebalance;
 const Phrasematch = require('../../../lib/geocoder/phrasematch').Phrasematch;
 const test = require('tape');
 
+function old_to_new(phrase_matches) {
```
aarthykc, line 8: Not a blocker and can come in a following PR but we're switching between camel casing and snake casing throughout this file, we should keep this consistent. Also, could we change the function name to `preparePhrasematches` or something more relevant to someone who looks at this repo and has no context about old vs new?

lib/geocoder/spatialmatch.js
CyanRook: It's part of trying to give a little extra push to entries in a stack that covered more tokens. So if an entry covers more tokens it gets slightly more to that particular entry and ones with less get slightly penalized but it doesn't change the overall stack relevance.  It was something I had early, not sure if it is still needed but helped at the time.

lib/geocoder/spatialmatch.js
```diff
@@ -97,34 +97,46 @@ function spatialmatch(query, phrasematchResults, options, callback) {
...
     let totalWeight = 0;
-    for (let k = 0; k < stack.entries.length; k++) {
-        const entry = stack.entries[k];
-        const phrasematch = flattenedPhrasematches[entry.phrasematch_id];
-        if (stack.entries[k].grid_entry.relev < phrasematch.weight) {
-            stackClone.entries[k].grid_entry.relev = stack.entries[k].grid_entry.relev;
-            continue;
+    let expectedWeight = 0;
+    for (let k = 0; k < stackLength; k++) {
+        const entry = JSON.parse(JSON.stringify(stack.entries[k]));
```
aarthykc, line 124: Do we need to JSON.parse? can this not just be `const entry = stack.entries[k];`

lib/geocoder/spatialmatch.js
aarthykc: What kind of penalities are these? The reduceRelevance ones, out-of-language match?

CyanRook: Whichever, we can't tell at this point and there seems to be multiple from different places.   There is a ticket here to try and track them to figure all this out.
https://github.com/mapbox/carmen/issues/928

test/unit/geocoder/spatialmatch.rebalance.test.js
CyanRook: Fixed

lib/geocoder/spatialmatch.js
CyanRook: There are tests to ensure you get a copy, not a reference, but I don't know if that still applies with carmen core changes? Without it, that test fails due to JS copy mechanisms.

PR #930, opened 2020-05-05
carmen-core@nearby-only

apendleton:
### Context
This PR is the companion to mapbox/carmen-core#89 . It switches to pointing at that gitsha, and also now passes in a `nearbyOnly` property in numerical-autocomplete cases. Turns out we used to want to flag exactly these same cases with a property called `extendedScan` that doesn't do anything in carmen-core anymore (carmen-cache used it and it's now ignored), so I just renamed that.

### Summary of Changes
- [x] rename extendedScan to nearbyOnly
- [x] switch to carmen-core@nearby-only

### Next Steps
No

cc @mapbox/search

### DIFF
--- a/lib/geocoder/phrasematch.js
+++ b/lib/geocoder/phrasematch.js
@@ -522,7 +522,12 @@ function Phrasematch(subquery, weight, mask, phrase, phrase_id_range, scorefacto
 }
 
 Phrasematch.prototype.clone = function() {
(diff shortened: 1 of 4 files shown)


apendleton:
This PR scope-creeped a little. It turns out the Phrasematch.clone() function, which this PR altered:
* was not tested, and
* because it wasn't tested, apparently didn't actually work (we don't use this function in our code... not even sure why it's there?)

This PR now:
- [x] fixes this function
- [x] tests this function to ensure that it works

PR #931, opened 2020-05-06
switch to carmen-core@bbox-aware-coalesce

apendleton:
### Context
This PR is the companion to mapbox/carmen-core#91 . It switches to pointing at that gitsha, and now calculates a zxy bounding box for each index at carmen start time, and passes that bounding box into the gridstore of the index at instantiation, so that it can be used in coalesce.

### Summary of Changes
- [x] switch to carmen-core@bbox-aware-coalesce
- [x] calculate an index's bounding box in tile coordinates, and pass it into the gridstore

### Next Steps
Currently this is PR'ed against nearby-only since it includes that branch's work as well, but I will probably retarget once that branch lands.

Also: currently for indexes whose bounds cross the antemeridian, a tile-oriented bounding box is calculated that includes all latitudes. It might be better to instead either add support for AM-crossing bounding boxes to carmen-core, or to add support for including multiple bounding boxes to carmen-core, and decompose an index's bounding box into per-hemisphere components before instantiation. I'm leaving this to a possible follow-up PR, however.

cc @mapbox/search

### DIFF
--- a/index.js
+++ b/index.js
@@ -247,12 +247,10 @@ function Geocoder(indexes, options) {
 
             if (source.bounds[0] < source.bounds[2]) {
                 source.tileBounds = bbox.insideTile(source.bounds, source.zoom).slice(1);
-                //console.log("no am", source.tileBounds);
             } else {
                 // this index crosses the antemeridian; just blow it out around the earth
                 const blownBounds = [-180, source.bounds[1], 180, source.bounds[3]];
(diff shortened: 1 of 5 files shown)


apendleton:
This one scope-creeped too: turns out we didn't have any tests that had indexes with antemeridian-crossing bounds. Now we do.

PR #932, opened 2020-05-28
Use only the best address matches

apendleton:
#896 refactored the mechanism by which addresses are identified in an address cluster to be returned, and somewhat altered the way multiple matches can be returned from the same query. We do some degree of partial matching, to ignore, e.g., added-on letters (so e.g., matching "12" when "12B" was requested). In the now-present logic, both exact and approximate matches are treated equally, and after later deduplication, sometimes partial matches can beat out perfect matches in the final results. This means on a block with addresses 1, 1A, 1B, 1C, etc., only the coordinates for "1" might be surfaceable.

The new logic assigns each prospective match a rank: 0 for perfect, >0 for some level of approximate, with higher numbers worse. As the cluster is processed, we keep track of the best rank we've seen so far, and we don't retain any matches that are worse-ranked than our current best. If we see one better than our current best, we throw away all the current matches (which all must be worse) and use the new rank as best from then on.

I've updated the tests that checked partial matching as well. They now conform to the new behavior, which only uses the fall-back/approximate-match logic on numbers if there isn't an exact match. The affected tests now typically return fewer results total (but there aren't many affected).

### Summary of Changes
- [x] update the address cluster matching logic to differentiate between matches of differing quality
- [x] update tests

### Next Steps
Nope

cc @mapbox/search

### DIFF
--- a/lib/geocoder/addresscluster.js
+++ b/lib/geocoder/addresscluster.js
@@ -143,7 +143,6 @@ function forwardPrefix(feature, address) {
             const featureMatchStrings = generateMatchStrings(cluster[c_it][addressIndex], [addressStyle]);
 
             const rank = matchesStyle(queryMatchStrings, featureMatchStrings, addressStyle, true);
-            // console.log(queryMatchStrings, featureMatchStrings, rank);
             if (rank !== -1 && feature.geometry.geometries[c_it].type === 'MultiPoint') {
                 if (rank < matchQuality) {
                     // we got a better kind of match than we had seen so far;
@@ -372,7 +371,7 @@ function matchesStyle(queryMatchStringMaps, featureMatchStringMaps, style, prefi
  * @param {Map<String, String>} queryMatchStrings - Mapping from match string name to match string for the query
  * @param {Map<String, String>} featureMatchStrings - Mapping from match string name to match string for the feature
  * @param {boolean} prefixMatch - Whether to match on a partial result
- * @returns {boolean} - True if a match
+ * @returns {Number} - -1 if no match found; integer >= 0 if found a match, where lower is better
  */
 function matchesStandardAddress(queryMatchStrings, featureMatchStrings, prefixMatch = false) {
     if (prefixMatch) {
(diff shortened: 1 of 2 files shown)


apendleton:
Codecov seems to be broken:

image

I think maybe because the current tip of master is a docs change that doesn't affect coverage, so it might be missing a coverage report? I tried rerunning Travis for both master and this branch and it did nothing, even though CI says the report uploaded successfully. In the screenshot above, the "uploads" section is blank, even though CI claims it uploaded each time it ran. Very odd. But not sure what else to do, so I'm going to post screenshots of the coverage numbers below, and with approval, merge anyway.

Master:
image

This branch:
image

Overall improvement of 0.1% 🎉 .

lib/geocoder/addresscluster.js
```diff
@@ -371,11 +396,14 @@ function matchesStandardAddress(queryMatchStrings, featureMatchStrings, prefixMa
  */
```
aarthykc, line 396: We probably should change the documentation for what the function returns as well?

lib/geocoder/addresscluster.js
```diff
@@ -68,6 +68,7 @@ function forward(feat, address, num = 10) {
     const queryMatchStrings = generateMatchStrings(address, addressStyles);
 
     const matchedAddressFeatures = [];
+    let matchQuality = Number.MAX_SAFE_INTEGER;
```
aarthykc, line 71: Oh wow, didn't know this. Nice!

lib/geocoder/addresscluster.js
```diff
@@ -130,7 +142,18 @@ function forwardPrefix(feature, address) {
 
             const featureMatchStrings = generateMatchStrings(cluster[c_it][addressIndex], [addressStyle]);
 
-            if (matchesStyle(queryMatchStrings, featureMatchStrings, addressStyle, true) && feature.geometry.geometries[c_it].type === 'MultiPoint') {
+            const rank = matchesStyle(queryMatchStrings, featureMatchStrings, addressStyle, true);
+            // console.log(queryMatchStrings, featureMatchStrings, rank);
```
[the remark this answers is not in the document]
apendleton: Fixed 🤦

lib/geocoder/addresscluster.js
apendleton: Fixed

apendleton:
Fixed feedback. No change in coverage per:

image

but codecov is still broken, so that ^ is documentation that the coverage is fine, and I'm going to admin-merge.

PR #935, opened 2020-06-26
Uncomment tests

aarthykc:
### Context
We had commented out a couple of tests during the early stages of the rewrite of stackable/coalesce given that we hadn't added any limits in places. Uncommenting some of the tests now that we've added some limits in place to throw away less relevant results. We're still pulling in a lot more matches than we did previously, so some tests still fail because of the number of candidates we're still considering. @apendleton would love a second set of eyes on this PR

### Next Steps
- [x] Reviewed by @apendleton 
- [ ] Add more limits in place and make sure tests pass now 

cc @apendleton 

cc @mapbox/search

### DIFF
--- a/test/acceptance/geocode-unit.cutoffs.test.js
+++ b/test/acceptance/geocode-unit.cutoffs.test.js
@@ -80,7 +80,6 @@ const addFeature = require('../../lib/indexer/addfeature'),
         q.awaitAll(t.end);
     });
 
-    // @FIXME limit
     tape('max_correction_length > query length', (t) => {
         // Number of words in the query = 6
         // parameterized max_correction_length = 5
@@ -88,8 +87,7 @@ const addFeature = require('../../lib/indexer/addfeature'),
         // for a query whose length is greater than the max_correction_length
         c.geocode('place places 11 unitted states america however extreme', { max_correction_length: 0 }, (err, res) => {
             t.ifError(err);
-            console.log(res);
-            t.equals(res.features.length, 0, 'ok, does not return a result for max_correction_length > query length');
+            t.equals(res.features[0].relevance < 1, true, 'ok, returns a feature with relevance < 1');
             t.end();
         });
     });
--- a/test/acceptance/geocode-unit.cutoffs.test.js
+++ b/test/acceptance/geocode-unit.cutoffs.test.js
@@ -87,7 +87,7 @@ const addFeature = require('../../lib/indexer/addfeature'),
         // for a query whose length is greater than the max_correction_length
         c.geocode('place places 11 unitted states america however extreme', { max_correction_length: 0 }, (err, res) => {
             t.ifError(err);
-            t.equals(res.features[0].relevance < 1, true, 'ok, returns a feature with relevance < 1');
+            t.equals(res.features[0].relevance < 0.6, true, 'ok, returns a feature with relevance < 0.6');
             t.end();
         });
     });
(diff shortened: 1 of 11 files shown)


test/acceptance/geocode-unit.cutoffs.test.js
```diff
@@ -81,17 +81,18 @@ const addFeature = require('../../lib/indexer/addfeature'),
...
-    //         t.ifError(err);
-    //         t.equals(res.features.length, 0, 'ok, does not return a result for max_correction_length > query length');
-    //         t.end();
-    //     });
-    // });
+    tape('max_correction_length > query length', (t) => {
+        // Number of words in the query = 6
+        // parameterized max_correction_length = 5
+        // this test case should not return results because we should not attempt fuzzy search
+        // for a query whose length is greater than the max_correction_length
```
aarthykc, line 88: We're still somehow fuzzy matching and picking up the right feature, which we shouldn't.

test/acceptance/geocode-unit.emoji.test.js
```diff
@@ -97,16 +95,16 @@ tape('should handle a query including emoji', (t) => {
...
-// tape('should handle a CJK query including emoji that triggers stacking', (t) => {
-//     // Black star
-//     const query = 'Anarres 南🗾';
-//     c.geocode(query, {}, (err, res) => {
-//         t.ifError(err);
-//         t.equal(res.features.length, 0, 'finds no features');
-//         t.end();
-//     });
-// });
+//@FIXME limit
```
aarthykc, line 98: We're still returning a result here.

aarthykc: Relevance is around ~0.4

test/acceptance/geocode-unit.geocoder_stack.test.js
```diff
@@ -414,14 +413,14 @@ const { queueFeature, buildQueued } = require('../../lib/indexer/addfeature');
     //         t.end();
     //     });
     // });
-    // tape('Place', (t) => {
```
aarthykc, line 417: Need to take a closer look at the stacking, the commented out tests in this file are still failing. Any initial thoughts @apendleton?

test/acceptance/geocode-unit.cutoffs.test.js
```diff
@@ -80,16 +80,14 @@ const addFeature = require('../../lib/indexer/addfeature'),
...
     tape('max_correction_length > query length', (t) => {
         // Number of words in the query = 6
         // parameterized max_correction_length = 5
         // this test case should not return results because we should not attempt fuzzy search
         // for a query whose length is greater than the max_correction_length
         c.geocode('place places 11 unitted states america however extreme', { max_correction_length: 0 }, (err, res) => {
             t.ifError(err);
-            console.log(res);
-            t.equals(res.features.length, 0, 'ok, does not return a result for max_correction_length > query length');
+            t.equals(res.features[0].relevance < 1, true, 'ok, returns a feature with relevance < 1');
```
apendleton, line 90: We might want to be tighter with the check here than `< 1`. What we're trying to assert is "we didn't get this by fuzzy-matching," and if we *had* fuzzy-matched this, we'd still have a relevance < 1, but it would be pretty high (maybe .8 for the phrase, so .9-something for the whole stack). We're getting it now even though its overall relevance is super-low, because we don't do relevance cutoffs before coalesce anymore, but we should maybe confirm that the relevance is super low, to assert that we're not matching this phrase at all, because of the fuzzy search limits. I'm not sure what it actually is now, but if it's < .5, maybe we should assert that.

PR #878, opened 2019-09-09
New templating library

aarthykc:
### Context

This PR replaces our home-grown templating system with an off-the-shelf that is more maintainable. After trying out a few different templating engine, we chose to go with Handlebars as our templating engine. Along with reworking/refactoring our templating engine, this PR also adds the following functionalities - 
1. Adds a mechanism for registering helper functions and making them available within templates to provide more customisable formatting options.
2. Allow feature specific template overrides - so for every feature if we have a `carmen:format`, we will prefer that format over the index format.

The way we select which template to apply to the feature is in the following order: 
- language-specific per-feature template - `carmen:format_<lang>`
- default per-feature template - `carmen:format`
- language-specific index-level template  - `geocoder_format_<lang>`
- default index-level template - `geocoder_format`

### Summary of Changes
- [x] Tears out home grown templating engine replacing it with handlebars
- [x] Adds mechanism to utilise helper functions

### Next Steps
- [x] Add per feature overrides
- [x] Get tests to pass
- [x] Add more tests for which format will be selected based on the query
- [x] Review 
- [ ] Merge and Release once we get bench results

cc @mapbox/search

PR #940, opened 2020-07-17
Fix template selection with per-feature overrides

apendleton:
### Context
#878 introduced a way for individual features to override their template for producing their `place_name` geocoder responses. It turns out there was a bug in the way this was implemented such that if a feature in the *context* of a result, rather than the result itself, had an overridden template, that template would be used to render the response even though it was for the wrong feature (for example, if you looked for an address and it had an overridden template on its postcode, the address would be rendered with the postcode's template, so you'd end up with a place name without an address in it).

While I was fixing that, I noticed some other tidbits we hadn't caught the first time around:
- if a feature had both a language-specific template and a generic one, we'd use the generic one instead of the language-specific one (but would render both, and through away the language-specific one)
- we'd only use language overrides for features from indexes that also had an index-wide template, but there doesn't seem to be a reason that'd be necessary
- we weren't using the `closestLang` mechanism for choosing which language-specific template to use, so e.g., if you were querying with language=en_US, we wouldn't use the en template.

This PR fixes all of that.

### Summary of Changes
- [x] pick one template before rendering rather than rendering multiple
- [x] pick using closestLang, and prefer language-specific over generic
- [x] do this regardless of whether or not the index has an index-wide template
- [x] add some more tests to cover these cases

### Next Steps
Nope

cc @mapbox/search

### DIFF
--- a/lib/geocoder/format-features.js
+++ b/lib/geocoder/format-features.js
@@ -48,6 +48,19 @@ function getFormatString(context, format, language) {
  */
 function getPlaceName(context, formatString, language, languageMode, matched, source) {
     const feat = context[0];
+
+    if (Object.keys(feat.properties).filter((k) => k.match(/^carmen:format/)).length) {
+        const templateLang = language ? closestLang.closestLangLabel(language, feat.properties, 'carmen:format_') : undefined;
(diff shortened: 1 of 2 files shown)


apendleton:
Merging without coverage again, because codecov is still broken. Coverage looks fine on CI.

PR #942, opened 2020-07-24
Enable worldviews filter in carmen cli query

bsrinivasa:
### Context

This PR adds ability to filter worldviews in the carmen cli query and is an extension of Worldviews work at https://github.com/mapbox/carmen/pull/918. 

### Summary of Changes
- [x] Updated `bin/carmen.js` file to add worldviews parameter and associated values

### Next Steps
- Review and merge
- Bump and release a version

cc @mapbox/search @apendleton @aarthykc

### DIFF
--- a/bin/carmen.js
+++ b/bin/carmen.js
@@ -73,7 +73,8 @@ if (argv.tokens) {
 }
 
 const carmen = new Carmen(opts, {
-    tokens: tokens
+    tokens: tokens,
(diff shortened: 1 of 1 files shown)


bsrinivasa:
As per the discussion, it would not be ideal to hardcode specific values to the worldview parameter in the carmen and would be good to pass this as arguments or values back to carmen from downstream.
cc @apendleton @aarthykc

PR #946, opened 2020-09-29
Squishy context

apendleton:
### Context
This branch adds a new variant of the squishy bonus. In the original version, we look for candidate results where the result feature has a name that is the same as a feature in its context, and boost that result (so if you look for "new york," "new york new york" gets a boost). It doesn't do anything, though, for cases where both of the nested features are in the context, so for example, "1 main st new york" might match either "1 main st" features in NYC, or in other cities in New York, with equal weight. The earlier gappy penalty would have helped to prefer the address-in-city examples, but as that's no longer in place (and can't really be introduced because it caused lots of weird collateral behavior), we're out of luck here, especially when the scores of the results are tied.

This branch introduces a new mechanism that's like the Squishy bonus, but applied to contexts. We piggyback on the same Carmen setting (`geocoder_inherit_score`), and examine each context to see if any text-matched features have this property set. If they do, on subsequent (more general) features that we encounter while processing the context, we check to see if the name is similar to the name of the earlier-identified feature. We will only consider at most one feature as a candidate to be matched to, so whichever is the most specific wins. Also note: while the regular squishy bonus operates on score, context items don't contribute to score, so this bonus operate on relevance instead.

Because sometimes the winning features are all tied at 1.0, I made an additional tweak, to allow relevance values to temporarily exceed 1.0; they're still clamped to 1.0, but the clamping has now been moved to after the final sort, so that the temporary 1.01 relevance can allow the better features to float to the top before they become tied again.

Lastly: I added an additional sort criterion in spatialmatch for features coming out of coalesce on their way into verifymatch. In some cases with lots of tied matches, I was finding that the feature I wanted never had a chance to make it to the point where it would get the squishy bonus applied, because it wasn't making it over the verifymatch limit. I added an additional criterion for cases that are otherwise tied (same score, same relevance, same index) to prefer results whose average idx value (index ID) is higher -- this has the effect of preferring items whose text-matched context is more specific, because more specific indexes will have higher IDs. So if you have "1 main st new york," a candidate where "new york" is a city will have a higher average index ID than a candidate where "new york" is a state. Note that all of this will get sorted again in verifymatch anyway, so this only matters with respect to making it above or below the limit, not to the final relative order of features that both make it through.

### Summary of Changes
Nope

### Next Steps

cc @mapbox/search

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -25,11 +25,15 @@
         }],
         "valid-jsdoc": ["warn", {
             "requireReturn": false
-        }]
+        }],
+        "node/no-unsupported-features": [ "error", { "version": 10 } ]
     },
     "env": {
         "node": "true",
         "es6": "true"
+    },
+    "parserOptions": {
+        "ecmaVersion": 2019
     },
     "extends": ["@mapbox/eslint-config-geocoding"]
 }
--- a/test/acceptance/geocode-unit.override.test.js
+++ b/test/acceptance/geocode-unit.override.test.js
@@ -212,7 +212,7 @@ const { queueFeature, buildQueued } = require('../../lib/indexer/addfeature');
         c.geocode('9B FAKE STREET PARKER 20001', { limit_verify: 10 }, (err, res) => {
             t.ifError(err);
             t.equals(res.features[0].place_name, '9b fake street Parker 20002', 'found 9b fake street parker 20002 w/ 20001 query');
-            t.equals(res.features[0].relevance, 0.818519);
+            t.equals(res.features[0].relevance, 0.725926);
             t.deepEquals(res.features[0].context, [
                 { id: 'place.4', text: 'Parker' },
                 { id: 'postcode.5', text: '20002' }
--- a/test/unit/geocoder/verifymatch.test.js
+++ b/test/unit/geocoder/verifymatch.test.js
@@ -111,3 +111,36 @@ tape('verifymatch.verifyFeatures', (t) => {
     t.end();
 
 });
+
+tape('verifymatch.textAlike', (t) => {
+    const feats = [
+        {
+            'carmen:text': 'Apple',
+            'carmen:text_fr': 'Orange',
+            'carmen:text_de': 'Pear',
+            'carmen:geocoder_stack': 'us'
+        },
+        {
+            'carmen:text': 'Grapefruit',
+            'carmen:text_fr': 'Lemon',
+            'carmen:text_de': 'Lime',
+            'carmen:geocoder_stack': 'us'
+        },
+        {
+            'carmen:text': 'Grapefruit',
+            'carmen:text_fr': 'Orange',
+            'carmen:text_de': 'Kumquat',
+            'carmen:geocoder_stack': 'us'
+        },
+    ];
+
+    const textAlike = verifymatch.textAlike;
+    t.assert(!textAlike(feats[0], feats[1]), 'Features with no shared text properties aren\'t alike');
+    t.assert(textAlike(feats[1], feats[2]), 'Features with shared default text are alike');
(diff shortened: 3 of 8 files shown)


test/acceptance/geocode-unit.override.test.js
```diff
@@ -212,7 +212,7 @@ const { queueFeature, buildQueued } = require('../../lib/indexer/addfeature');
         c.geocode('9B FAKE STREET PARKER 20001', { limit_verify: 10 }, (err, res) => {
             t.ifError(err);
             t.equals(res.features[0].place_name, '9b fake street Parker 20002', 'found 9b fake street parker 20002 w/ 20001 query');
-            t.equals(res.features[0].relevance, 0.818519);
+            t.equals(res.features[0].relevance, 0.725926);
```
apendleton, line 215: Explanation here: overrides have two different ways of calculating the relevance for the postcode component that's being added by the override. If the match had a postcode in it, we reuse that postcode's weight. If not, we do a kinda-sorta-hazy-math thing to guess what it might have been, given rebalancing, etc.

In this case, there are two different ways to arrive at a stack for this feature, one that includes the town and one that includes the postcode, and their relevance is tied. Whichever one we process first wins and the other gets deduped away, but which one is first is basically random. Even though they're tied before the override, though, they produce different override results because we end up using a different method to calculate the correct weight. Because this PR perturbs the way one of the sorts happens, it jumbles this result, and might do something similar in production occasionally by chance, but this isn't systematic (half the time, the other one would have won anyway). Long term, we should change how the weights for overrides are calcualted.

aarthykc (approved):
So excited to get this out! @apendleton not blocking this PR, but maybe we should spend some time coming up with better names for backy, squishy and other penalties here or document the behaviour somewhere in our README?

apendleton:
@aarthykc per earlier discussion, I revised this a bit. The `textAlike` function that is used to determine if two features in the context have the same name is now optionally language-aware: it can accept a `Set` as an optional argument of the languages it should consider when deciding if the names are similar. It always considers the default text, regardless, and still considers all languages if a set isn't specified. Now, for the context-squishy call to this function, we pass in whatever the languages are that we would treat as default at indexing time for that country, if specified (via the `geocoder_languages_from_default` index property, which specifies default languages per country, plus the `geocoder_stack` property of the feature being considered).

I also added some unit tests for `textAlike`, since there weren't any.

Last: the way I wrote the newer/better textAlike uses a regular expression with a [named capture group](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions/Groups_and_Ranges#:~:text=Named%20capturing%20group), which ends up with cleaner code than if I had used numbered groups. This has been supported in Python, etc., forever, but didn't come to JS/Node until Node 10, and our eslint rules said we only allowed things supported in Node 8. We don't use Node 8 anymore, and it's EOL, so I went ahead and allowed Node 10 in our config (and bumped our package.json engines too for good measure, since they still said we supported Node 6, which we definitely don't). Hope that's cool.

test/unit/geocoder/verifymatch.test.js
```diff
@@ -111,3 +111,36 @@ tape('verifymatch.verifyFeatures', (t) => {
     t.end();
 
 });
+
+tape('verifymatch.textAlike', (t) => {
```
aarthykc, line 115: 👌🏽  for adding a test!

.eslintrc.json
```diff
@@ -25,11 +25,15 @@
...
+        "node/no-unsupported-features": [ "error", { "version": 10 } ]
     },
     "env": {
         "node": "true",
         "es6": "true"
     },
+    "parserOptions": {
+        "ecmaVersion": 2019
+    },
     "extends": ["@mapbox/eslint-config-geocoding"]
```
aarthykc, line 38: Should we think about switching this to mapbox-config at some point?

apendleton:
Codecov is still broken. Coverage on this branch is fine per Travis. Admin-merging anyway.

PR #949, opened 2020-10-23
Better address number matching if unit numbers are present

apendleton:
### Context
For standard (non-Queens) address numbers, currently the two strategies we have in place to match the number from a query with the number in the address are exact matching, and matching of versions of each where all the non-numeric characters are stripped out (so, "7-1" becomes "71," and we can match that if "71" was queried). We've noticed a common pattern in Europe where address numbers might include some unit information, and queries might omit that information, so, "7-1" might get searched for as "7." Our current strategy doesn't handle that case.

This PR adds a third strategy, which is to only include all of the numeric characters until we encounter the first non-numeric one (so, for "7-1," we'd produce "7"). This strategy is lower-ranked than either of the first two strategies, so if there are successful matches using either of those, they'll be preferred. This kind of matching is preferred over interpolation, if both are possible.

### Summary of Changes
- [x] add new house number matching strategy
- [x] add unit test for this strategy

### Next Steps
Nope

cc @mapbox/search

### DIFF
--- a/lib/geocoder/addresscluster.js
+++ b/lib/geocoder/addresscluster.js
@@ -317,9 +317,14 @@ function generateStandardAddressMatchStrings(address) {
     const numericAddress = typeof address === 'string'
         ? address.replace(/[^\d]/, '')
         : address.toString();
+    // A numeric version of just the part of the query preceding the first non-numeric character (e.g., for "7-1", this will be "7")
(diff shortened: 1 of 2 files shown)


aarthykc (approved):
🌮 Makes sense to me!

apendleton:
Going to override codecov again (still need to find time to fix that)

PR #952, opened 2020-12-02
Template helpers

apendleton:
### Context
#878 moved Carmen's templates to use Handlebars. This branch builds a little on that to make some template operations in Carmen consumers a bit easier.

### Summary of Changes
- [x] add a new dependency [handlebars-helpers](https://github.com/helpers/handlebars-helpers), and make the `comparison` operations from that library available to all templates. This makes it possible to put conditionals in templates based on equality (so like "if the country equals X, display this way, otherwise display that way").
- [x] add a new utils file for sticking custom helpers we find useful
- [x] add one specific helper for reformatting address strings, which is helpful for a particular customer usecase (I think it's arguable that this should live in Carmen consumers instead of Carmen itself, but it seems pretty harmless to just have a collection we can start to build up in Carmen itself in case these are broadly useful)

### Next Steps
Nope

cc @mapbox/search

### DIFF
--- a/lib/geocoder/format-features.js
+++ b/lib/geocoder/format-features.js
@@ -44,6 +44,7 @@ function getFormatString(context, format, language) {
  * @param {String} languageMode option that can be set to 'strict' filter results strictly to a specific language. Undefined defaults null.
  * @param {Boolean} matched If set, use matched text in place name
  * @param {Carmen} geocoder A carmen object used to generate the results
+ * @param {String} worldview Optionally, the current worldview (or "_all") -- templates will have access to this if present (or undefined otherwise) in case it's useful
  * @return {String} formatted place_name string
  */
 function getPlaceName(context, formatString, language, languageMode, matched, source, worldview) {
--- a/lib/geocoder/format-features.js
+++ b/lib/geocoder/format-features.js
@@ -46,7 +46,7 @@ function getFormatString(context, format, language) {
  * @param {Carmen} geocoder A carmen object used to generate the results
  * @return {String} formatted place_name string
  */
-function getPlaceName(context, formatString, language, languageMode, matched, source) {
+function getPlaceName(context, formatString, language, languageMode, matched, source, worldview) {
     const feat = context[0];
 
     if (Object.keys(feat.properties).filter((k) => k.match(/^carmen:format/)).length) {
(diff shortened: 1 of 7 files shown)


lib/geocoder/format-features.js
```diff
@@ -46,7 +46,7 @@ function getFormatString(context, format, language) {
  * @param {Carmen} geocoder A carmen object used to generate the results
  * @return {String} formatted place_name string
  */
-function getPlaceName(context, formatString, language, languageMode, matched, source) {
+function getPlaceName(context, formatString, language, languageMode, matched, source, worldview) {
```
bookest, line 49: Should the `worldview` parameter be added to the JsDoc for this function?

lib/geocoder/format-features.js
```diff
@@ -141,7 +141,15 @@ function toFeature(context, format, languages, languageMode, debug, geocoder, cl
...
 
-    const source = geocoder ? geocoder.byidx[feat.properties['carmen:idx']] : {};
+    const sources = context.map((f) => {
+        return geocoder ? (geocoder.byidx[f.properties['carmen:idx']] || geocoder.indexes[f.properties['carmen:index']]) : {};
+    });
+    const source = sources[0];
+
+    let worldview = '_all';
+    for (const s of sources) {
+        if (s.geocoder_worldview !== '_all') worldview = s.geocoder_worldview;
```
bookest, line 151: Is the intent that `worldview` be set to the worldview of the last source with a non `_all` worldview in order or can this break out once the conditional has matched?

apendleton: @bookest yeah, while I was mucking around in here, I wanted to add the ability to condition template decisions on the worldview of the feature (not necessary for either of the changes I'm working on that will go out with this release, but useful for another customer request in the pipeline). Unfortunately at this point in the process, we don't have easy access to the query parameters the user submitted, and it would be a sort of tortuous refactor to pass them through to this point, but the current worldview can be inferred from the combination of worldviews in the context stack, so I opted (hackily) for that path. Basically, every request has a worldview, implicitly, and every index can either be all-worldviews or worldview-specific, so the context stack for the feature we're processing will either be all all-worldview indexes, or a combination of all-worldview indexes and indexes from at most one specific worldview. So we can figure out the worldview, if there is one, by hunting through the indexes to figure it out.

Where this might be useful would be, for example, if we have a POI layer that's valid for all worldviews, but something about the context we display under it should change by worldview (above and beyond just showing a different country name, which we get from worldviews for free). Assuming countries change per worldview but POIs don't, we'd notice that even though our feature is from an all-worldviews layer, it's being stacked on a worldview-specific country layer, and can infer the worldview from that layer.

lib/geocoder/format-features.js
apendleton: Added

CyanRook (approved):
lgtm

apendleton:
Admin-merging because codecov is still broken 😢

PR #953, opened 2021-01-11
Terminal word boundaries

apendleton:
### Context
This PR makes a change to how queries that end in word boundaries are handled. Previously, whitespace at the ends of queries was stripped, and other terminal word boundaries were ignored, which for autocomplete queries might mean that we might treat the final token in the query as being the prefix of a longer word, even though the user had an expressed an intent that that word was complete (by typing a space or other boundary character). We now preserve this terminal word boundary info, and if it's present, use the `WORD_BOUNDARY` prefix scanning mode instead of the regular prefix scanning mode in our fuzzy_phrase queries, which forces the final token to be treated as a complete word and limits phrase results accordingly.

### Summary of Changes
- [x] preserve terminal whitespace at query time (we still want to strip it at index time, so some calls to `.trim()` needed to be moved into indexing code and out of functions shared between indexing and querying to make this work)
- [x] use WORD_BOUNDARY fuzzy-phrase querying if terminal word boundary characters are present

### Next Steps
- [x] add a test or two to more explicitly test this new behavior

cc @mapbox/search

### DIFF
--- a/lib/geocoder/phrasematch.js
+++ b/lib/geocoder/phrasematch.js
@@ -80,9 +80,19 @@ module.exports = function phrasematch(source, query, options, callback) {
         hypo.replaced = token.replaceToken(source.complex_query_replacer, hypo.query);
         hypo.gapExpansionMasks = gapMasks(hypo.replaced);
         hypo.normalized = termops.normalizeQuery(hypo.replaced);
-        hypo.endingType = options.autocomplete ?
-            (hypo.normalized.lastWord ? ENDING_TYPE.wordBoundaryPrefix : ENDING_TYPE.anyPrefix) :
-            ENDING_TYPE.nonPrefix;
+        if (options.autocomplete) {
+            const endsInBoundary = hypo.normalized.separators[hypo.normalized.separators.length - 1] !== '';
+            if (hypo.normalized.lastWord || endsInBoundary) {
+                // if it's autocomplete but either we token-replaced the last word, or
+                // the query ends with a word-boundary, only prefix-match whole-word prefixes
+                hypo.endingType = ENDING_TYPE.wordBoundaryPrefix;
+            } else {
+                // otherwise all prefixes, including partial-world ones, are fine
+                hypo.endingType = ENDING_TYPE.anyPrefix;
+            }
+        } else {
+            hypo.endingType = ENDING_TYPE.nonPrefix;
+        }
     }
 
     let subqueries;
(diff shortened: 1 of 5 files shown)


lib/text-processing/token.js
```diff
@@ -409,7 +409,7 @@ module.exports.createGlobalReplacer = function(tokens) {
  */
 module.exports.replaceGlobalTokens = function(replacers, text) {
     for (const replacer of replacers) {
-        text = text.trim().replace(replacer.from, replacer.to);
+        text = text.replace(replacer.from, replacer.to);
```
apendleton, line 412: Now only some places where this function is used will want the trim of trailing space, so I've deleted it here and moved it to those individual callsites instead.

lib/geocoder/phrasematch.js
```diff
@@ -80,9 +80,19 @@ module.exports = function phrasematch(source, query, options, callback) {
...
+                // if it's autocomplete but either we token-replaced the last word, or
+                // the query ends with a word-boundary, only prefix-match whole-word prefixes
+                hypo.endingType = ENDING_TYPE.wordBoundaryPrefix;
+            } else {
+                // otherwise all prefixes, including partial-world ones, are fine
+                hypo.endingType = ENDING_TYPE.anyPrefix;
+            }
+        } else {
+            hypo.endingType = ENDING_TYPE.nonPrefix;
+        }
```
apendleton, line 95: This is mostly just a reformatting, but the ternaries got a little too messy with one more thing to check.

aarthykc: I like it!

test/acceptance/geocode-unit.tokens.test.js
```diff
@@ -359,7 +359,7 @@ const addFeature = require('../../lib/indexer/addfeature'),
         });
     });
     tape('test token replacement', (t) => {
-        c.geocode('Tal st ', { limit_verify: 1, fuzzyMatch: 0 }, (err, res) => {
```
aarthykc, line 362: Was this change intentional?

aarthykc (approved):
Makes sense to me, had one question about a test - otherwise looks good to me 🌮

test/acceptance/geocode-unit.tokens.test.js
apendleton: @aarthykc yeah, this test doesn't have anything to do with word boundaries as far as I can tell -- I think the space at the end might have been a mistake on my part? Not sure (3 years ago is a bit fuzzy). But the expectation was that it match "...strasse", which it won't anymore because the space makes the "st" no longer prefix match. Taking out the space fixes it again, and I think is in line with the original intent of the test.

apendleton:
Coverage is still broken. Merging anyway.

PR #955, opened 2021-01-28
Move where the ghost feature dedup happens in verifymatch

apendleton:
### Context
Deduplication occurs at several points in Carmen. There's a deduplication operation specific to ghost features that happens in verifymatch. Currently it takes place after we've done coarse spatial alignment in carmen-core, but before we've done fine spatial alignment in verifymatch. A corner case exists where a ghost feature and a non-ghost feature could share a label, and the ghost feature could get deduplicated away, and then after that, the non-ghost feature could get rejected because it fails fine spatial alignment, such that neither feature makes it into the final result set even though the ghost feature might have been fine, and might have been better than any remaining features.

This PR swaps the order of those operations, so that we deduplicate only after fine spatial alignment has occurred, and if we throw out a ghost feature, we know it will be in favor of something that spatially aligns.

(Sidenote: I'm sort of skeptical that this deduplication operation is necessary at all, given that we have later deduplication operations as well, and those later ones take feature context into account, which this one does not. Still, this change seems like the least-invasive possible change, so it's good enough for now.)

### Summary of Changes
- [x] move where the ghost dedup happens in verifymatch
- [x] revamp ghost/dedup tests to cover this circumstance

### Next Steps
Nope

cc @mapbox/search

### DIFF
--- a/lib/geocoder/verifymatch.js
+++ b/lib/geocoder/verifymatch.js
@@ -524,20 +524,7 @@ function verifyFeatures(query, geocoder, spatialmatches, loaded, options, startP
     // the best results at this point.
     result.sort(sortFeature);
 
-    // Eliminate any score < 0 results if there are better-scored results
-    // with identical text.
-    const filtered = [];
-    const byText = {};
-    for (let i = 0; i < result.length; i++) {
-        const feat = result[i];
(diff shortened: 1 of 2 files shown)


aarthykc (approved):
🌮

apendleton:
Going to override codecov again, because it's still broken

PR #957, opened 2021-03-01
Upgrade to Node 14

apendleton:
### Context

This PR makes Carmen compatible with Node 14.

### Summary of Changes
- [x] upgrade to [carmen-core@modernize-plus-fuzzy](https://github.com/mapbox/carmen-core/pull/98), which builds Node-14-compatible binaries, upgrades to current Neon, and incorporates a direct dependency on `fuzzy-phrase`, which was previously a separate module
- [x] accordingly, remove dependency on `node-fuzzy-phrase`, which will now be deprecated
- [x] adapt imports according to the above
- [x] update the following additional deps:
    - [x] mbtiles
    - [x] vtquery (which we got updated as part of this work 🎉)
    - [x] mapnik
    - [x] murmurhash-native
- [x] update some additional dependencies (tape, nyc) to get tests passing, and make some slight tweaks in accordance with their changes
- [x] change the order of some results in tests -- Node 12 introduced a [new sorting algorithm](https://v8.dev/blog/array-sort), and as a consequence, some results in tests that were otherwise tied have seen their orders arbitrarily shift
- [x] fix a weird crasher in tests (explained below)

### Next Steps
- [ ] release carmen-core and switch to released carmen-core

cc @mapbox/search

### DIFF
--- a/bin/carmen-index.js
+++ b/bin/carmen-index.js
@@ -96,6 +96,7 @@ function index(err) {
     carmen.on('open', () => {
         carmen.index(process.stdin, conf.to, config, (err) => {
             if (err) throw err;
+            process.exit(0);
         });
     });
 }
--- a/test/unit/geocoder/context.test.js
+++ b/test/unit/geocoder/context.test.js
@@ -512,11 +512,12 @@ test('contextVector near miss polygons', (t) => {
         context.contextVector(source, 39, 39, false, {}, null, false, false, undefined, (err, data) => {
             t.ifError(err);
             t.equal(data['properties']['carmen:text'], 'A');
-        });
-        context.contextVector(source, 40.0001, 40.0001, false, {}, null, false, false, undefined, (err, data) => {
-            t.ifError(err);
-            t.equal(data, false);
-            t.end();
+
+            context.contextVector(source, 40.0001, 40.0001, false, {}, null, false, false, undefined, (err, data) => {
+                t.ifError(err);
+                t.equal(data, false);
+                t.end();
+            });
         });
     });
 });
--- a/test/unit/geocoder/phrasematch.test.js
+++ b/test/unit/geocoder/phrasematch.test.js
@@ -5,7 +5,7 @@ const tape = require('tape');
 const phrasematch = require('../../../lib/geocoder/phrasematch');
 const termops = require('../../../lib/text-processing/termops');
 const token = require('../../../lib/text-processing/token');
-const ENDING_TYPE = require('@mapbox/node-fuzzy-phrase').ENDING_TYPE;
+const ENDING_TYPE = require('@mapbox/carmen-core').ENDING_TYPE;
 
 function bearablePermutations(permutations) {
     return permutations.map((v) => {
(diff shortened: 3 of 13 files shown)


test/unit/geocoder/phrasematch.test.js
```diff
@@ -364,14 +364,14 @@ tape('fuzzyMatchMulti - correct address permutations: all numbers', (t) => {
...
-                { phrase: ['1##', '200'], mask: 3, ender: false, ending_type: 0, phrase_id_range: [0, 0] },
-                { phrase: ['2##', '300'], mask: 6, ender: true, ending_type: 0, phrase_id_range: [0, 0] },
-                { phrase: ['200', '300'], mask: 6, ender: true, ending_type: 0, phrase_id_range: [0, 0] },
+                { phrase: ['3##', '100', '200'], mask: 7, ender: false, ending_type: 0, phrase_id_range: [0, 0] },
                 { phrase: ['100', '200'], mask: 3, ender: false, ending_type: 0, phrase_id_range: [0, 0] },
+                { phrase: ['200', '300'], mask: 6, ender: true, ending_type: 0, phrase_id_range: [0, 0] },
+                { phrase: ['1##', '200'], mask: 3, ender: false, ending_type: 0, phrase_id_range: [0, 0] },
                 { phrase: ['2##', '100'], mask: 3, ender: false, ending_type: 0, phrase_id_range: [0, 0] },
+                { phrase: ['2##', '300'], mask: 6, ender: true, ending_type: 0, phrase_id_range: [0, 0] },
+                { phrase: ['3##', '200'], mask: 6, ender: false, ending_type: 0, phrase_id_range: [0, 0] },
```
apendleton, line 374: This order is arbitrary, and has changed because of the sort algorithm change discussed above.

test/unit/geocoder/context.test.js
```diff
@@ -512,11 +512,12 @@ test('contextVector near miss polygons', (t) => {
...
-        context.contextVector(source, 40.0001, 40.0001, false, {}, null, false, false, undefined, (err, data) => {
-            t.ifError(err);
-            t.equal(data, false);
-            t.end();
+
+            context.contextVector(source, 40.0001, 40.0001, false, {}, null, false, false, undefined, (err, data) => {
+                t.ifError(err);
+                t.equal(data, false);
+                t.end();
+            });
```
apendleton, line 520: The new version of tape is pickier about calling `t.whatever` after `t.end()` has been called, so this and other tests were rearranged a bit (not functionally changed) to be compliant with the update.

bin/carmen-index.js
```diff
@@ -96,6 +96,7 @@ function index(err) {
     carmen.on('open', () => {
         carmen.index(process.stdin, conf.to, config, (err) => {
             if (err) throw err;
+            process.exit(0);
```
apendleton, line 99: The `bin` tests were crashing running this file. `carmen-index.js` doesn't crash when run ordinarily, and the test doesn't crash except when run under `nyc`, the coverage analyzer, so I *think* this was due to a bug in `nyc` or `istanbul`, possibly interacting with the Node upgrade or Neon ugprade. This `exit` fixes it, though, and this is the end point of the program anyway, so it's not a functional change.

bookest (approved):

bookest (approved):
LGTM

apendleton:
Switched carmen-core to the released version. Also upgraded `documentation.js,` which is a dev dependency -- it turns out the version we were using has a dependency ('chokidar`) that has a dependency (`fsevents`) that specifically lacks a Node 14 binary on Mac, and I had mostly been working on Linux machines for this work and hadn't noticed. Upgrading fixed it, and we don't actually use it anywhere, except to manually build docs, so it affected nothing.

Codecov is still broken. Going to admin-merge as per usual.

PR #959, opened 2021-03-08
Roll back explicit exit in carmen-index.js

apendleton:
### Context
We use `nyc` to run coverage analysis in CI for Carmen. It intercepts imports of JS source files and instruments them for coverage. We optionally accept JS source files as command line arguments in our `carmen-index.js` process, and it turns out that something about doing this under `nyc` on Node 14 caused a segfault during runs. For #957, I added an explicit `process.exit(0)` at the end of the indexing process, which seemed to prevent the segfault. It now seems, however, that because we pass data from Carmen to subsequent indexing stages via STDOUT, this exit, while it takes place after all JS code has completed, can cause the process to terminate before the STDOUT buffer has been fully flushed, potentially resulting in features not being included in built tiles.

This branch rolls back that exit, and instead just skips the test that triggers the nyc/Node 14 bug altogether, by detecting whether or not we're running under the coverage runner. I added a naked call to `tape` just for that test as part of the `yarn coverage` routine, to ensure that this test still gets run on CI (just without coverage instrumentation).

I confirmed in a manual build of a large index that this fixes the issue.

### Summary of Changes
- [x] revert process.exit() change
- [x] skip test if running under nyc
- [x] explicitly run affected test without nyc

### Next Steps
No

cc @mapbox/search

### DIFF
--- a/bin/carmen-index.js
+++ b/bin/carmen-index.js
@@ -96,7 +96,6 @@ function index(err) {
     carmen.on('open', () => {
         carmen.index(process.stdin, conf.to, config, (err) => {
             if (err) throw err;
-            process.exit(0);
         });
     });
 }
--- a/package.json
+++ b/package.json
@@ -64,7 +64,7 @@
   "scripts": {
     "lint": "eslint index.js lib test bin",
     "test": "yarn run lint && (retire -n || echo 'WARNING: retire found insecure packages') && TESTING=true tape './test/**/*.js' && yarn run bench",
-    "coverage": "TESTING=true nyc tape 'test/**/*.js' && nyc report --reporter=text-lcov > coverage.lcov && codecov",
+    "coverage": "TESTING=true tape test/bin.test.js && nyc tape 'test/**/*.js' && nyc report --reporter=text-lcov > coverage.lcov && codecov",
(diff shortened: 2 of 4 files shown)


bin/carmen-index.js
```diff
@@ -96,7 +96,6 @@ function index(err) {
     carmen.on('open', () => {
         carmen.index(process.stdin, conf.to, config, (err) => {
             if (err) throw err;
-            process.exit(0);
```
bookest, line 99: If we depend on `stdout` being flushed for correct operation, would it make sense to add an explicit call to `fs.fsyncSync(config.output.fd)` to ensure the buffer is flushed before we exit?

bookest (approved):
One suggestion, but LGTM.

bin/carmen-index.js
apendleton: @bookest yes, probably, though honestly using stdout for this purpose is probably ill-advised in general, and I'm also not 100% sure which buffer needs to be flushed (like, if `console.log` output gets buffered elsewhere first). Without the explicit exit call, node doesn't exit until it's all done anyhow, so I think we're safe just leaving it alone.

apendleton:
Yolo-merging around codecov again as per usual

PR #960, opened 2021-03-12
Force geocoder_coalesce_radius to be a number

bookest:
### Context

In some cases when reading metadata from mbtiles sources parameters that should be numeric, such as geocoder_coalesce_radius are converted to strings. This leads to errors like this when instantiating a GridStore:

    TypeError: invalid type: string "400", expected f64

The root problem of parameters changing types still needs to be tracked down, but this approach mitigates the problem for now.

### Summary of Changes
- [X] convert geocoder_coalesce_radius to a number

### Next Steps
n/a

cc @mapbox/search


apendleton (approved):
🌮

Issue #963, opened 2021-04-21
Move `mapnik` to `devDependencies`


mattficke:
### Context

In https://github.com/mapbox/carmen/pull/873, @CyanRook replaced Mapnik with VTQuery. However, we're still installing Mapnik as a production dependency: https://github.com/mapbox/carmen/blob/c050f95b9237dcfdff354947e3b49cd2f8eed5da/package.json#L38

The only place we're still using Mapnik is in the `addfeature` module, which is marked as deprecated and is only used internally in unit tests:
https://github.com/mapbox/carmen/blob/c050f95b9237dcfdff354947e3b49cd2f8eed5da/lib/indexer/addfeature.js#L2-L7

Mapnik is a relatively large library, around 90MB installed. If it's not needed for downstream users of Carmen, we should move it to `devDependencies` to reduce the size of its dependencies.

### Next Steps

Move mapnik from `dependencies` to `devDependencies`

@mapbox/search

PR #873, opened 2019-08-26
VTQuery - Geocoding #930

CyanRook:
* Update carmen to use VTQuery

### Context
Changing Carmen to use VTQuery of Mapnik.
https://github.com/mapbox/geocoding/issues/930

### Summary of Changes
- [x] Upgrade carmen from mapnik to VTQuery
- [x] Adjust distScore to match mapnik performance
- [x] Adjust tests in alignment with https://github.com/mapbox/carmen/pull/856

### Next Steps

cc @mapbox/search

PR #964, opened 2021-04-21
move mapnik to devDependencies

mattficke:
### Context

Refs #963 

With the switch to VTQuery in https://github.com/mapbox/carmen/pull/873, mapnik is now only used in unit tests, and no longer needs to be installed as a production dependency.

### Summary of Changes
- [ ] Move mapnik from `dependencies` to `devDependencies`

cc @mapbox/search


mattficke:
While I was in here I removed a couple of other dependencies that aren't needed anymore after the changes from https://github.com/mapbox/carmen/pull/712.

CyanRook (approved):

PR #972, opened 2021-11-01
256indexes

allisonzheng:
### Summary of Changes
Adds a unit test for `tmpid` collisions with 256 indexes to test that every index's `tmpid` is unique

cc @michaelskahn


test/acceptance/geocode-unit.index-limit.test.js
```diff
@@ -77,6 +77,70 @@ tape('reverse place', (t) => {
...
+
+    for (let i = 0; i < 256; i++) {
+        geocodeArray.push(new Promise((resolve,reject) => {
+            collisonCarmen.geocode('country' + i + '_', { limit_verify: 1, debug: 1 }, (err, res) => {
+                if (err) return reject(err);
+                t.equals(res.debug.spatialmatch.covers[0].tmpid, (i * 2 ** 24) + 1, 'tmpid does not overflow');
+                t.equals(res.debug.spatialmatch.covers[0].tmpid, res.debug.verifymatch[0].properties['carmen:tmpid'], 'tmpids consistent across phases');
+                t.equals(res.features[0].place_name, 'country' + i + '_', 'found place');
+                t.equals(res.features[0].relevance, 1.00);
+                tmpIdCount[res.debug.spatialmatch.covers[0].tmpid] = tmpIdCount[res.debug.spatialmatch.covers[0].tmpid] ? tmpIdCount[res.debug.spatialmatch.covers[0].tmpid] + 1 : 1;
```
michaelskahn, line 129: This applies to other parts of my PR as well, but especially in this test case, we're referencing `res.debug.spatialmatch.covers[0].tmpid` enough times that storing it in a new variable would make the code easier to read.

michaelskahn (changes requested):
The logic here makes sense to me. One thing to note is that carmen uses `yarn` rather than `npm`, so it uses `yarn.lock` instead of a `package-lock.json`.

michaelskahn (approved):

